Issue #3, opened 2025-11-21
Smoother page changes


becked:
There is a bit of jank when changing pages, brief show of a loading... text and then a flash of a chart.

Issue #2, opened 2025-11-21
Fix janky import progress bar


becked:
It moves back and forth and window changes size.

Priority is fixing that but also take a look and see how "time remaining" can be made more accurate or perhaps consider removing it.

PR #35, opened 2026-05-05
Cloud rewrite: browser-first SvelteKit + Cloudflare D1/R2/KV

becked:
Implements the cloud-first rewrite specified in `docs/cloud-rewrite-spec.md`: Tauri + Rust + DuckDB → SvelteKit + Cloudflare Worker (D1/R2/KV) with all save parsing in a browser Web Worker. Tauri code is left in place for now and removed in a follow-up.

## Summary

- **Parser TS port** — 46/46 entities ported from Rust to TypeScript, runs in a Web Worker. Backed by a Rust↔TS parity harness so output is byte-equivalent for the share assembly.
- **Cloudflare Worker API** — D1 schema with numbered migrations (users, games, player_summaries, game_player_turn, tech/law events, online_ids), R2 for parsed-game JSON + raw save ZIPs, KV for sessions. Multipart upload pipeline with Valibot validation, SHA-256 dedup, and batched D1 inserts respecting the 100-param cap.
- **Auth** — Discord OAuth (PKCE, KV sessions, `/v1/auth/{start,callback,me,logout}`), `safeNext` open-redirect sanitizer, OAuth `next` round-trip via `OAuthPending`.
- **App surface** — Games library, full game detail view (shared component reused from desktop), dashboard mirroring desktop overview, /upload, /account (Discord identity + linked OnlineIDs + sign out), persistent CloudHeader on cloud routes.
- **Public sharing** — `is_public` toggle, anonymous read of `/games/[id]`, OG-card unfurling, online_id PII strip on anonymous reads, Cache-Control split + `Vary: Cookie, Origin`.
- **SSR via adapter-cloudflare** — `BUILD_TARGET` env switch keeps Tauri build untouched, per-route `ssr=false` for /upload + /auth/callback, hooks.server.ts security headers, kit.csp hash mode (no `unsafe-inline` scripts).
- **Rate limiting** — anonymous reads 200/hr per-IP via Cache API, with Discord/Slack/Twitter/Facebook/LinkedIn UA bypass; PATCH visibility 60/hr per-user; download 50/hr per-user + 100/hr per-IP; 429 surfaces as proper 429 page (not 500) on SSR'd game route.
- **Productionization touches** — raw save download (`GET /v1/games/:id/download`, RFC 6266 filename + audit logged), re-import on parser version bump (banner + ZIP round-trip + semver-gated overwrite preserving created_at/is_public), owner Cache-Control `no-store` so post-mutation `invalidateAll()` actually re-hits the worker, dev CSP allowing `localhost:8787`.
- **/v1/stats + /dashboard** with `player_summaries` backfill (16 milestone columns populated on every upload).

## What's not in this PR (tracked as remaining work)

- Atlas migration to R2 (`assets.per-ankh.app/v/atlases/...`)
- Production cutover: KV/D1/secret setup on prod, custom domain wiring (per-ankh.app SSR + api.per-ankh.app API), `wrangler-frontend.toml`, rate-limit tuning
- Tauri removal: drop `src-tauri/`, `/game/`, `Header.svelte`, `GameSidebar.svelte`, `ShareControl.svelte`, Tauri-bound `api.ts`, root-layout cloud-route gate
- Dynamic per-game OG image (satori + resvg-wasm)
- `_routes.json` tuning (adapter-cloudflare warns about 292 dropped exclude rules)
- Cross-subdomain cookie auto-forwarding verification on staging
- Real Discord/Slack/Twitter unfurl test (needs deployed public URL)
- Mobile-width header layout pass on `/games/[id]`
- Account-deletion path; unlink-Discord once a second auth provider exists
- Desktop-side: align Rust `get_completed_techs` to dedupe on (player, tech)

## Test plan

- [ ] `npm run check` (TS) and `npm run lint` clean
- [ ] Worker tests: `cd cloud && npm test`
- [ ] Parser parity harness passes against the corpus (Rust output ≡ TS output for share assembly)
- [ ] Local dev: `npm run dev` (SvelteKit, BUILD_TARGET=cloud) + `cd cloud && npm run dev` — end-to-end sign-in → upload → games list → game detail → public toggle → anon view → re-import → download → delete
- [ ] Anonymous read of public game (logged-out browser session) returns 200 with PII stripped
- [ ] Owner mutations (visibility toggle, re-import) reflect immediately (no stale cache)
- [ ] 429 page renders on SSR'd game route when rate limit hit
- [ ] OG card preview via Discord/Slack debugger (after deploy)


becked (commented):
# Code review

## Overview

A 17K-line cloud-first rewrite: parser ported Rust→TS (46/46 entities) running in a Web Worker, Cloudflare Worker API (D1+R2+KV), Discord OAuth, public sharing with PII strip, SvelteKit SSR via adapter-cloudflare, dual-build that keeps Tauri intact. Backed by a Rust↔TS parity harness. Architecturally sound; the issues below are mostly tightening, not blockers.

---

## Findings by severity

### High — worth fixing before merge

- **Open-redirect sanitizer gaps** (`cloud/src/auth.ts:62`, `src/lib/utils/safe-next.ts`) — `safeNext` allows query strings and fragments through. If `next` is ever logged, reflected in a UI, or consumed by anything beyond a same-origin path-only navigator, `?x==https://evil.com` and `#https://evil.com` payloads slip the check. Tighten by rejecting `?`, `#`, and `%3d` after decoding.

- **CSRF on state-changing endpoints** (`cloud/src/games.ts` PATCH/DELETE) — sessions ride on `SameSite=Lax` cookies with no token check. Lax covers most browsers for top-level navigation, but a top-level form POST from a malicious page can still trigger visibility toggles or deletions in some browser/version combinations. Either add an `X-CSRF-Token` requirement (sourced from the session response) or document the SameSite assumption explicitly.

- **`CF-Connecting-IP` trust** (rate-limit paths in `cloud/src/index.ts:248`, `games.ts:595,969`) — only safe while traffic is exclusively Cloudflare-fronted. Add an assertion (`CF-RAY` presence) so a misconfigured proxy or staging route doesn't silently break per-IP limits.

### Medium — design choices to confirm

- **OnlineID PII strip is shallow** (`cloud/src/games.ts:159` `stripOnlineIds`) — only walks `player_roster`. Other arrays in the share blob are typed `v.unknown()` in the Valibot schema, so future fields could leak `online_id` without tripping a check. Add a comment locking in the assumption, or do a defensive deep walk.

- **OAuth callback read-then-delete** (`cloud/src/auth.ts:248`) — KV doesn't offer atomic compare-and-delete, so two parallel callbacks with the same `oauth_pending` cookie both pass validation; Discord's single-use code blocks the second exchange. Acceptable but worth a comment noting the dependency on Discord as the gate.

- **Owner endpoint `Cache-Control: no-store`** — the SvelteKit side correctly forwards `cache-control` from the worker (`hooks.server.ts:22`) and calls `invalidateAll()` after mutations, but reviewer didn't verify the worker code actually sets `no-store` on owner reads (commit `466e062` claims it does). Worth a quick grep before deploy — this is the kind of thing that regresses quietly.

- **Parser version not enforced server-side** — `parser_version` is set to `\"2.3.1\"` (`src/lib/parser/types.ts:363`) and round-trips through the upload, but the worker doesn't appear to validate it during re-import. Re-import banner relies on client-side semver compare, which is fine for the happy path but not adversary-resistant.

### Low

- **`stripOnlineIds` documentation** — note assumption inline.
- **Compound index `games(user_id, is_public)`** missing in `cloud/migrations/0002_cloud_schema.sql`. Low impact at expected row counts.
- **`CloudHeader.svelte:22`** — active-link `startsWith` is broad on `/games`, fine for UX.
- **fast-xml-parser entity decoding** — auto-decodes `&amp;` etc. Verify Rust does the same; OW XML uses few entities so risk is low.
- **`dump_index` field** must be emitted by both Rust `dump_parsed` and TS `dump.ts` for parity tie-breaking — confirm before relying on the harness output.

---

## What looks solid

- All D1 queries parameterized; no SQL injection surface found.
- PKCE + nanoid(32) session/verifier tokens — sufficient entropy.
- D1 batch chunking respects the 100-param cap (`D1_MAX_PARAMS = 99`); 33×3 game_player_turn math checks out.
- ZIP extraction (`fflate`) mirrors Rust's bomb/size guards (50MB compressed, 100MB uncompressed, 100× ratio).
- Web Worker correctly transfers ArrayBuffer (zero-copy), no Maps/Sets/BigInt cross the postMessage boundary.
- i64 fields parsed via BigInt range-check then emitted as JSON strings — matches Rust semantics, avoids 2^53 loss.
- Parity harness covers all 48 entities with explicit sort keys + `dump_index` tiebreak; no reliance on JS Map insertion order.
- CSP is tight: `hash` mode, no `unsafe-inline` scripts, dev-only `localhost:8787`. Discord CDN scoped to `img-src` for avatars only.
- Dual-build cleanly gated via `__BUILD_TARGET__` Vite define; cloud imports don't bleed into Tauri build.
- Svelte 5 runes used consistently; no leftover `$:` or `export let` in new code.
- `ssr=false` confirmed on `/upload` and `/auth/callback`.

---

## Test coverage gaps

The PR description's checklist is unchecked. Before merge, the parity harness against the smoke corpus and the worker test suite are the two must-runs. The 429-page-on-SSR and OG-unfurler items can only be validated post-deploy.

## Recommendation

Approve with the **High** items addressed (open-redirect tightening + CSRF stance + `CF-Connecting-IP` assertion). Everything else is follow-up safe.

becked:
## High-severity findings — analysis & response

Read each finding against the actual code paths. Two warrant a small change, one I'm pushing back on.

### 1. Open-redirect sanitizer gaps — pushing back

The reviewer's claim that `?x=https://evil.com` and `#https://evil.com` slip past `safeNext` doesn't constitute an open redirect at the actual usage sites. `safeNext("/?x=https://evil.com")` returns `"/?x=https://evil.com"` — when fed to `goto()` or a `Location:` header it stays same-origin. The `evil.com` substring is just a query-param value, not the host.

The hypothesized escalations ("if `next` is ever logged, reflected in a UI…") are different bug classes (log injection, XSS) and belong at the reflection site. Tightening `safeNext` to reject `?` and `#` would also block legitimate post-login destinations like `/games/abc?tab=cities`.

Edge cases I did verify the sanitizer handles correctly:
- `/%2F/evil.com` → decodes to `///evil.com`, caught by the `//` check
- `/%5cevil.com` → decodes to `/\evil.com`, caught by the `\` check
- decode failures → fall back to default

Leaving `safeNext` as-is.

### 2. CSRF on PATCH/DELETE — comment-only

Two reasons the SameSite=Lax + form-POST attack vector doesn't apply here:

- The endpoints are **PATCH** and **DELETE**, not POST. HTML forms only emit GET/POST — there's no top-level navigation that produces a PATCH/DELETE request shape.
- Cross-origin `fetch('PATCH'/'DELETE')` is non-simple → CORS preflight → blocked by the `ALLOWED_ORIGINS` allowlist.

Real CSRF protection is already in place via the method choice + CORS posture. Adding an `X-CSRF-Token` would be belt-and-suspenders without closing a real attack path.

Did add a comment block above `handleGamePatch` documenting the stance so a future loosening of CORS or acceptance of POST mutations triggers a revisit. Commit 42623ee.

### 3. `CF-Connecting-IP` trust — fixed

This was the most legitimate concern, in a narrow way: Workers only receive traffic via Cloudflare's edge, so `CF-Connecting-IP` isn't currently spoofable by upstream clients. But the reviewer's framing ("doesn't silently break per-IP limits") is right — if the topology ever changes (custom domain wired without proxy, staging route, etc.), per-IP buckets would silently collapse onto a shared `null`/`"unknown"` key without any alert.

Added a `getClientIp(request)` helper in `cloud/src/util.ts` that asserts `CF-RAY` presence (Cloudflare always sets it on edge traffic). When absent it logs a warn and returns null — surfacing the misconfiguration in worker logs rather than letting per-IP enforcement degrade silently. Rate-limit call sites in `games.ts` and `index.ts` route through the helper; cache-keyed rate limits fall back to a shared `"untrusted"` bucket so enforcement still applies. Event-logging sites keep raw access since null is fine for the audit row. Commit 42623ee.

---

Recommendation stands: not merge blockers, but #2 and #3 worth the cheap fixes.

becked:
## Medium-severity findings — analysis & response

### 1. `stripOnlineIds` is shallow — fixed (deep walk)

Real concern. Confirmed in `cloud/src/schemas/game.ts`: only `player_roster` is typed strictly; ~19 other fields (`player_history`, `player_nations`, `game_religions`, `current_laws`, `completed_techs`, `map_tiles`, `improvement_data`, etc.) are `v.unknown()` / `v.array(v.unknown())` — opaque parser payloads round-tripped to R2 without field-level validation.

Today's parser doesn't emit `online_id` outside `player_roster`, so nothing leaks. Forward risk: a future parser version that adds `online_id` somewhere new would silently leak it past the roster-only sweep on every anonymous read.

Replaced with a recursive walk that nulls any field literally named `online_id` at any nesting depth. Optimization: returns the input subtree by reference whenever no rewrite happens, so unchanged branches don't allocate — the cost is essentially one tree traversal, not a deep clone. Commit f6c551f.

### 2. OAuth callback read-then-delete race — comment-only

Verified at `cloud/src/auth.ts:248-249`: KV `get` + `delete` is non-atomic. Two parallel callbacks with the same `oauth_pending` cookie can both pass state validation; both attempt the Discord token exchange; Discord rejects the second `code` (single-use authorization codes), so at most one session is ever issued.

Acceptable but the dependency on Discord's behavior is load-bearing. Added a comment block flagging this for future auth provider swaps. Commit f6c551f.

### 3. Owner endpoint `Cache-Control: no-store` — verified, no action

Confirmed at `cloud/src/games.ts:1017-1022`:
```ts
"Cache-Control": isOwner
  ? "private, no-store"
  : "public, max-age=3600, s-maxage=3600",
```
With `Vary: Cookie, Origin` so the cache key changes by session. Behavior matches commit `466e062`.

### 4. Parser version not enforced server-side — pushing back, reviewer wrong

Reviewer's claim that the worker doesn't validate `parser_version` server-side is contradicted by the code:

- `cloud/src/schemas/game.ts:130-136`: `FullGameDataSchema` validates `parser_version` against a `KNOWN_PARSER_VERSIONS` Set on every upload via valibot. Unknown values rejected at request parse time before any DB write.
- `cloud/src/games.ts:743`: re-import path runs `compareSemver(blob.parser_version, existing.parser_version)` server-side and returns 409 if `cmp <= 0`. The gate is server-enforced, not banner-enforced — the UI banner is just a hint surfaced from the same comparison.

A client lying about its parser_version either gets rejected at the schema layer (unknown version) or gets a 409 if it's not strictly newer. No action.

---

That's all four mediums addressed. Moving on to the lows next.

becked:
## Low-severity findings — analysis & response

No code changes needed. Three already addressed or verified, two are reviewer mistakes.

### 1. `stripOnlineIds` documentation — already done

The medium-severity deep-walk fix (commit f6c551f) included an inline comment block explaining the assumption ("most of FullGameData is typed as `v.unknown()`… today only `player_roster` carries `online_id`…"). Same code site. Done.

### 2. Compound index `games(user_id, is_public)` — pushing back, no query needs it

Grepped every `FROM games` query in the worker. Access patterns:
- `WHERE user_id = ?` (4 sites: stats counts, library list, dedup) — covered by `idx_games_user`
- `WHERE game_id = ?` (5 sites: ownership checks, PATCH, DELETE, download) — covered by primary key
- `WHERE user_id = ? AND file_hash = ?` (1 site: dedup) — covered by the `UNIQUE (user_id, file_hash)` constraint

No query filters on `(user_id, is_public)` together. The existing `idx_games_public WHERE is_public = TRUE` partial index handles the only `is_public`-keyed access pattern (anonymous public reads via game_id). A compound index would be dead weight on writes with no read benefit.

### 3. `CloudHeader.svelte:22` startsWith — correctly scoped, not "broad"

```ts
return path === href || path.startsWith(href + "/");
```

The `+ "/"` separator excludes prefix collisions like `/games-settings` or `/gamesfoo`. `/games/[id]` activating the Games tab is the intended behavior. No action.

### 4. fast-xml-parser entity decoding — verified, parity-tested

Both parsers auto-decode XML entities:
- Rust: `roxmltree` 0.19 — XML 1.0 conformant, decodes `&amp;`/`&lt;`/`&gt;`/`&quot;`/`&apos;`
- TS: `fast-xml-parser` with default entity processing (no `processEntities: false` override in `parse-xml.ts:40-50`)

Real saves contain entities — `match_426504725_05-moroten-droner.zip` has 458 occurrences of `&gt;`/`&lt;`. The Rust↔TS parity harness validates byte-identical output across the corpus, which transitively validates decoding. No action.

### 5. `dump_index` field parity — verified, both sides emit it

- Rust: `src-tauri/src/bin/dump_parsed.rs:114` injects `dump_index` into every row
- TS: `scripts/parity/dump.ts:321` does the same via `rows.map((row, i) => ({ ...row, dump_index: i }))`
- Sort tiebreak: `scripts/parity/sort.ts:48-49`
- Correctly excluded from diff comparison: `scripts/parity/diff.ts:160-161`

Both sides emit it. No action.

PR #79, opened 2026-06-30
feat(game-detail): head-to-head matchup view on the Military tab

alcaras:
## What

Rebuilds the **Military tab** into an owglick-style head-to-head *matchup* view for two-player games. It's computed entirely from the existing game blob — **no parser or Worker changes** (`units` and tech-discovery data were already shipped; this just surfaces them). Everything degrades gracefully when data is missing (older blobs, the legacy `web/` share viewer), so it's safe to ship frontend-first.

### Military-power chart + per-nation event rail

Gradient area-filled power curves, with a timeline bar **per nation** below the plot (uploader first). Each nation has rows for:

- **Leaders** — archetype icon; hover shows name / cognomen / archetype + the four character ratings with the in-game stat icons.
- **Laws** — every adoption/swap, with the law's icon + running count (swap-aware).
- **Unit-tech unlocks** — techs that field a 5/6/8/10-strength unit (naval-gated, so boat techs only show when a sea unit was actually built).

Hovering any rail icon drops a guide line on the chart at that turn.

Military tab — chart + per-nation event rail

Hover — guide line + leader tooltip with rating icons

### Head-to-head build comparison

Military Built and Ending Army as center-split diverging bars (shared → A-only → B-only), each with army-composition donuts. "Military Built" also shows total **military power built** (Σ units built × strength × 10).

Head-to-head build comparison with donuts

### Also

- A `unit-stats` table (unit → strength / tech-prereq / naval) baked from `Reference/XML` via `npm run bake:unit-stats` (committed snapshot, like the other generated tables).
- The active game-detail tab is deep-linked through the URL hash, so a reload / shared link restores the tab instead of falling back to Overview.

## Notes

- Frontend-only — no migrations, no Worker changes, no deploy ordering.
- The H2H comparison + event bars are gated to two-player games; FFA keeps the existing N-player pies + pivot table.
- `npm run check` and lint are clean.
- The screenshots above are hosted on a side branch and are **not** part of this PR's diff.


becked:
Review findings, grouped by severity. Line numbers are against the head of `feat/stat-views`.

## Correctness

**1. Event rail isn't gated to two-player games — it must be.**
The per-nation event rail (`MilitaryTab.svelte:771`) is gated solely by `{#if militaryChartOption}` and then iterates `{#each orderedPlayers}` with no `matchup` check — unlike the H2H block at `:861` (`&& matchup`) and the standalone donuts at `:844` (`&& !matchup`). So in a 3+ player FFA it renders one leader/law/tech bar per nation, which the matchup view isn't built for and doesn't hold up visually beyond two sides. Gate the rail to two-player games with the same `&& matchup` condition as the H2H block (the standalone donuts + pivot table remain the N-player view). The PR description's "event bars are gated to two-player" claim then becomes true — but the fix is in the code, not the description; N-player rendering isn't an acceptable resolution here.

**2. Naval tech markers gate on *any* player's sea unit, not the player's own.**
`navalRelevant` is `unitsProduced.some((u) => UNIT_STATS[u.unit_type]?.naval)` — whole-game — but it's applied per-player inside `eventRail` (`navalRelevant || !s.naval`). If player A builds a boat, player B's Bireme/Trireme/Dromon tech-unlock markers show up too, even though B never touched the water. That contradicts the comment "boat techs only show when a sea unit was actually built." Compute relevance from the specific player's production (filter `unitsProduced` to that player before `.some`).

**3. "Ending Army" mixes a fog-limited count with a complete stat.**
`endingArmy()` counts units from the `units` roster, which the parser builds by walking `<Tile>` → `<Unit>` (`src/lib/parser/parsers/units.ts`) — i.e. only units visible on the uploader's map. The opponent's units under fog aren't in the save, so their count is undercounted, sometimes badly. Meanwhile `endingPower()` reads the complete `military_power` history stat. The same sub-line can produce readings like "3 alive · 1,200 power", and when `units` is absent on an older blob the block renders "0 v 0 alive" next to real power numbers. "Military Built" doesn't have this problem (both sides come from complete production aggregates). Options: source the ending-army count from a complete per-player stat rather than the visible roster, or drop/caveat the "alive" count for the non-uploader side.

## Conventions / new patterns

**4. `?? "#888888"` deviates from the per-player series-color pattern.**
Three sites — `MilitaryTab.svelte:360`, `:537`, `:780`. The per-player game-detail tabs pass the possibly-undefined color straight through for ECharts series and let the `CHART_THEME` palette fill in (`EventsTab.svelte:91`, `LawsTab.svelte:148`, `LeadersTab.svelte:158`, `TechsTab.svelte:120`, and the original MilitaryTab); a gray literal isn't used for this. So line 537 (series) should be just `rp?.color`. Lines 360 and 780 need a concrete string (they feed `toRgba`/a CSS value), so they do need a fallback — but the in-repo way to get a guaranteed series color is `getNationChartColor(player.nation, i)` rather than a gray. (`rp` is the resolved `DetailPlayer`, whose `.color` is never undefined anyway, so the gray is effectively dead code.) Note: hardcoded color fallbacks do exist elsewhere for non-series UI (`config/terrain.ts:68`, `users/OverviewTab.svelte:172–174`), so this is about matching the series-color pattern, not a blanket rule.

**5. Color+alpha is now done three different ways.**
There's an existing idiom for translucent ECharts fills — pass a solid color plus ECharts' `opacity` field: `areaStyle: { color: getChartColor(0), opacity: 0.18 }` (`stats/charts/yields.ts:73,99`). The power-chart gradient (`MilitaryTab.svelte:548–557`) instead pre-bakes alpha into `colorStops` via a new `toRgba()` helper. That divergence is justified — `areaStyle.opacity` only produces a flat fill and can't do the top-to-bottom fade the gradient needs, so per-stop rgba is genuinely required there. The unjustified part is `BuildComparison.svelte:135,145`, which is plain CSS (`linear-gradient(90deg,{ca}30,{ca})`) using hex-alpha string concat — a third mechanism where `toRgba(ca, 0.19)` would slot in directly. It also silently assumes 6-digit hex and breaks to invalid CSS if a color ever arrives as `rgb(...)`. Consolidate the CSS cases on `toRgba`; and note `toRgba` itself re-implements the `hexToRgb` logic already in `SpriteMap.svelte:598`, so a shared util would be better than a second copy.

**6. Two new ad-hoc tooltip grays (minor consistency nit).**
The new tooltip builders introduce `#cfc9bd` and `#9b948a` for secondary text. Hardcoded hex inside chart/tooltip option objects is the existing norm in this codebase (`Chart.svelte:7`, `SpecialistsTab.svelte:201`, `users/OverviewTab.svelte:175`, and `#666`/`#D2B48C` already in this file pre-PR), so this isn't a convention violation — but the two new shades don't match any gray already in use. Worth reusing an existing chart-text gray (or defining one constant) instead of adding two more one-off values.

## Cleanup / robustness

**7. Dead `combat` parameter + stale comment.**
`builtUnits(p, combat)` is only ever called with `combat = true` (`:159`–`160`), so the civilian branch is unreachable, and the `:81` comment still advertises a "Civilian Built" block that isn't rendered. Drop the parameter and the comment, or wire up the civilian block if it was intended.

**8. `bake-unit-stats.ts` can mistype multi-prereq units.**
`TechPrereq` is typed/handled as a single `string`. A unit with multiple `<TechPrereq>` elements parses as an array under fast-xml-parser and would serialize wrong / break the `string | null` shape. No such unit exists in the current XML, so this only bites on a future `Reference/XML` refresh — worth a guard (take the first, or assert single) so the bake fails loudly instead of emitting bad data.

alcaras:
Thanks for the thorough pass — all eight are addressed in `de11ee9`. Quick notes per finding:

**1.** Gated the event rail with `{#if matchup}` (donuts + pivot stay the N-player view).
**2.** Replaced whole-game `navalRelevant` with `navalRelevantFor(playerId)`, filtering `unitsProduced` to that player before `.some`.
**4.** Series `itemStyle` is now just `rp?.color`; the concrete gradient-fill color uses `getNationChartColor(player.nation, i)`. Gray literals gone.
**5.** Extracted a shared `toRgba()` to `src/lib/utils/color.ts` and switched the `BuildComparison` CSS gradients onto it (the `{ca}30` hex concat is gone, so `rgb(...)` inputs no longer break).
**6.** Hoisted the two tooltip greys to `TOOLTIP_TEXT` / `TOOLTIP_MUTED` constants.
**7.** Dropped the dead `combat` param and the stale "Civilian Built" comment.
**8.** `TechPrereq` is now `string | string[]` and the baker takes the first prereq, so a multi-prereq unit on a future Reference refresh won't emit bad data.

**3 — I dug into this one and I don't think the roster is fog-limited.** The `units` roster isn't a visibility snapshot — an Old World save is the *full* authoritative game state (it has to be: Play-By-Cloud passes the save around as the shared game state, so every player's units must be in it). Fog of war is a *separate* per-tile `<Revealed*>` mask layered on top at render time, not omission of units from the save.

I verified against a real 1v1 PBC save (`V25NUHB9…`):
- units are tile children keyed by a `Player="N"` attribute, and **all** players' units are present: player 0 → 22 units, player 1 → 41, tribes (`Player="-1"`) → 53;
- the save carries ~4,300 `<Revealed*>` entries — the fog mask, stored independently of the units;
- note the opponent had *more* units than the uploader (41 vs 22), which couldn't happen if the roster were limited to what the uploader can see.

So the "3 alive · 1,200 power" mismatch you'd get from a fog-limited count doesn't actually arise — both sides' counts are complete. The one real failure mode is an **older blob that has no `units` roster at all**, which would render "0 v 0 alive" next to real power. I kept that guard: the Ending Army block now only renders when `units.length > 0` (comment in `buildBlocks` explains the full-game-state rationale). If you've seen a save where the roster genuinely undercounts a fogged opponent, I'd love a repro — but everything I can find says the roster is complete.

becked:
Pushed a couple of follow-up commits to `feat/stat-views`:

**Event-rail positioning** — the rail markers now take their x-position from the live chart via `convertToPixel(turn)` instead of fixed pixel insets. This removes the hardcoded grid insets, `containLabel: false`, and the `EVENT_RAIL_*` constants that had to be kept in sync with `grid.left`/`grid.right`, so the rail stays aligned to the plot on resize with no magic numbers. `Chart.svelte` gains two optional, backward-compatible props (`onReady` / `onLayout`) to expose the ECharts instance and signal re-layout — nothing else that uses `Chart` is affected.

**Chart cleanup** — the power chart is now a single value-axis plot; the "Turn" / "Military Power" axis titles are removed and the grid is centered in the panel. Rail markers render through the shared `SpriteIcon` component, matching the other tabs.

**Asset cleanup** — removed `static/laws-icon.png`; it was committed but never referenced anywhere.

Two notes:

- The law vs archetype rail icons can read as slightly different sizes. That's a pre-existing sprite-baking inconsistency (some sprites are trimmed to content, others are a padded canvas), not specific to this view — filed separately as #85.
- The matchup power chart now renders via `Chart` directly rather than `ChartContainer`, so it no longer has the fullscreen-expand button (the FFA view still does). Easy to wire back if you'd like it.

PR #81, opened 2026-06-30
feat(tournament): match parts — per-sitting schedule, casters, VODs (1/4)

alcaras:
Match **parts**: a match is one game played across one or more sittings, each with its own scheduled time, ordered caster list (streamer first), and VOD links. This is **PR 1 of 4** from the previous larger branch, per review — match numbers, admin copy tools, and caster self-service follow as separate PRs once this lands.

## Data model

- Migration **0029**: `parts TEXT` (JSON list) + `parts_rev INTEGER` on `tournament_matches`; the flat 0025 schedule columns are backfilled into a single part and left dead (D1 is forward-only). `loadMatch`/`loadMatches` select explicit columns so responses never re-emit them.
- Part shape: `{id, scheduled_at, casters: [{user_id, name}], vods: [{url, label}]}` — linked casters snapshot the canonical username (batch-resolved, mirroring the slot-prelink pattern); VOD hosts allowlisted to youtube/twitch, re-checked on read by the lenient `parseParts` (shared `isSafeHttpUrl` with `parseLinks`).

## Review items addressed

- **Concurrency**: every write bumps `parts_rev` and conditions on the value it read; the editor echoes `expected_rev` and a stale save gets **409 CONFLICT** (surfaced as a reload toast) instead of silently erasing a concurrent writer.
- **Decided matches are admin-only**: pending → admin or either participant; complete/forfeit → admin only (a participant can no longer wipe the VOD/caster archive with `parts: []`); byes rejected.
- **Overdue rows keep their timestamp**: an in-progress match (time passed, unreported) shows its last-started sitting's time + "· in progress" in the table, and sorts by it.
- **Write-response contract**: `patchMatchSchedule` is typed `Promise<void>` — callers refresh via `invalidateAll`; the body is documented as the raw-row shape, not the GET serialization.
- Cleanup: dead `matchVods` API dropped; status-chip label map shared (`MATCH_STATUS_LABEL` in `parts.ts`); `isSafeHttpUrl` extracted; rules doc updated (no stale `/cast` reference).

## Reconciled with main since the review

Built on current main: keeps your `isDifferentYear` formatters and current-month calendar seeding (8b703e8), the reworked Match column + casted/uncasted filter (fe26995 — `casted` = any part has casters; the Stream column becomes **VODs**, first VOD across parts), and the nulls-last sort property (cf141f1) — timed matches now sort ascending (live/overdue → soonest upcoming) with unscheduled/completed pinned last in both directions. The status filter gains **In progress**.

## UI

- **Schedule editor** (match popover): add/remove sittings; per-part date+time entered in the viewer's timezone (stored UTC, DST-aware), casters (streamer + co-casters, linked or free-text), VOD links. Reads "VODs" on decided matches.
- **Match popover**: per-part display with countdown ("in 2 days") and VOD links.
- **Bracket cards**: status chip (Scheduled / In progress / Completed) from the shared 4-way display status.
- **Matches page**: calendar renders one chip per sitting ("· Pt N"); list keyed by match+part (backfilled parts share the id `p1`).

## Deploy notes

- Worker first, frontend second (frontend sends `parts`/`expected_rev`).
- `0029` rehearsed via the migration test-suite path; the backfill only touches rows with any 0025 metadata.

Tests: **387 pass** — parts round-trip, caster resolution, policy gates, rev conflict, VOD host validation, parseParts sanitization.


becked (changes requested):
Thanks for this — the feature set is exactly what tournament ops needs, the parts model is a sound shape for split matches, and the test coverage on the worker side is real. Review is against the current head (0cf35b0). Two structural asks up front, then file-level items.

## Please split this into separate PRs

The PR body already describes the stacking; the seams are clean and each layer has a different risk profile:

1. **Parts model core** — migration 0029, worker schema/handlers, the parts editor, popover display. This is the deploy-sensitive piece (worker-first ordering, backfill) and it's in the best shape; it can land and settle first.
2. **Match numbers** — frontend-only today, but it needs a design decision (below) that shouldn't hold the parts core hostage — and its likely resolution touches the schema anyway.
3. **Admin copy tools** (DM / thread title / sesh export) — small, frontend-only, depends on 1+2.
4. **Caster self-service** — the cast/uncast endpoints + Cast view, depends on 1.

## Match numbering shouldn't be client-side (blocking design question)

`src/lib/tournament/match-numbers.ts` computes numbers per render by sorting phase → round → division and filtering byes. The header comment promises "Match 12 must never change," but nothing enforces that, and there are two live renumbering paths:

- **Divisions advance independently.** `maybeAdvanceAfterMatchReport` closes only the reporting division's round, so Division A can be on round 5 while Division B generates round 4. B's new matches sort *before* A's round-5 matches (round 4 < round 5), shifting every number after them — the "Match 34" an admin already pinged players with now points at a different game on every surface, including thread titles created from the old number.
- **Retro-bye edits.** `PatchMatchSchema` allows flipping an existing match's `status` to `"bye"`; the match drops out of the numbering and everything after shifts down by one.

Since the entire value of the feature is a stable handle admins can paste into Discord, the number needs to be persisted server-side — e.g. a `match_number` column assigned monotonically per tournament at round-generation time (byes skipped), which makes append-only true by construction instead of emergent. If you'd rather not do that yet, the UI/docs should stop promising stability. Either way this is why I'd pull it out of this PR: landing the client-derived version now means renumbering live data later.

## Remaining issues on the parts core

- **The admin schedule PATCH still races caster self-service.** `handlePatchMatchSchedule` writes `UPDATE tournament_matches SET parts = ? WHERE match_id = ?` (admin.ts ~2067) from a popover snapshot that's deliberately seeded once at open and held for the whole edit session. A caster's "I'll cast" landing mid-edit is silently erased on Save. The CAS engine you added in `player.ts` (`mutateMyCasterEntry`) is the right shape — the admin path is now the only unguarded writer of `parts` and needs the same conditional write (or at minimum conflict detection with a reload prompt).
- **Either participant can wipe a decided match's parts.** The endpoint accepts any non-bye match, the schema has no `minLength` on `parts`, and authz admits either participant — so a losing player can PATCH `parts: []` on a completed match and destroy every VOD/caster attached after the game, with no recovery (the audit event records only the match id). The rules doc now documents this as intended, so treating it as a policy question: I'd gate decided matches to admins for destructive edits (or restrict participants to additive changes there).
- **Casters can't remove themselves once a match is reported.** Self-service is pending-only (`MATCH_NOT_PENDING`), while the admin PATCH deliberately supports editing decided matches for exactly this after-the-fact case. A caster who signed up but never actually cast is stuck credited, and only an admin replace-all can fix it. Self-removal (at least) should work on decided matches.
- **Cast/uncast respond with the raw row typed as `TournamentMatch`.** `api-cloud.ts` declares slot display names and per-caster `display_name`/`avatar_url` on the return type, but the handlers return the un-serialized match row (`{ ...match, parts }`), which has none of that. Latent today (the Cast view discards the body), but the typed contract invites someone to use it — return `serializeMatch` output or narrow the declared type.

## Smaller items

- `docs/tournament-rules.md` (§ Scheduling) still references `/tournaments/[slug]/cast`, which no longer exists after folding Cast into the Matches page — the PR description advertises the route too.
- `scripts/bake-map-caveats.ts` isn't registered in `package.json` (`bake:map-caveats`) or `bake:all`, unlike every other baker — an atlas refresh won't re-run it.
- An overdue pending match (scheduled time passed, result unreported) now sorts correctly but its table cell shows only "In progress" with no timestamp — admins chasing overdue reports lose the when; consider showing the last-started part's time.
- `handleCastMatchPart` resolves the session twice (`sessionFromRequest` in the handler plus `authedSession` inside the engine) and runs the username SELECT before the rate-limit gate, so 429'd calls still pay it.

## Duplication / consistency cleanup (non-blocking, but please take a pass)

- `matchVods`/`NumberedVod` in `parts.ts` are exported but never imported anywhere — dead speculative API; drop them until a consumer exists.
- `bracketLabel` in `CastView.svelte` is a copy of `phaseLabel` in the matches page, and `MatchPopover` inlines the same championship/division ternary a third time — one shared helper in `$lib/tournament`.
- The `STATUS_LABEL` map + status-chip CSS are duplicated between `ChampionshipBracketTree` and `SwissFlowBracket` (each with a comment pointing at the other as "the shared palette") — a tiny shared chip component or a label/class map next to `MatchDisplayStatus` in `parts.ts`.
- `parseParts` re-inlines `parseLinks`' http(s)-protocol check in the same file (`data.ts`) — extract a shared `isSafeHttpUrl`, especially since this is the security-relevant path for rendered VOD links.
- The "upcoming scheduled parts" pipeline is built twice — `CastView` (with a 2h grace) and `seshExport` (without) — so the two surfaces already disagree about what "upcoming" means; one helper in `parts.ts` with an explicit grace parameter.
- The copy-button feedback (`$state` flag + 1500 ms `setTimeout` + "Copied!" swap) is hand-rolled three times (thread title, DM, sesh export) — fold it into `clipboard.ts` or a small `CopyButton`.
- `MAX_CASTERS_PER_PART` in `player.ts` duplicates the schema's `v.maxLength(10)` as an unlinked constant — single-source the limit.
- List endpoints parse every match's `parts` JSON twice per request (`loadUserIdentitiesForMatches` and again in `serializeMatch`) — parse once and thread the result through.
- The scheduled-column `sortValue` recomputes `nextScheduledAt`/`Math.max`-over-parts for both operands of every comparison, inside a `$derived` that re-runs per search keystroke — precompute a numeric key per match before sorting.

Happy to re-review the split PRs quickly — the parts core in particular is close.

alcaras:
Thanks for the thorough review — all taken. This PR is now **just the parts core** (single commit, rebased on current main), with your parts-core findings addressed in-branch:

- **Admin PATCH vs caster race** → a `parts_rev` column on 0029 (still unshipped, so the migration just grew it): every writer bumps it and conditions its UPDATE on the value it read; the editor echoes `expected_rev` and a stale save 409s with a reload toast instead of silently erasing anyone.
- **Participants wiping decided matches** → decided = admin-only; pending stays admin-or-participant; test added.
- **Write-response contract** → `patchMatchSchedule` typed `Promise<void>` with the body documented as raw-row shape; callers refresh via `invalidateAll`.
- **Overdue rows** → in-progress table rows show the last-started sitting's time (and sort by it).
- Cleanup pass: dead `matchVods` dropped, shared status-chip map, shared `isSafeHttpUrl`, stale `/cast` doc reference gone.

Also reconciled with your post-review main: your `isDifferentYear` formatters + current-month calendar seed are kept as-is; the Match column + casted/uncasted filter translate to parts (`casted` = any part has casters; Stream → VODs); your nulls-last sort property is preserved with timed matches ascending (live/overdue → soonest).

**The rest of the split, coming as follow-ups in order:**
2. Match numbers — agreed on persisted `match_number`; next PR adds the column, assigns at round-generation (byes skipped, monotonic per tournament), backfills in current display order, and the frontend reads the server value.
3. Admin copy tools (DM/thread title/sesh) — with the caster-thread link resolved from tournament Links, `bake:map-caveats` registered in package.json, and the copy-feedback helper deduped.
4. Caster self-service — with self-removal allowed on decided matches, the single-sourced caster cap, the session double-resolve fixed, and the rate-limit gate before the username SELECT.

alcaras:
The rest of the split is up, stacked in order: #87 (persisted match numbers), #88 (admin copy tools), #89 (caster self-service + shareable filter links). Each is one commit on its predecessor — their diffs collapse as the stack merges.

PR #87, opened 2026-07-02
feat(tournament): persisted global match numbers (2/4)

alcaras:
**PR 2 of 4** — stacked on #81 (parts core). Until #81 merges this diff shows both commits; **review only the top commit** (`feat(tournament): persisted global match numbers`).

Resolves the blocking design question from the #81 review: match numbers are now **server-assigned and persisted**, not derived client-side.

## How

- Migration **0030**: `match_number INTEGER` on `tournament_matches`.
- **Assignment at round-generation** by all three insert sites (swiss round builder, championship transition, championship rounds 2+) via a `MAX(match_number)+1` subquery per tournament, inside the D1 batch transaction — dense, race-free, and **append-only by construction**. Byes stay `NULL` (never numbered).
- **Backfill** numbers existing matches per tournament in exactly the order the old client-side numbering displayed (phase → round → division → position), so any "Match 12" an admin has already pasted into Discord keeps pointing at the same game.
- Serializers expose it; the client's `matchNumbers()` is now just a field-collection into the id→number map its display sites consume (bracket-card badge, "· Match N" in the popover header).

## Why the two renumbering paths from the review can't recur

- *Divisions advancing independently*: a later-generated Division B round takes the next numbers **after** everything already assigned — it can't sort itself in front of Division A's matches anymore.
- *Retro-bye edits*: the number is a stored column, so flipping a match's status doesn't shift anyone else. Test pins exactly this: after a retro-bye, every other match keeps its exact number.

Tests: sequential assignment skipping byes; retro-bye stability; monotonic append across swiss→championship (390 total).


becked:
## Design follow-ups on the match-numbers work — remaining items

Notes from a review pass so we can resume cleanly in a fresh session. **Everything below is design/fit, not functionality** — the feature is green (frontend lint + `svelte-check` 0/0, cloud `tsc --noEmit`, **390/390** cloud tests, clean virtual merge into `main`). Branch is `feat/match-numbers-server`; leave in **draft**, not ready to merge.

### Already done & pushed on this branch
- **`7b17f9c`** — Popover now renders `· Match N` in the header. The original `matchNumber` prop was dead: it was declared + imported (`padMatchNumber`) but never rendered, and only **1 of the 3** `MatchPopover` call sites (`[slug]/+page.svelte`, not the matches-list page or `TournamentUpNextPanel`) passed it. Fixed by reading `match.match_number` off the row directly (it's on every `TournamentMatch`), which lights up all three sites and drops the prop. Same commit also: collapsed the `.match-num` `font-family` to the terse `ui-monospace, monospace` used by sibling rules in `SwissFlowBracket`, and hoisted `const status` in `buildSwissRoundStatements` (was binding `isBye ? "bye" : "pending"` twice).
- **`9ace4a0`** — Dropped the `matchNumbers()` helper + the `matchNumberById` derived map/prop entirely. The badge (`SwissFlowBracket`) and popover (`MatchPopover`) both read `match.match_number` straight off the row now, so the id→number map was pure indirection over a field already on each row. `src/lib/tournament/match-numbers.ts` is now just `padMatchNumber()`.

### Remaining — worth doing
**1. DRY: the `MAX(match_number)+1` subquery is copy-pasted into all three insert sites.**
`cloud/src/tournament/admin.ts` — `handleTransitionChampionship` (~L2577), `buildSwissRoundStatements` (~L2748), `buildChampionshipRoundStatements` (~L2807). The correlated subquery `(SELECT COALESCE(MAX(m2.match_number),0)+1 FROM tournament_matches m2 JOIN tournament_rounds r2 ON r2.round_id=m2.round_id WHERE r2.tournament_id=?)` appears verbatim 3×, and two of them wrap it in an identical `CASE WHEN ? = 'bye' THEN NULL ELSE … END`. The championship-rounds site deliberately omits the bye guard (assumes championship rounds never bye) — which is exactly the kind of silent divergence duplication invites. **Fix:** hoist a shared SQL-fragment constant (+ a bye-guarded variant) and interpolate into the three INSERTs, mirroring how `MATCH_COLUMNS` is built in `cloud/src/tournament/data.ts`. Low-risk; safe to just implement.

**2. Consistency: the "Match N" badge is on Swiss cards but not championship cards.**
`SwissFlowBracket.svelte` (~L377) renders the `.match-num` badge from `m.match_number`; `ChampionshipBracketTree.svelte`'s `matchBody` snippet (~L406) renders the same-shaped `class="match"` card but no badge — even though championship matches *are* numbered (their popover header shows it). The `PositionedMatch` view model that `matchBody` consumes doesn't even carry `match_number`, so this is a structural gap, not just missing markup. This is the **same "uneven coverage across sibling surfaces"** class as the popover-prop bug above. **Decision needed:** thread `match_number` into `PositionedMatch` and render the badge in `matchBody` for parity, or consciously keep the badge Swiss-only (and note why).

### Remaining — minor / decisions
**3. Narrow `padMatchNumber`'s signature.**
`src/lib/tournament/match-numbers.ts` — `padMatchNumber(n: number | undefined | null)`; both callers already guard `{#if match.match_number != null}` before calling, so the null branch is dead and the 3-wide signature advertises tolerance nothing uses. Narrow to `(n: number): string`. Trivial.

**4. Padding decision in the popover header.**
`MatchPopover.svelte` (~L620–626) renders `· Round 2 · Match 007` — a raw `round_number` next to a zero-padded, 3-wide match number. The padding earns its place on the column-aligned card badges but reads inconsistently in flowing header text. **Decision:** pad everywhere (keep as-is) or use the raw number in the popover header and reserve the padded/monospace form for the badge.

### Resume
`git checkout feat/match-numbers-server` (this branch, pushed to the fork). **#1 and #3** are safe to implement directly; **#2 and #4** want a design/aesthetic call first.

becked:
## Session update — the four review follow-ups are resolved (+ a seed-tool fix)

Continuing from the [design follow-ups note](https://github.com/becked/per-ankh/pull/87#issuecomment-4871136032) above. All four items are done, plus one adjacent fix that surfaced while building a verification fixture. Head is now `821ec9f`, and the PR is marked ready for review.

**Two items needed a design call — decisions made:**

- **Championship badge (item 2):** added for parity — championship bracket cards now render the same "Match N" badge Swiss cards already had. Byes and synthesized future-round placeholders carry `null` and show no badge.
- **Popover header padding (item 4):** raw (unpadded) number in the flowing header (`· Round 2 · Match 7`) — it read oddly next to the unpadded round number, and the handle admins paste into Discord is unpadded anyway. The zero-padded/monospace form stays on the column-aligned card badges, where the fixed width earns its place.

### Commits

- `9555b57` **refactor** — DRY the `MAX(match_number)+1` subquery: it was copy-pasted verbatim into all three round-generation insert sites (two bye-guarded, one bare). Hoisted into shared `NEXT_MATCH_NUMBER_SQL` / `NEXT_MATCH_NUMBER_OR_NULL_FOR_BYE_SQL` constants. No behavior change. *(note item 1)*
- `166c942` **feat** — Match N badge on championship cards: threaded `match_number` through `BracketMatch` and `PositionedMatch`, rendered in `matchBody` reusing the Swiss styling. *(note item 2)*
- `a748af8` **refactor** — popover header shows the raw number; with no caller passing `null` anymore, `padMatchNumber` is narrowed from `(n: number | undefined | null)` to `(n: number)` and the dead null branch dropped. *(note items 3 + 4)*
- `821ec9f` **fix** — the seed planner/CLI never set `match_number`, so `./per-ankh admin tournament seed` produced fixtures with all-NULL numbers (badges invisible in local dev). Assigned in a final canonical-order pass (phase → round → division → within-round index; dense; byes `null`) mirroring migration 0030's backfill, and emitted from the CLI writer. *(new — pre-existing tooling gap, unrelated to the Worker path)*

### Verification

- cloud `tsc --noEmit`: clean
- cloud `vitest`: **391/391** (was 390 + a new test pinning seed-numbering density, bye-nulling, and append-only order across the swiss→championship boundary)
- frontend `svelte-check`: 0 errors / 0 warnings
- frontend `eslint src`: clean
- End-to-end: re-seeded a `mid-championship` fixture through the fixed CLI → 36 non-bye matches numbered densely 1–36, 2 byes NULL, championship cards numbered 33–36.

### Note for reviewers

The PR description still says "review only the top commit" (from when this was a single commit stacked on PR #81). It now carries five commits of its own — the original persisted-match-numbers change plus the four above.

PR #89, opened 2026-07-02
feat(tournament): caster self-service + Cast view with shareable filters (4/4)

alcaras:
**PR 4 of 4** — stacked on #88 → #87 → #81. Until those merge this diff shows the whole stack; **review only the top commit** (`feat(tournament): caster self-service + Cast view with shareable filters`).

## Cast view

A third view mode on the Matches page (List / Calendar / **Cast**): upcoming scheduled sittings soonest-first with countdowns, match number, players, division, and the atlas-linked map; flagged **"needs a caster"** with an "I'll cast" button. "Needs a caster / All scheduled" filter (needs-only by default). Uses the same `upcomingScheduledParts` helper as the sesh export, with a 2h grace so a just-started match is still claimable.

## Self-service endpoints

`POST`/`DELETE` `…/matches/:id/parts/:part_id/casters/me` — open to any logged-in user, scoped to the caller's own entry (never replace-all):

- Add/move me: streamer takes index 0 (bumping the current streamer to co-caster), else co-caster; omitted role → streamer if the part is empty.
- Drop me — **also allowed on decided matches** (per review: a caster who signed up but never actually cast shouldn't stay credited). New casting stays pending-only; byes rejected.

Endpoint fixes from the #81 review, all in: mutations **CAS on `parts_rev`** with retry (concurrent sign-ups are the expected case); the **rate-limit gate runs before any per-request work**; the **session resolves once**; `MAX_CASTERS_PER_PART` is **single-sourced** from the schema; both endpoints respond **204** (no raw-row body masquerading as the GET shape — clients refresh via `invalidateAll`). `tournament_schedule` ledger events move to the 24h retention bucket per the events rule (they're metadata-free counters read over a 1h window).

## Shareable filter links

The Matches page's view/zone/status/caster filters are URL-addressable (`?view=cast`, `?caster=uncasted`, `?status=scheduled`, `?zone=local`; defaults omitted; `replaceState`, not push) — so an admin can point people straight at "scheduled matches that still need a caster".

6 new integration tests (streamer/co-caster ordering, self-promotion bump, drop-only-self, 401/404, decided-match policy); **397 worker tests** green on the stack tip.


becked:
Pulled this down, rebased onto current `main` (single commit — dropped the already-merged #87/#88 dups the branch was still carrying), then reviewed it for app-fit and fixed the deploy blockers.

### Fixed
- **Deploy-blocking lint** in the matches-page filter URL-sync (the branch predated a lint-config tightening on `main`, so `npm run lint` was failing):
  - `svelte/prefer-svelte-reactivity` — the `$effect` built a mutable `URLSearchParams`; now builds the query from plain string parts.
  - `svelte/no-navigation-without-resolve` — annotated the deliberate shallow `replaceState` (dynamic filter query string; `resolve()`'s branded types don't admit it), matching the precedent in `upload/+page.svelte`.
- **Four dead `eslint-disable no-unused-vars` directives** in `player.ts` — type-position param names are never flagged; ESLint reported all four as unused, and they were the only such directives in `cloud/src`.
- Verified: `npm run lint` clean, `svelte-check` 0/0, worker suite **398/398**.

### Still open (picking up in a follow-up)
- **URL-sync primitive.** Shallow `replaceState` from `$app/navigation` is new to this app — everything else navigates via `goto(resolve(...), { replaceState: true })`. It's arguably the *right* primitive for pure client-side filter state (no `load` re-run), but it's a new pattern. Decide whether to adopt shallow routing app-wide or conform. Kept + annotated for now.
- **`mutateMyCasterEntry`** callback-injection (the `prepare`/`mutate` closures) diverges from the flat, linear handler style in `admin.ts`. Consider named callback type aliases or flattening into two handlers over a small shared CAS helper.
- **Matchup formatting is duplicated** — `CastView`'s `vs`, the page's `shortMatchup`, `seshText`'s inner `vs`, and `matches-table` each rebuild "A v B" with a *different* missing-side fallback (`?` / `—` / `Bye`). Consolidate into one shared helper in `match-occupant.ts`.
- **`bracketLabel`** (championship / division-name mapping) is re-derived in `CastView` and re-inlined across several tournament components — extract a shared helper.
- **`tournament_schedule` rate-limit budget.** This PR makes it a live counter for the first time, but only the caster path writes it; the admin/participant schedule endpoint (`authedMatchScheduler`) reads the same 60/hr budget yet never records to it, so its own limit stays effectively a no-op. Confirm the shared-budget coupling is intended.

becked:
Follow-up: the **Still open** items above are now handled.

- **URL-sync primitive** → conformed to `goto`. The matches-page filters now sync via `goto(url, { replaceState: true, keepFocus: true, noScroll: true })`, matching the app's existing filter-URL writers (`StatsView`, `GamesTable`, `ScopeRow`). Not literally `goto(resolve(...))` — `resolve()`'s branded route types can't express a *dynamic query string*, so this keeps a dynamic URL + the documented `eslint-disable svelte/no-navigation-without-resolve` (precedent: `upload/+page.svelte`). The `[slug]` layout load reads only route params, so the sync doesn't refetch.
- **`mutateMyCasterEntry` callbacks** → named the two closure types (`PrepareCasterEntry` / `ApplyCasterMutation`). Chose aliases over "flatten into two handlers": flattening would either duplicate the shared auth + rate-limit + ledger scaffolding, or break the deliberate resolve-`me`-once-after-the-429-gate ordering.
- **Duplicated matchup formatting** → one `matchupLabel` helper in `match-occupant.ts` that centralizes the bye rule; the Casts view, matches table, calendar, and sesh export all route through it (each passing its own missing-side fallback).
- **`bracketLabel`** → extracted `divisionName` / `matchBracketLabel` into `bracket-label.ts`, now used by the Casts view, match popover, and the signup / signed-up popovers in place of the re-inlined division-name ternaries.
- **`tournament_schedule` rate-limit budget** → already handled in `de4d000`: the schedule PATCH now records the ledger event, so scheduling and casting share the one 60/hr per-user budget (it was inert before).

Two UI changes in the same pass: relabeled the matches view toggle to **Casts**, and removed the redundant Casted/Uncasted filter from the List tab now that the Casts view owns caster status (the sortable Caster column stays; the PR-only `?caster=` param is dropped).

Green: `npm run lint`, `svelte-check` 0/0, `cloud/` `tsc --noEmit`, worker suite **398/398**.

Committed on the branch — will push on your go-ahead.

PR #91, opened 2026-07-02
fix(tournament): let a deliberately added blank part survive save

alcaras:
Found in live use on the deployed parts editor (#81): a match ran long and is heading into a second sitting — the admin clicks **+ Add part** to record "Part 2, time TBD", hits Save, and the new part silently vanishes.

Cause: the save path drops parts left entirely blank, which is right for the **auto-seeded** row (a never-scheduled match opens with one blank part; saving without filling it shouldn't persist a phantom sitting) but wrong for a row the user explicitly added — an unscheduled Part 2 is real state.

Fix: `EditPart.added` marks rows created via "+ Add part"; the blank-row filter now only drops the pristine seeded row (blank + tmp id + not added). Blank parts that were explicitly added, or that already exist server-side, survive the save. The read side already handles them ("Not yet scheduled").

### DIFF
--- a/src/lib/tournament/SchedulePartEditor.svelte
+++ b/src/lib/tournament/SchedulePartEditor.svelte
@@ -29,6 +29,12 @@
 		time: Time | undefined;
 		casters: EditCaster[];
 		streams: EditStream[];
+		// True when the user explicitly clicked "+ Add part" (vs the blank row
+		// the editor seeds for a never-scheduled match). A deliberately added
+		// part saves even while still blank — "this match is going to a Part 2,
+		// time TBD" is real state — while the pristine seeded row stays a
+		// throwaway that saving won't persist.
+		added?: boolean;
 	}
 
 	// Mirror of the Worker's StreamUrlSchema host allow-list. The cloud schema is
--- a/src/lib/tournament/SchedulePopover.svelte
+++ b/src/lib/tournament/SchedulePopover.svelte
@@ -134,40 +134,44 @@
 	);
 
 	async function save() {
+		// Pair each source part with its request payload so the blank-part filter
+		// can read both the editor-only state (added? / temp id) and the computed
+		// payload, then project down to just the payloads.
 		const payload = parts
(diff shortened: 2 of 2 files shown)


becked:
Pushed a small follow-up refactor in b6f3f79 — **behavior unchanged**, purely a design-consistency cleanup.

The `keepIfBlank` flag was being injected into the request payload inside the `.map()`, used by the `.filter()`, then destructure-stripped back out (`.map(({ keepIfBlank: _keep, ...p }) => p)`). That add-a-transient-field-to-the-request-then-strip-it pattern doesn't appear anywhere else in the codebase, and it means the request-shaped object briefly carries a field that isn't part of the request.

Reworked `save()` to pair each source part with its computed payload, filter on both, then project down to just the payloads:

```js
const payload = parts
  .map((p) => ({ part: p, payload: { id, scheduled_at, casters, streams } }))
  .filter(({ part, payload }) =>
    part.added === true ||
    !part.id?.startsWith("tmp-") ||
    payload.scheduled_at !== null ||
    payload.casters.length > 0 ||
    payload.streams.length > 0,
  )
  .map(({ payload }) => payload);
```

This reuses the annotate-then-filter idiom already used by `castingParts` in `MatchPopover.svelte` (pair source with derived data, filter on the pair, project), keeps a single source of truth for "non-empty caster/stream" (still reads `payload.casters.length` after the map drops blanks), and never puts a non-request field on the request object. Same outcome: an explicitly added blank part and an already-persisted part survive the save; the pristine seeded row still drops when blank. `npm run check`, eslint, and prettier all clean.

One behavior observation while in here (not changed by this commit — it's in the original fix and looks intentional): the `!tmp-` clause also means a **persisted** part that gets blanked out (time cleared, casters/streams removed) now survives as a "Not yet scheduled" part rather than being dropped. That's defensible since **Remove part** is the real delete affordance, but it's broader than the added-part case the title describes — might be worth a line in the PR body so it's a conscious call.

Issue #95, opened 2026-07-03
Workflow: reviewing & integrating a collaborator's Claude Code contributions


becked:
## Context

We're getting a steady stream of PRs from an external collaborator (@alcaras) whose contributions are Claude Code-generated. Current process: pull the PR into a local branch that tracks their fork, review it, make fixes on top, push back to the fork, then merge. The recent tournament caster work came in as a 4-PR stack (#81 → #87 → #88 → #89) handled this way.

It works, but it's ad hoc. This issue is to think through how we *want* to work with agent-generated contributions from someone else — both to make review faster and to hold the quality bar we care about (this isn't a vibecoded project).

## The recurring problem

The code itself is generally fine (it's Claude-written), but the contributions reliably tend to:

- **Invent patterns where the app already has one** — e.g. #89 introduced shallow `replaceState` URL-sync when the app standardizes on `goto(resolve(...), { replaceState: true })`.
- **Drift stylistically** from established conventions.
- **Duplicate existing helpers** instead of reusing them (e.g. a 4th "A v B" matchup formatter, each with a different missing-side fallback).
- **Miss app context** — the contributor's agent can't see the whole codebase's conventions, so it produces locally-correct but globally-inconsistent code.

So review needs an explicit **app-fit lens** on top of correctness: does this reuse what we already have, match our patterns, and fit the app we're building?

## Workflow friction observed

- Stacked PRs built on an older `main` need rebasing (`git rebase --onto`) to drop already-merged ancestors; semantic conflicts can hide behind clean-looking auto-merges.
- The pull → fix → push-to-fork loop relies on standing authorization (we can push to the contributor's fork); worth having a clear, repeatable runbook for it.
- Gate gotchas: root `npm run lint` (`eslint src`) doesn't cover `cloud/src`; worker tests live under `cloud/`.

## Proposals to discuss

1. **A `collab-pr` skill / documented runbook** encoding the pull → review → fix → push workflow and the app-fit review lens. Rough shape:
   - Scope the PR (`gh pr view`), note stacked chains (only the top commit is the PR's real work), add the fork remote.
   - Checkout a branch tracking the fork; rebase onto current `main` if stale; resolve conflicts against *current* patterns, not the PR's stale base.
   - Review with the app-fit lens; verify every claim against the codebase before asserting it.
   - Apply must-fixes; green `lint` / `check` / `test` (remember `cloud/` isn't in root lint).
[...]

PR #98, opened 2026-07-04
feat(tournament): copy button for matches still needing casters

alcaras:
From tonight's casting session: the sesh copy exports **every** upcoming match; recruiting casters wants just the uncovered ones.

On the **Cast tab**, a second icon CopyButton (camera glyph, admin-only, beside the existing sesh clipboard) copies the same sesh-style lines — `Match NNN (Part N) - @A v @B - <t:…:F> (<t:…:R>)` — filtered to sittings with **no caster signed up** (the same rule as the tab's "needs a caster" flag), under a "Matches needing casters" heading.

Kept single-sourced per the app-fit lens (#95): it's a `scope` parameter on the existing `seshText` builder (no second line-formatter), rendered through `CopyButton`'s existing icon snippet.

Deliberately covers upcoming **scheduled** casterless sittings only (matching what the Cast tab shows): a caster needs a concrete time to commit to, so matches without one stay in the sesh export's "To be scheduled" until they get scheduled.

### DIFF
--- a/src/lib/tournament/CastView.svelte
+++ b/src/lib/tournament/CastView.svelte
@@ -17,7 +17,7 @@
 		matchSlotDisplayName,
 		matchupLabel,
 	} from "$lib/tournament/match-occupant";
-	import { upcomingScheduledParts } from "$lib/tournament/parts";
+	import { upcomingScheduledParts, CAST_GRACE_MS } from "$lib/tournament/parts";
 	import type { ScheduleZone } from "$lib/tournament/schedule";
 	import {
 		atlasMapUrl,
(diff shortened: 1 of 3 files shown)


becked:
Pushed a fix on top (e2f359b) before merging — flagging one seam this had.

**Problem:** the new "matches needing casters" copy reuses `seshText`, which calls `upcomingScheduledParts(data.matches)` with the default `graceMs = 0` (strictly future). But the button lives on the **Cast tab**, and that tab computes its "needs a caster" list with a **2h grace** (`CastView.svelte`: `upcomingScheduledParts(matches, GRACE_MS)`) — deliberately, so a match that began a few minutes ago is still claimable by a late caster.

So the caster *predicate* (`part.casters.length === 0`) matched the tab, but the *window* didn't. A casterless sitting that started, say, 30 min ago still shows as "needs a caster" on the tab yet was **missing from the copied recruit list** — and the header's `needsCasterCount` could exceed what you copied. That undercuts the recruiting use-case, since those just-started sittings are exactly the ones a late caster can still grab.

**Fix:** extracted the 2h window into a shared `CAST_GRACE_MS` in `parts.ts` (so it's not a magic number living only inside `CastView`), pointed `CastView` at it, and threaded it through the `needs-casters` scope of `seshText`:

```ts
const scheduled = upcomingScheduledParts(
    data.matches,
    scope === "needs-casters" ? CAST_GRACE_MS : 0,
)
```

Now the copy mirrors the tab exactly — same predicate *and* same window. The `all` (sesh schedule) scope stays at grace 0, since a forward-looking scheduling post shouldn't name sittings that are already under way.

Kept it single-sourced in the spirit of the PR: one grace constant, no second window definition. `npm run check` + eslint + prettier clean.

PR #107, opened 2026-07-07
feat(tournament): "copy caster post" button on the match popover

alcaras:
Right after scheduling a session, a TO wants to drop a quick "just scheduled — appreciate a caster!" note in the casters channel with proper Discord timestamps, without hand-converting times on sesh.fyi or screenshotting the schedule editor.

Adds a **megaphone CopyButton** to the match popover header, beside *Copy thread title* / *Copy DM*. Open to any viewer — a player announcing their own match, or a TO. It copies this match's still-upcoming scheduled sittings:

```
Just scheduled:
Match 018 (Part 2) - @A v @B - <t:…:F> (<t:…:R>)
Appreciate a caster!
```

- `<t:…:F>` = full local date, `<t:…:R>` = "in X hours". Player mentions ping via `<@id>` for admins; for everyone else they fall back to display names (discord_id is admin-only) — everything in the post is public info.
- The "Appreciate a caster!" line appears only when a listed sitting still needs one — if it's already covered, you get just the "just scheduled" announcement.
- Hidden when there's nothing scheduled ahead (byes, placeholders, a fully-past match).

**App-fit (per `src/lib/tournament/CLAUDE.md`):** the `Match NNN (Part Y) - A v B - <t:…>` line was inline in the matches-page sesh export and would have been a second copy here — so **commit 1** extracts it to `seshMatchLine()` (new `sesh.ts`) and points the sesh export at it (byte-identical output, verified), and **commit 2** adds the button on top. Reuses `matchupLabel`, `upcomingScheduledParts`, and `CopyButton` rather than re-inlining any of them.

### DIFF
--- a/src/lib/tournament/MatchPopover.svelte
+++ b/src/lib/tournament/MatchPopover.svelte
@@ -37,10 +37,9 @@
 		matchSlotAvatarUrl,
 		matchSlotNation,
 		matchSlotDisplayName,
-		matchupLabel,
 	} from "$lib/tournament/match-occupant";
 	import { upcomingScheduledParts } from "$lib/tournament/parts";
-	import { seshMatchLine } from "$lib/tournament/sesh";
+	import { seshMatchLine, seshVersus } from "$lib/tournament/sesh";
 	import { mapScriptLabel } from "$lib/tournament/map-scripts";
 	import {
 		atlasMapUrl,
@@ -235,13 +234,7 @@
 	const casterPostText = $derived.by(() => {
 		const parts = upcomingScheduledParts([match]);
 		if (parts.length === 0) return null;
-		const versus = matchupLabel(match, (side) => {
-			const id =
-				side === "a" ? match.slot_a_discord_id : match.slot_b_discord_id;
-			return id
-				? `<@${id}>`
-				: (matchSlotDisplayName(match, side, slotLabels) ?? "?");
-		});
+		const versus = seshVersus(match, slotLabels);
 		const lines = parts.map(({ part, partNumber, split }) =>
 			seshMatchLine({
 				matchNumber: match.match_number,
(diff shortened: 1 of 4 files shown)


becked:
Reviewed and pushed two small follow-up commits onto this branch (we have push access to the fork) before merging. Both are on top of your two commits; nothing here changes the button's behavior.

**`7c437a7` — refactor(tournament): extract the shared sesh matchup formatter**
The "A v B" matchup line with `<@id>` Discord mentions (admin-only id → display name → `"?"` fallback) was still duplicated between the matches-page sesh export and this PR's caster post. Extracted it to `seshVersus()` in `sesh.ts` alongside `seshMatchLine()` and pointed both sites at it — so the mention/fallback logic lives in one place. Byte-identical output. This mirrors what the first commit did for the line format, and matches the `src/lib/tournament/CLAUDE.md` guidance ("if a shared helper doesn't exist yet, add one and point all sites at it").

**`281aed4` — fix(tournament): age out upcoming parts on the reactive clock**
`upcomingScheduledParts()` read `Date.now()`, while every sibling time-reading helper in `parts.ts` reads the reactive `nowMs()` (whose documented purpose is exactly "up next" filtering). Because of that, any `$derived` listing these parts only refreshed on a data refetch, not as a sitting's time passed — this affected the new popover caster post **and** the pre-existing `CastView` upcoming list. Switched to `nowMs()` so both age out on the 30s tick. Only side effect: the matches-page sesh export (a click handler) now uses a cutoff up to 30s stale, which is within the tolerance the app already accepts for minute-precision schedules.

Review notes:
- Extraction is behavior-preserving — traced every sesh call site (scheduled line, unscheduled match, unscheduled part) to confirm identical output.
- Gating verified: `slot_*_discord_id` is populated only for admin viewers, so the caster post correctly falls back to display names for everyone else; nothing PII leaks.
- `svelte-check` (0 errors / 0 warnings), `eslint`, and Prettier all clean.

Nice feature — thanks! Merging after this.

PR #117, opened 2026-07-10
feat(tournament): atomic guarded slot-occupant swap

alcaras:
Came out of a live-season question (Siontific in Discord): a round-1 match is blocked — one player has a medical issue — and the TO wants to pair the healthy player against someone from another still-pending match instead, buying the delayed pairing time.

## What exists today

The occupant-substitution endpoint can already achieve this, but as a fragile **three-step dance**: sub player A to a free-text placeholder (the `DUPLICATE_USERNAME` guard blocks a direct move), sub B into A's old seat, sub A into B's. It can be abandoned halfway (leaving a phantom `swap-temp` player in a live tournament), and — the real problem — **nothing stops an admin doing it after a seat has results**, which silently hands one player's record to another.

## This PR

`POST /v1/tournaments/:id/slots/swap` `{ slot_a_id, slot_b_id }` — one atomic, audited batch:

- **Identity moves** (`discord_username`, `discord_id`, `user_id`, `signup_answer`, `claim_banner_dismissed_at`); **the seat stays** (seed, division, committed pairings, history). Every pending match either seat is in simply reads with the other player.
- **Guards**, the core one being what the manual path can't enforce:
  - `409 SLOT_HAS_RESULTS` once either seat has any decided match — **including a bye** (+1 win already banked) — so results can never be reattributed;
  - `409 TOURNAMENT_LOCKED` for cross-division swaps after start (mirrors `handlePatchSlot`'s division lock; fine during setup, where this doubles as the documented odd-division roster fix);
  - same phase, neither withdrawn, distinct slots, both in the tournament.
- Audited as `slots_swapped`; bumps `updated_at` (stats/caches drift correctly).

Match-level pairings stay immutable — this doesn't re-pair rounds, reassign byes, or touch matches at all.

## Scope notes

- **No UI in this PR** — the endpoint is the reviewable core; if you're happy with the mutation semantics I'd follow up with a small "Swap with…" affordance next to the existing substitute pencil. (Happy to hear you'd rather keep it API-only or kill it entirely — the manual path still works and is now described in the rules doc either way.)
- Docs: `api-reference.md` entry (route drift guard green) + the rules-doc "can an admin fix pairings?" FAQ updated to describe occupant-level swapping and its safety rule.
- Six integration tests: swap semantics (identity crosses, seat stays), decided-match guard, bye-counts-as-result, cross-division lock vs setup, authz, cross-tournament 404.

Validation: worker typecheck + suite green (the only failures are the pre-existing `canonical-map-options` mirror drift, which fails on clean `main` too — separate issue), root lint/check clean.


becked:
## Follow-up: add the "Swap" UI (handoff brief for a fresh session)

Reviewed and validated the endpoint locally (worker typecheck + the 6 integration tests green; route-drift guard green; the only suite failures are the pre-existing `canonical-map-options` mirror drift, which fails on clean `main` too). The backend is solid; this note captures the agreed UI follow-up so it can be picked up cold.

### Decision

Land a **"Swap" button in the Swiss standings**, not a CLI command and not API-only. Rationale:

- The standings table is admin-only and collapsed by default, so adding one more per-row action doesn't add visible clutter for players.
- Every sibling slot mutation (substitute / withdraw / reinstate / reorder) already has both a `cloudApi` method and UI. An API-only endpoint is the lone exception and isn't usable by the TO — the whole point was to replace the fragile manual dance, so it needs an affordance.
- A CLI command was considered but rejected: the admin CLI reaches D1 by shelling out to `wrangler d1 execute` (`scripts/admin/wrangler.ts`), so it can't reuse this endpoint's guarded handler — it would have to re-implement the guards in raw SQL and lose the CI-testable coverage this PR already has.

### What to build

A per-row **Swap** button next to Withdraw. Clicking it opens a small popover (follow the inline-edit idiom, **not** a modal) listing eligible players in the same division; picking one calls the endpoint and refreshes.

Three facts from the data model make this clean:

1. **`SwissStandings.svelte` is rendered per-division** (props `divisionName`, `standings: SlotStanding[]`). So the picker just lists the *other rows in this same table* — same-division comes for free, and the server's cross-division `TOURNAMENT_LOCKED` can never be hit from this UI.
2. **Each row already knows swap-eligibility.** Swiss `wins`/`losses` come only from decided matches (a bye is +1 win), so `s.wins + s.losses === 0 && !s.withdrawn` is an exact client-side mirror of the server's `SLOT_HAS_RESULTS` guard. Gate the button on it: if the player already has a result, disable with a tooltip ("Can't swap — already has a result this phase").
3. **Candidate rows carry their current pending opponent** (derivable from the round's matches). Show `name · seed · vs <current opponent>` per candidate — the admin is really choosing the resulting matchup, so surfacing the opponent is what makes the choice legible.

Confirm line before firing: *"Swap Napalm ⇄ Quetzal? Napalm's pending match (vs Minsk) becomes Quetzal vs Minsk; Quetzal's (vs Aran) becomes Napalm vs Aran."*

### Where (anchors current as of this branch — re-grep, don't trust line numbers blindly)

- **`src/lib/tournament/SwissStandings.svelte`** — Withdraw/Reinstate buttons live in the `ml-auto` action cluster (~L145–167); copy that button style. Add an `onSwap?: (slotId, otherSlotId) => void` prop beside `onWithdraw`/`onReinstate`. Row fields available: `slot_id`, `rank`, `wins`, `losses`, `withdrawn`, `status`, `display_name`, `discord_username`, `avatar_url`.
- **`src/lib/api-cloud.ts`** — add `swapSlots(tournamentId, slotAId, slotBId)` next to `patchSlot` (~L1000) / `withdrawSlot` (~L1048) / `reorderSlots` (~L1069). This client method does not exist yet. Endpoint contract: `POST /v1/tournaments/:id/slots/swap` `{ slot_a_id, slot_b_id }` → `{ slots: [SlotRow, SlotRow] }`; errors `400 SAME_SLOT`, `404 SLOT_NOT_FOUND`, `409 PHASE_MISMATCH | SLOT_WITHDRAWN | TOURNAMENT_LOCKED | SLOT_HAS_RESULTS`.
- **`src/routes/tournaments/[slug]/+page.svelte`** — add a `swapSlots` handler modeled on `substituteSlot` (~L466) and the withdraw/reinstate handlers (~L525/532): call `cloudApi.swapSlots(...)` then refresh. Thread `onSwap` into **every** admin Swiss-standings call site — `onSubstitute` is currently passed at ~L725, ~L899, ~L1068, ~L1125; the swap prop must reach the same set (minus championship). Uneven coverage across sibling call sites is this repo's most-repeated defect — wire all of them.

### Reuse / idiom

- Follow the substitute editor's inline-expand + ✓/× pattern (`SlotUsernameCell.svelte`), but note its picker is `UserAutocomplete` (searches *registered users* to sub in) — **not** reusable here, because the swap candidate set is *existing slots in this division*. Style the candidate list like the existing popovers; don't introduce a new modal system.
- Scope to **Swiss only**. The championship is a bracket, not this table, and cross-round swaps there are moot.

### Two gotchas

1. **Empty-state is the real UX risk.** Eligibility collapses the moment results land, so mid-round the picker is often empty (that's the safety guard working). If Swap opens an empty list it reads as broken — so when no eligible partner exists, disable the button with "No swap-eligible players (others have results this round)."
2. **Same-match edge.** If the two are already paired against each other, swapping just flips slot_a/slot_b (a no-op). Cleanest to exclude the player's own current opponent from the picker.

### Validation checklist for the follow-up

`svelte-check` + root `lint` clean; drive it in the real app (seed a local tournament, advance to a generated round-1, swap two pending seats, confirm the pending matches re-read with the players crossed and seeds/divisions unchanged); confirm the button disables correctly once a result lands.

PR #128, opened 2026-07-15
feat(techs): science event rail, source breakdown, and owtt planner links

alcaras:
Adds science analysis to the Techs tab, plus per-player deep links into the [owtt tech-tree planner](https://alcaras.github.io/owtt/).

Most of it is UI-only and works against existing blobs (verified on a 2.6.1-era game and a current 2.10.0 upload). The one exception is the leader's **Court** science, which needs data the blob didn't carry — so this PR also bumps the parser to **2.11.0**, adding two fields (`game_options` and `leader_character_xml_id`) to the parser and the Worker validation schema. Both are optional/nullable: older blobs simply lack them and degrade gracefully (the court reads "unknown" and folds into "Other"). Existing games won't show the court itemization until reparsed at 2.11.0.

Techs tab overview

## What's added

**Science event rail under Tech Discovery Over Time** (1v1s; other games keep the plain chart). The Military tab's rail is extracted into a shared `EventRail.svelte` (a behaviour-preserving refactor for Military). The feature commit also adds a colored-dot fallback for markers whose value has no sprite — currently only reachable from the Techs rail; it's dormant for Military, whose three marker categories all resolve sprites today. Three rows per player:

- **Key science techs** — shown only when the tech's unlock was demonstrably used: the unlocked improvement line standing at game end (per-tier counts, e.g. Odeon/Theater/Amphitheater; shrines typed by god domain from `AssetVariation`), its specialist staffing by tier (Apprentice/Master/Elder), the base science income earned ("at least +X 🧪/turn"), laws adopted, steal-research missions, and expedition story events for Rhetoric. Steal-research polarity (`memory_data` owner = thief, target = victim) is derived from the game source — see the header comment in `science-techs.ts`.
- **Free techs** — turns with multiple completions (research finishes one/turn) and Sages seat foundings.
- **One-off science gains** — turns where the cumulative total jumps more than the turn's rate, attributed best-effort to same-turn steal missions / story events, summed per player in a caption.

Key-tech tooltip

**"Where the Science Comes From"** — an end-state, fully itemized decomposition of each player's science/turn, rendered as a single columnar table (per-source bars on a shared scale for every game size; click a player header to rank the rows by them):

- Specialists and buildings by exact name/tier (groves earn their science off the luxury — `improvementClass.xml` `aaiResourceYieldOutput`).
- Laws: Constitution (+1/urban specialist) and Centralization (capital ×culture level, a floor — `getCultureStep` isn't in the blob) — both fed into each city's base per `City.cs` so the modifier estimates amplify them — plus science law *upkeep* (Divine Rule/Holy War/Autarky, × city count per `Player.getYieldUpkeepNet`).
- Modifiers: libraries/Musaeum with their exact percentages, points computed against each city's base.
- **Court**: the reigning leader's court science (exact — the leader has no opinion of themselves, so no opinion modifier applies) and, under Competitive Mode, its flat science stipend. Both key off the now-parsed `GAMEOPTION_COMPETITIVE_MODE`; on a blob without `game_options` the option is UNKNOWN and the whole court section is omitted rather than priced against a guessed default (the normal and Competitive curves diverge sharply at high Wisdom).
- **"Other" is the signed remainder** vs the real rate — governors and the rest of the court (spouses, successors, courtiers, council) live here, because each is scaled by that character's opinion of the player, which the save doesn't store. A negative remainder would mean the itemized floor over-counted; it's deliberately not clamped.

All rates were validated against the game source and Reference XML.

Science breakdown
*(screenshot predates the review pass — of the Governors/Court/Competitive rows shown, only Governors now folds into "Other"; the leader Court row and the Competitive Mode stipend itemize as pictured.)*

**Techs Only One Player Got** — per-player chip rows (turn + icon + name).

**owtt planner links** — "Full research order in the tech-tree planner: [player ↗]" under the discovery chart; encodes the player's actual research order into the planner's `?n=&o=` format.

## New bakes (generated files checked in)

- `bake:owtt` — snapshots the owtt planner's positional tech/nation encoding from a **local owtt checkout** (`OWTT_DIR`, see `.env.example`), cross-checking every planner tech id against `Reference/XML` `tech.xml` so upstream drift fails the bake loudly.
- `bake:science-yields` — extracts base science of improvements/specialists, per-resource grove science, and shrine domains from `Reference/XML`. Runs in `bake:all`.

## Known limits (candidate follow-up parser PR)

The blob still can't see: council seats (Spymaster court science + real Portcullis evidence), courtiers (`<Courtier>` is in the raw save, unparsed), agents, city projects (Forums→Philosophy, Archive→Metaphysics), per-building "ever built" counts (`<CompletedBuild>`), theologies/religion spread (Dualism), character opinions (the court-rate modifiers that keep governors and the non-leader court out of the itemized Court section), and `story_events` is capped at 100 (limits spike attribution). Those land in "Other" or degrade gracefully.


becked:
Thanks for this — the XML transcription work is genuinely careful, and I verified the citations rather than taking them on faith. Every symbol you name-drop exists where you say it does, and the constants are right: spouse −50% (`globalsInt.xml` `LEADER_SPOUSE_YIELD_MODIFIER`), upkeep −0.5/−1/−2 on exactly Divine Rule/Holy War/Autarky and really × city count (`Player.cs:17984` → `return (iUpkeep * getNumCities())`), Constitution +1/urban specialist, Centralization 2/culture level, Competitive +4, grove +2/luxury with no flat science of its own, Library 10/20/30, Musaeum 50. The `triangle`/`triangleOffset`/`triangleBoost`/`modifyRating` ports are faithful to `InfoHelpers.cs:1210-1236`. "UI-only" checks out (empty diff vs `cloud/` and `src/lib/parser/`), `science-yields.ts` is in sync (I re-ran the bake, no diff), the `EventRail` extraction is behaviour-preserving for Military, and `npm run lint` + `svelte-check` pass clean.

The findings below all need resolving before this can merge. They're ordered by severity. I've quoted the game-source evidence inline so you don't have to re-derive it.

---

## 1. `bake:owtt` fetches and executes remote third-party JavaScript

`scripts/bake-owtt.ts:42-52`:

```ts
const res = await fetch(TECH_DATA_URL);   // https://alcaras.github.io/owtt/tech-data.js
const src = await res.text();
new Function("window", src)(shim);        // arbitrary remote JS, full Node privileges
```

No sandbox, no pinning, no integrity check. It's the only network `fetch` in any bake script in the repo — everything else reads from local disk. Whoever controls that Pages site (or compromises it) gets code execution on any machine that runs `npm run bake:owtt`.

**Fix:** we already have the pattern for this, from your own other repo. `scripts/bake-map-caveats.ts` derives from `alcaras/owtournamentatlas` but reads it from a **local checkout**, resolved via `resolveAtlas()` in `scripts/lib/paths.ts:58-67` and documented in `.env.example`. Please add a `resolveOwtt()` alongside it, read `tech-data.js` from a local `alcaras/owtt` clone, and document the new external checkout in `.env.example` (it's currently undocumented). That makes the input diffable, pinnable to a commit, and drops the RCE surface.

**Related, same file:** the encoding is *positional* against a mutable array, and `gameVersion` is recorded as a comment only (`src/lib/generated/owtt.ts:4`) — never asserted on re-bake. A 404 fails loudly, which is the benign case; but if upstream reorders `techData.techs`, the checked-in table silently becomes wrong and every deep link points at the wrong techs, with nothing in the repo detecting it. Please assert the pinned version/commit on re-bake, or add a test cross-checking `OWTT_TECH_ENC`'s keys against `Reference/XML` tech types.

## 2. "All rates validated against the game source" — the rates are, but the *trigger* is invented

`src/lib/game-detail/TechsTab.svelte:489`:

```ts
const competitiveScaling = $derived(gameDetails.game_mode === "NETWORK");
```

The game gates on `GAMEOPTION_LOWER_CHARACTER_YIELDS` (`InfoHelpers.cs:1212` and `:1226`), not on network mode. I checked: the blob carries **no game-option field at all** — `GameDetails` has `map_options`, which is a different thing.

This isn't cosmetic. That one boolean switches the entire court *and* governor rating curve, and fabricates a "Competitive Mode +4" line item. A network game with the sub-option off gets a phantom +4 attributed to it. Worse, `science-techs.ts:703-713` clamps the remainder with `Math.max(0, …)`, so the over-count silently vanishes rather than surfacing as a negative — the one place this would be visible. The PR text calls `Other` "the honest remainder vs the real rate"; the clamp makes it honest in only one direction.

**Fix:** this needs a decision, not a patch. Either (a) drop competitive scaling and let it fall into `Other`, (b) keep it but don't present an inferred flag's output as an itemized, exact-looking line, or (c) land the parser follow-up that surfaces the real game option first. Happy to discuss which — my preference is (a) or (c). Separately, please reconsider the `Math.max(0, …)` clamp regardless: a breakdown that structurally cannot show it over-counted is suppressing exactly the signal that tells you it's wrong.

## 3. Espionage attribution: "the memory's owner is the thief" is not settled

The header comment in `science-techs.ts:16-20` justifies this empirically ("verified against cached prod games: every owner had Portcullis + Cartography before the steal turn; targets often had neither"). That's a reasonable test and it came out your way, so I'm not asserting you're wrong — but the XML reads the other direction and this needs a definitive answer, because it's the highest-consequence correctness claim in the PR.

What the data says:

- `memory-player.xml:241-244` — `MEMORYPLAYER_STEAL_RESEARCH` is `MEMORYLEVEL_NEG_MEDIUM_NORMAL`, a **negative** memory.
- `mission.xml:1780-1786` — the mission has two success results. `bonus.xml:7996` (`..._SUCCESS`) and `bonus.xml:8011` (`..._SUCCESS_EXPOSED`) grant the **same** +30 base / +10 per science. The *only* thing `_EXPOSED` adds is `<Memory>MEMORYPLAYER_STEAL_RESEARCH</Memory>`.

A negative memory that only exists when you're *caught* reads as the victim's grudge against the thief, i.e. owner = victim.

**The tell:** `MemoryInfo` (`src/lib/parser/types.ts:236-245`) already carries **both** `player_xml_id` and `target_player_xml_id`, and this PR ignores the target field entirely. If the polarity is backwards, every espionage marker lights up on the wrong player — which is precisely the per-player-vs-whole-game trap `src/lib/game-detail/CLAUDE.md` calls out by name.

**Fix:** settle it definitively (trace `PlayerBonus.cs:5142` `addMemory(eMemoryPlayer, ePlayer: eOtherPlayer)` back through `PlayerEvent.cs:16102`'s `doBonus(eTargetBonus, …, zParams.eTargetPlayer, …)` to establish which player object `this` is), then use `target_player_xml_id` explicitly rather than relying on owner polarity. Whichever way it lands, please replace the empirical justification in the comment with the source derivation.

## 4. Science math divergences

**4a. A remarried ruler's living spouse is dropped entirely.** `TechsTab.svelte:518-531` takes the *first* non-divorced marriage via `.find()`, then filters `death_turn == null` and gets `undefined`. A ruler who remarried after a spouse died has two non-divorced marriages, so the living spouse contributes nothing. `Player.cs:18250-18258` loops **all** living spouses:

```csharp
for (int i = 0; i < leader().getNumSpouses(); ++i) {
    Character pLoopSpouse = leader().getSpouseAtIndex(i);
    if (pLoopSpouse.isAlive()) iTotal += getLeaderSpouseYieldRateCharacter(eYield, pLoopSpouse);
}
```

This is a total drop, not an undercount. (Minor, same area: `court` is a `Map` keyed by label at `science-techs.ts:668-673`, so same-first-name members collapse.)

**4b. Constitution and Centralization are applied at the wrong level.** Both are EffectCity yields — they land in the city's *base* (`City.cs:4420` → `calculateBaseYield`) and `City.cs:4682` then multiplies that base by the city's total modifier, so in-game they **are** amplified by libraries, Musaeum, and governor Wisdom. `science-techs.ts:627-649` parks them in the flat player-level `laws` bucket, excluded from `cityFlat` (`:573-599`) — never multiplied, and never contributing to the modifier estimate. Undercounts twice.

**4c. Centralization drops `getCultureStep()`.** `City.cs:11916`:

```csharp
iRate += (iYield * (int)(getCulture() + getCultureStep() + 1));
```

`science-techs.ts:641` uses `CULTURE_LEVELS.indexOf(capitalCulture) + 1`, capped at 4. `getCultureStep()` increments each time an already-Legendary city grows culture again (`City.cs:6818-6824`), so a post-Legendary capital that kept growing earns 2×(4+N) but reports a flat 8.

**4d. `Math.floor` where C# truncates toward zero.** `science-techs.ts:523` in `boostRating`. C# `iValue /= iEquivalentRatingValue` (`InfoHelpers.cs:1234`) truncates toward zero; `42·w/5` is not an exact division. A −1 Wisdom governor is **−8%** in-game and **−9%** here. Negative Wisdom is reachable (`trait.xml:302, 916, 2036` grant −1/−2) and nothing filters it — `TechsTab.svelte:555` and `science-techs.ts:616` only skip `=== 0`. (Note `modifyRating` is fine: 10·w·8/5 and 5·w·8/5 both divide exactly.)

## 5. Repo fit

**5a. `MilitaryTab` doesn't use the types `EventRail` exports.** This is the one I'd most like fixed. The PR extracts `EventRail.svelte` with exported `RailMarker` (`:6`), `RailRow` (`:15`), `RailGroup` (`:18`), wires `TechsTab.svelte:24-27` to them — and leaves `MilitaryTab`, *the file it was extracted from in the same commit*, redeclaring all three locally:

- `MilitaryTab.svelte:309` — `RailEvent` restates all five fields of `RailMarker` (it's exactly `RailMarker & { kind: … }`)
- `MilitaryTab.svelte:489` — redeclares the exported `RailRow`
- `MilitaryTab.svelte:490` — inlines `{ player; rows }[]` instead of `RailGroup[]`

Two call sites, one wired — see CLAUDE.md rule #2 ("wire all N parallel surfaces"), our single most-repeated defect. Knock-on: `export type RailRow` currently has zero external consumers, which trips rule #4.

**5b. Duplicated constants and derivations** (rule #3, extract don't copy-paste):

- `TOOLTIP_TEXT` / `TOOLTIP_MUTED` are byte-identical in `TechsTab.svelte:237-238` and `MilitaryTab.svelte:67-68`. These are named constants copied between components, not inline chart-option literals — the justification comment at `TechsTab.svelte:236` doesn't cover this case. They should have moved with the extraction (along with `TOOLTIP_SEP`'s `#4a453d` at `EventRail.svelte:63`).
- `orderedPlayers` is an identical 8-line `$derived` in `TechsTab.svelte:112-120` and `MilitaryTab.svelte:78-86`. `helpers.ts` already hosts every sibling per-player helper (`resolveDetailPlayers:804`, `ownedByPlayer:822`, `findByPlayer:837`) — it belongs there.
- `techName` is added at `TechsTab.svelte:240-241` but the file's two pre-existing inline copies (`:170`, `:1251`) weren't converted — three spellings in one file now.
- `TechsTab.svelte:547-552` and `:577-582` recompute the identical four-arg `ownedByPlayer(cityStatistics.cities, …)` call.

**5c. `improvementDisplayName` is a fifth copy of the improvement-label idiom, in the wrong home.** `science-techs.ts:202-206`. The idiom is already inlined at `MapTooltip.svelte:60`, `TimelineTab.svelte:137`, `ImprovementsTab.svelte:337`, `OverviewTab.svelte:460`. Right extraction, wrong module (it's not science-specific — put it in `helpers.ts`), and no call site was converted — so the shrine-domain suffix ("Shrine of Nabu (Wisdom)") lands only on the Techs tab while the other four keep the bare label. That's a new inconsistency, not just a missed cleanup.

**5d. Bake plumbing:**

- `bake:science-yields` is missing from `bake:all` in `package.json`. It's a pure local-`Reference/XML` bake exactly like `bake:map-caveats`, which *is* in the chain — so the documented "re-bake when Reference/XML is refreshed" workflow will silently leave it stale. (`bake:owtt` correctly stays out.) Note `bake:unit-stats` is already absent too, so the precedent is 9-of-10, not clean — but please add `science-yields`.
- `scripts/bake-science-yields.ts` is the only bake script that fails `npx prettier --check` (`:139-142`, `:175`, `:180`, `:183-184`). `npm run format` would reformat it and produce a spurious diff. It slipped through because `npm run lint` is `eslint src` and doesn't cover `scripts/`.
- Minor: `sorted()` at `:179-180` only sorts top-level keys, so `IMPROVEMENT_RESOURCE_SCIENCE`'s inner resource keys emit in XML order. Deterministic, so not a repro bug — just inconsistent with the sorting applied elsewhere.

## 6. Two PR-description corrections

- "pixel-identical" is true of c6a3641 itself, but the colored-dot fallback you bundle into that sentence actually landed in `deb5ecf`, a different commit. The description conflates the two.
- The dot fallback is **dormant** for Military, not active. I checked all three marker categories against the baked manifest: laws all have sprites; the 26 sprite-less units all have `tech: null` so they can never be selected as `unlocked[0]`; all 10 archetype traits have sprites. Zero sprite-less selections reachable today. Worth saying so, since "markers gained a fallback" implies a visible change that isn't there.

---

Also worth a rebase — you're 3 commits behind `main`.

Nothing here is a rewrite; #2 and #3 are the two that need a decision rather than a patch, and I'd rather settle those before the mechanical fixes land. Happy to talk through either.

alcaras:
Thanks — this is exactly the scrutiny it needed. #1, #4, #5, #6 are all accepted as-is; per your sequencing I'm holding the mechanical batch until we settle the two decisions, so those first.

## 3. Steal-research polarity — settled via the trace you suggested: owner = thief

Followed the chain end to end:

1. `mission.xml` `MISSION_STEAL_RESEARCH`: `<SubjectCharacter>SUBJECT_SPYMASTER</SubjectCharacter>`, `<TechPrereq>TECH_CARTOGRAPHY</TechPrereq>` — the mission runs on the acting player's side.
2. `PlayerEvent.cs:15851 doMission(MissionData)` is `partial class Player` code executing on the mission's owner: `this` = the thief (it's the thief's mission list, `pMissionCharacter` is their Spymaster, `incrementGoalMissionsCompleted` ticks the thief's goals). The result then runs `doBonus(eTargetBonus, …, eOtherPlayer: zParams.eTargetPlayer /* victim */)`. Cross-check that `this` can't be the victim: the same `doBonus` invocation is what pays out the +30/+10-per-tech science, which goes to the thief.
3. `PlayerBonus.cs` meMemory branch → `Player.cs:14046 addMemory` appends `MemoryData(eMemory, ePlayer = victim, …)` to **`this` = the thief's** memory list.
4. The read side resolves the direction of the grudge: `PlayerOpinion.cs:1841 calculatePlayerOpinionOfUsRate(ePlayer)` — the function computing "**other player has this opinion of us**" (the code comment at the end of that function says exactly that) — iterates **our own** `getMemoryList()` matching `pLoopMemory.mePlayer == ePlayer`. A memory on P's list targeting Q therefore feeds **Q's opinion of P**.

So `MEMORYPLAYER_STEAL_RESEARCH` on the thief's list targeting the victim is precisely "the victim resents the thief." Your reading of the XML was right about *whose grudge it is* — the storage convention is just inverted from the intuitive one: the offender's list is the ledger. It also explains why the empirical check was structurally sound: the mission's TechPrereq is Cartography, every memory **owner** had it before the steal turn, and in several games the **target** provably didn't (so the target could not have run the mission).

I'll replace the empirical comment with this derivation and document `player_xml_id` = thief / `target_player_xml_id` = victim explicitly.

## 2. Competitive scaling — agreed, the trigger was invented; proposal: your (a) now, (c) restores it

Two source facts that shape the fix:

- `GAMEOPTION_LOWER_CHARACTER_YIELDS` is `<zSubTypeOf>GAMEOPTION_COMPETITIVE_MODE</zSubTypeOf>` (`gameOption.xml`), and `GameParameters.cs:166 IsGameOption` resolves sub-typed options recursively to their parent — the sub-option is never stored independently. So the rating curve *and* the `EFFECTPLAYER_COMPETITIVE_MODE` +4 both key off `GAMEOPTION_COMPETITIVE_MODE` being set.
- Raw saves record it: `<GameOption>GAMEOPTION_COMPETITIVE_MODE</GameOption>` is present in the tournament saves I checked. The blob just doesn't carry the option list yet.

So, in this PR: drop everything option/curve-dependent — governors, court, and the Competitive Mode line fold into `Other` — and remove the `Math.max(0, …)` clamp so the remainder is signed and an over-count is visible instead of suppressed (agreed the clamp was hiding exactly the diagnostic that matters). The rating-curve ports come out with them (no dead code). The parser follow-up adds `game_options: string[]` from the Root option list (trivial parse) plus `<Courtier>`, council seats, and agents — and reinstates governors + court behind the real flag. That also collects your 4a (loop **all** living spouses, matching `Player.cs` `getNumSpouses`), the label-keyed-`Map` name collision, and 4d's trunc-vs-floor in one place, gated correctly. If you'd rather keep governors in now under a hedged presentation I'm open, but (a)-now-(c)-later reads cleanest to me too.

## Everything else — accepted, queued behind your ack on #2

- **#1**: `resolveOwtt()` local-checkout pattern alongside `resolveAtlas()`, documented in `.env.example`, plus a bake-time cross-check of `OWTT_TECH_ENC` keys against `Reference/XML` `tech.xml` zTypes so an upstream array reorder fails the bake loudly instead of silently poisoning every deep link.
- **#4b**: Constitution/Centralization move into the per-city base so the library/Musaeum point estimates amplify them (matching `City.cs:4682`); itemization unchanged.
- **#4c**: `getCultureStep()` isn't derivable from the blob (`culture_level` stops at Legendary) — will mark the Centralization figure as a floor in the comment; exact needs another parser field.
- **#5a–d**: all of it — MilitaryTab wired to the exported rail types (`RailEvent` becomes `RailMarker & { kind }`), tooltip constants move with the extraction, `orderedPlayers` → `helpers.ts`, `techName` unified (three spellings → one), the duplicated `ownedByPlayer(cityStatistics.cities, …)` computed once, `improvementDisplayName` → `helpers.ts` with the four pre-existing inline sites converted so the shrine-domain suffix is consistent everywhere, `bake:science-yields` added to `bake:all`, prettier on the bake script, inner-key sorting in the emitter.
- **#6**: description fixed — the dot fallback landed in deb5ecf not c6a3641, and it's dormant for Military today (nice reachability check on all three categories).

Plus the rebase onto current `main`. Say the word on #2 and the batch lands.

alcaras:
Batch landed (alcaras gave the go-ahead on the #2 direction, so I didn't wait — shout if you'd have preferred otherwise). Rebased onto current `main`; the review fixes are two new commits on top so the delta is reviewable: `7b358f1` (bake plumbing) and `6a661d7` (everything else).

Per finding:

1. **owtt bake** — now reads `tech-data.js` from a local checkout via `resolveOwtt()`/`OWTT_DIR` (documented in `.env.example`, same pattern as `resolveAtlas`), and cross-checks every planner tech id against `Reference/XML` `tech.xml` at bake time, so planner-side drift fails loudly. No network, no remote code execution. Positional reorders remain visible as a reviewable generated-file diff on re-bake.
2. **Competitive scaling** — governors, court, and the +4 stipend are out of the itemization and fold into `Other`, which is now **signed** (clamp removed; the butterfly/table render a negative remainder as-is). The rating-curve ports went with them — no dead code. They return behind the real `game_options` flag in the parser follow-up.
3. **Steal-research polarity** — code unchanged (owner = thief held up), but the empirical comment is replaced with the full source derivation and `target_player_xml_id` (= victim) is documented.
4. **Science math** — 4a/4d went out with the court/governor removal (they're queued for the parser follow-up, with all-living-spouses and trunc-vs-floor noted); 4b: Constitution/Centralization now feed each city's base before the modifier pass, matching `City.cs:4682`; 4c: Centralization commented as a floor (`getCultureStep` not in the blob).
5. **Repo fit** — all of it: `MilitaryTab` now consumes `EventRail`'s exported `RailMarker`/`RailGroup` (`RailEvent` is `RailMarker & { kind }`); `TOOLTIP_TEXT`/`TOOLTIP_MUTED` moved into `EventRail`; `orderPlayersUploaderFirst` and `improvementDisplayName` hoisted to `helpers.ts` with the four pre-existing inline improvement-label sites converted (MapTooltip, TimelineTab, ImprovementsTab, OverviewTab — shrine-domain suffix is now consistent everywhere; MapTooltip's deity-first rebrand fallback for override-less zTypes is subsumed); `techName` unified (three spellings → one); the duplicated `ownedByPlayer(cities…)` collapsed to one call; `bake:science-yields` added to `bake:all`; the bake script passes prettier; inner resource keys sorted. Bonus: all-empty breakdown sections are now hidden rather than rendering a `0 / 0` row.
6. **Description** — PR body updated: the dot fallback is correctly attributed to the feature commit and noted as dormant for Military (nice reachability check), and the breakdown section reflects the post-review shape.

`npm run lint`, `svelte-check`, and `prettier --check` all pass; re-verified live against a current 2.10.0 tournament blob and a 2.6.1-era one (totals still reconcile exactly — e.g. 108 + 64 + 32 + 65.7 = 269.7).

becked:
Pushed `f414809`: the `game_options` parse + exact leader court science (blob 2.11.0). The Competitive stipend and the leader's court science are now itemized instead of folded into Other — didn't want to release in the half-itemized state.

One scope note worth having: the flag turned out necessary but not sufficient. Only the **leader** is exactly computable (no self-opinion). Governors and the rest of the court — spouse, successors, courtiers, council — are opinion-scaled, and opinion isn't in the save (recomputed on load), so they stay in Other for now. `game_options` is a `Record<string, true>` off the `<GameOptions>` block (parent flags only; sub-options are never serialized). Filed #130 for surfacing the rest of the options + `opponent_level` on the Settings tab.

PR #133, opened 2026-07-16
feat(techs): side-by-side tech timeline + Science Sources icons and ordering

alcaras:
Ports owglick's side-by-side "tech — turn by turn" comparison to the Techs tab, per-ankh-styled, and folds in a round of Science Sources upgrades. UI + bake changes only — no Worker, schema, or parser changes.

Tech timeline

## Techs by Turn (new)

One row per turn anyone completed a tech, one column per player (N-player from the start), each cell stacking that turn's techs as icon chips:

- **Encoding is shape + fill, not hue alone** (several nation colors — Assyria, Kush, Yuezhi — sit on top of any gold): shared-by-everyone = gold text on a faint gold wash; only-they-have-it = the player's color with a left edge bar; bonus card = full outline; law-unlocking techs carry the game's golden-scales glyph.
- **The key is visual** — exemplar chips styled with the exact classes the real cells use, including one "only {player}" chip per player in their color, so it doubles as the color legend.
- **Hovering a tech spotlights it in every column**, so pacing gaps on the same tech read at a glance.
- **The owtt planner deep links move into the column headers** ("planner ↗" next to each player); the standalone planner-links row and the Unique Techs section are removed — unshared techs now read directly off the timeline. The tech pivot table is retired too (this view carries its information with sequence added), along with its `tables.techs` state plumbing.
- Law-unlock data comes from a new baked `TECH_LAWS` table (`lawClass.xml` `TechPrereq` × `law.xml` `LawClass`); the tooltip names the pair ("unlocks Tyranny / Constitution").

## Science Sources upgrades

Science Sources

- **Row icons everywhere**: two new baked sprite categories from pinacotheca — `improvements/` (2D set, resolved through `zIconName` so tiers share their line icon: `LIBRARY_2` → `IMPROVEMENT_ACADEMY`) and `specialists/` (one class icon, tiers share it) — plus law icons on the law rows.
- **Sections read early-tech → late-tech**: rows order by the research cost of the source's unlocking tech (baked `*_UNLOCK_COST` tables; specialist tiers of untiered workplaces — Elder Officer on the single Barracks — inherit their class's cost). Clicking a player still re-ranks by their contribution.
- **Specialists split into Rural / Urban sections**, classes contiguous with tiers ascending Apprentice → Master → Elder.
[...]

PR #136, opened 2026-07-17
feat(military): bonus-card unit grants + white flag glyphs on the event rail

alcaras:
Two small Military-tab rail improvements: bonus cards that grant units now show on the timeline, and unit markers use the game's white flag glyphs instead of the painted portrait art.

rail

## Bonus-card unit grants

Taking a free-unit card (e.g. 2× a unique unit) is a real military-power swing, but the rail only showed unlock techs. Now:

- `bake-unit-stats.ts` also emits `TECH_BONUS_UNITS`: tech.xml `<BonusDiscover>` → bonus.xml `<aiUnits>` pairs (32 bonus techs grant units). Civilian cards (Free Worker/Settler) are in the table but filtered out on the rail via `classifyUnit`.
- A grant gets a marker at the card's discovery turn, ringed in the player's color — the same "outlined = bonus card" encoding the Techs tab uses (new optional `outlined` flag on `RailMarker`). No `TARGET_STRENGTHS`/naval gate: granted units materialize regardless.
- Tooltip: card name (via `TECH_NAMES`, so "Free Hittite Chariot 1"), then "Bonus card · granted" with each unit, count, and strength. The unit line is shared with the unlock tooltip (`unitTooltipLine`).

carthage
hittite

## White flag glyphs for unit markers

The game ships **two** Sprite assets per unit name: the painted portrait (128²) and the small white glyph shown on in-game unit flags (~58²). Pinacotheca's sprite extraction is first-write-wins per name, so the glyph variant was silently dropped for every unit except Longbowman (where the glyph happened to win — which is why it always looked different).

- New `units-icons` bake category mirrors `extracted/sprites/units_icons/` (107 glyphs). Rail unit markers and tooltip unit lines use it — much crisper at 14px, and it matches how units read in game.
- `getSpritePath("units-icons", …)` falls back to the painted sprite for the handful of units that ship no glyph.
- Units that borrow another unit's icon (unit.xml `zIconName`, e.g. `UNIT_HITTITE_CHARIOT_1` → `UNIT_THREE_MEN_CHARIOT`, 36 units) now resolve through a baked `icon` alias on `UnitStat`, applied in `getSpritePath`/`resolveTechSprite` — this also fixes the previously-blank dot markers for Mounted Lancer, Beja/Medjay Archer, etc.

## Re-bake note (action needed on your side)

`bake:sprites` now expects `extracted/sprites/units_icons/` in your pinacotheca checkout, which your extraction won't have yet. I extracted the glyphs with a standalone script (groups `UNIT_*` sprites by name across all asset files, keeps the smallest variant, skips `UNIT_3D_*`/`UNIT_ACTION_*` and names whose only sprite is the 128² portrait). Happy to send it as a pinacotheca PR — either standalone or folded into the extractor as a variant-aware pass — whichever you prefer. Until then this PR's committed `sprite-manifest.ts` covers the runtime; as with #133, the manifest also carries hash refreshes from my pinacotheca extraction, so feel free to re-bake once the extraction lands.

## Testing

- `npm run lint`, `svelte-check` clean.
- Verified on a local 1v1 (Carthage v Hittite): Free African Elephant T41 and Free Hittite Chariot T46 markers, ringed, correct tooltips; Worker cards correctly absent; chariot glyph resolves via the `zIconName` alias.


becked:
Thanks for this — the bonus-card rail and the flag-glyph idea are both great. Heads-up on a change on my side that affects how the glyph part lands, plus how I'd like to split this.

## Pinacotheca now ships glyphs differently

Rather than a separate `units_icons/` mirror, pinacotheca's extractor now emits each flag glyph as a **co-named sibling in the same folder**: `<NAME>__ICON.png` next to the full-art `<NAME>.png`, under `units/`, `crests/`, `religions/`, `traits/`, etc. There is **no `units_icons/` directory** anymore. The split fires whenever two-or-more Sprite objects share an `m_Name` (largest → `NAME.png`, the rest → `__ICON`). So `bake:sprites` needs to read `units/UNIT_<NAME>__ICON.png`, not a mirror dir — the `copyUnitIcons()` pass in this branch would `readdir` a directory that no longer exists.

So no standalone pinacotheca extraction PR is needed on your end — the `__ICON` extraction already lives in pinacotheca.

## Plan for this PR

Splitting by what the pinacotheca change touches:

- **Bonus-card unit grants (A)** — keep as-is. 👍
- **`zIconName` alias (B)** — keep. This is exactly the resolution pinacotheca recommends for units that reuse another unit's art (below), so it's the right mechanism.
- **Flag glyphs (C)** — rework to the `__ICON` convention: resolve `zIconName` first (via B), then look up `units/<name>__ICON`, and **drop the portrait fallback**. Glyph-or-nothing: a glyph slot rendering the full portrait shows a categorically wrong image (the inverse of the Longbowman bug this PR fixes). Where no glyph resolves, the rail keeps its existing colored-dot marker. I'll regenerate `sprite-manifest.ts` / `unit-stats.ts` against my pinacotheca + Reference, which also supersedes the manifest hash drift this branch carried.

## On the units that lack a glyph

I checked the combat units without a direct `__ICON`. Against the canonical Reference XML **none of them has a `TechPrereq`, and none is a bonus-card grant — so none can appear on the military rail.** The dropped fallback never fires for this feature.

Pinacotheca diagnosed those units too, in three kinds:

1. **Reused art** — Hunnic Cavalry 1/2, Hanno Navigator: no sprite of their own; `zIconName` points at another unit (Amazon Cavalry / Elite Amazon Cavalry / Unique Scout). The glyph exists under the referenced name — resolve `zIconName` and it's there (this is exactly B).
2. **Singular/plural mismatch** — Kushan Warlords: `zType` is `UNIT_KUSHAN_WARLORDS` but `zIconName` is `UNIT_KUSHAN_WARLORD`; the singular ships both files. Resolve via `zIconName`.
3. **Split naming** — Buddhism/Hinduism Disciple: the one genuinely odd case. The game stored glyph and portrait under two different names (glyph at the bare `UNIT_*_DISCIPLE`, portrait under `UNIT_HINDU/BUDDHIST_DISCIPLE`), so the co-name splitter never paired them. Worth noting the bare `UNIT_HINDUISM_DISCIPLE.png` in pinacotheca's output is now the ~50px glyph, not a portrait — a watch-item for any consumer that renders a disciple expecting full art, but it doesn't touch this rail.

## Logistics

I can commit the part-C rework straight onto this branch (thanks for leaving maintainer edits on), so A and B ride along unchanged.

becked:
Rebased this branch onto current `main` and force-pushed — it had fallen ~22 commits behind.

**What changed:**
- Replayed the 4 commits onto `main` @ deploy 2026-07-18 → tip `1d0663d`; branch is now even with `main`.
- Conflict resolution:
  - `scripts/bake-sprites.ts` — `main` landed a tech-icons feature (improvements/specialists sprite baking) after this branch forked, so I kept that **alongside** the `__ICON` glyph handling and dropped the `units_icons` mirror in favor of the co-named `UNIT_*__ICON` siblings.
  - `src/lib/game-detail/helpers.ts` — merged `main`'s new `improvements`/`specialists` `getSpritePath` branches with this PR's `units`/`units-icons` ones.
  - `src/lib/generated/sprite-manifest.ts` — regenerated rather than hand-merged.
- Re-baked from pinacotheca 2.7.0 (`sprite-manifest.ts` + `unit-stats.ts`).
- `npm run lint` and `svelte-check` both clean (0 errors / 0 warnings).

Ready for review.

becked:
### Bonus-card markers now read apart from tech unlocks (`16f2419`)

Bonus-card grants and unit-tech unlocks shared the tech row and rendered as the same white flag glyph — tellable apart only on hover. They now carry a **gold glow**.

**What changed**
- `RailMarker` gains an optional `gold` flag. When set, the marker renders with a layered `drop-shadow` in the `--color-orange` token (no hardcoded hex). It hugs the glyph silhouette rather than sitting behind it as a plate, so the flag detail still reads at 14px.
- The flag is OR'd across a proximity merge, so a cluster stays gold whenever *any* collapsed marker is a bonus card — the highlight can't silently vanish when a grant merges into an adjacent unlock.
- Tooltip sub-header: `Bonus card · granted` → `Bonus card`.

**Why this over the alternatives**
- *A separate bonus lane* (mirroring the Techs tab's `key`/`free`/`event` rows) was the main contender. It's structurally cleaner — markers on different rows can never cross-merge — but it costs a row per player, and an unlabeled lane means nothing until you hover. The glow is compact, keeps grants and unlocks interleaved on one chronological row, and self-explains ("gold = special") without a hover.
- *Portrait vs. glyph* was rejected: it would reintroduce the muddy 128²→14px portraits this branch deliberately moved away from in 9e14889/1d0663d, and unlike a boolean flag the distinction can't survive a merge.
- This supersedes the player-colored ring from 8ff56b4 (parked in 9c1439a). Player color is already spent on the rail — the nation crest and the hover guide line both encode identity — so a player-colored ring restated identity instead of signalling a category. Gold is an unused channel here and reads as "bonus".

PR #137, opened 2026-07-17
feat(tournament): remember casters' stream links and auto-attach on cast

alcaras:
Fixes the "sion clicks *I'll cast* but his stream link never makes it onto the match" gap. The root cause is structural: `casters/me` is open to any logged-in user, but stream links live on the parts list behind the participant/admin-gated schedule editor — a non-participant caster **can't** attach their own stream at all.

## What this does

**Casters get a remembered stream link that rides along with the cast.**

- New `users.stream_url` (migration 0033), validated against the same YouTube/Twitch host allowlist as part streams (`StreamUrlSchema` is now shared with `schemas/user.ts`).
- **Taking the streamer slot auto-attaches the stored link** to the part's `streams`. Co-casters don't attach (the cast airs on the streamer's channel); promoting yourself to streamer does. Dedupe by URL; respects the 20-stream cap (now the exported `MAX_STREAMS_PER_PART`, single-sourced with the schema). All inside the existing `parts_rev` CAS loop.
- **Dropping detaches it again** (matched by URL), so an un-cast sitting isn't left pointing at a channel nobody is casting on.
- **First-time casters are prompted once, inline**: if you'd become the streamer and have no stored link, "I'll cast" unfolds a small input (`youtube.com/@you/live (optional)`) right in the row. Whatever you enter is persisted (`stream_url` on the cast body) — every later cast is a single click. Bare domains get `https://` prefixed client-side (the API requires a scheme). Empty = cast streamless, exactly today's behavior.
- **Profile control**: Account → Preferences gains a "Casting stream link" field — edit, or clear to turn auto-attach off. `POST /v1/auth/settings` became a partial update (all fields optional, `null` clears) and echoes the full settings; `GET /v1/auth/me` now returns `stream_url`.
- **Backfill**: the migration seeds `stream_url` from casting history — the first stream URL on each user's most recent streamer sitting. Heuristic (streams aren't attributed to users in the parts data), but it only sets a visible, editable default; on a copy of current data it correctly infers the links for 8 casters (sion, alcaras, aran, boldus, …), so existing casters get one-click casting from day one.

prompt
done
prefs

## Testing

- 7 new integration tests in `cast.test.ts` (13 total, all green): remember-once via the cast body, settings round-trip incl. partial update, streamer-vs-co-caster attach, promote-to-streamer attach, URL dedupe, drop cleanup leaving others' streams, host allowlist on both endpoints, and the 0033 backfill statement itself (duplicated in-test with a keep-in-sync pointer, since the harness applies migrations to empty tables).
- `npm run lint`, `svelte-check`, and the worker suite pass. Exercised end-to-end on the local stack (cast → auto-attach → drop → detach, plus the browser flow in the screenshots).

## Heads-up, unrelated to this PR

`main` currently fails 4 unit tests in `canonical-map-options.test.ts`: b39f690 re-baked `map-option-defs.ts` (new `MAP_OPTIONS_MULTI_DOTA_INNER_TERRAIN`) but `cloud/src/tournament/canonical-map-options.ts` didn't get the matching entry. Left untouched here to keep this PR scoped — happy to send the four-line mirror fix separately if useful.

### DIFF
--- a/cloud/migrations/0033_user_stream_url.sql
+++ b/cloud/migrations/0033_user_stream_url.sql
@@ -1,6 +1,20 @@
--- Per-user casting stream link (twitch.tv / youtube.com, validated in the
+-- Per-user casting stream link (youtube.com / twitch.tv, validated in the
 -- Worker). Set from account preferences or remembered from the first cast
 -- signup that provides one; when a user takes the streamer slot on a match
 -- part, this URL is auto-attached to the part's streams so "I'll cast"
 -- doesn't leave the sitting streamless. NULL = no auto-attach.
 ALTER TABLE users ADD COLUMN stream_url TEXT;
+
+-- Backfill from casting history: for each user, the first stream URL on the
+-- most recent part where they held the streamer slot (casters[0]). Streams
+-- aren't attributed to users in the parts data, so this is a heuristic — in
+-- practice the first stream on a cast sitting is the streamer's channel —
+-- and it only seeds a default the user can change or clear in preferences.
+UPDATE users SET stream_url = (
+	SELECT json_extract(p.value, '$.streams[0].url')
+	FROM tournament_matches m, json_each(m.parts) AS p
+	WHERE m.parts IS NOT NULL AND json_valid(m.parts)
+		AND json_extract(p.value, '$.casters[0].user_id') = users.user_id
+		AND json_extract(p.value, '$.streams[0].url') IS NOT NULL
+	ORDER BY m.rowid DESC LIMIT 1
+) WHERE stream_url IS NULL;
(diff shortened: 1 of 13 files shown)


becked:
Pushed an update (`037024d`):

- **Rebased onto current `main`** (was 12 commits behind) — conflict-free, since none of those commits touch files this PR does.
- **Extracted a shared URL helper.** The "prepend `https://` to a bare domain" idiom was inlined in both new stream-link inputs (`CastControls`, account preferences) and, separately, already duplicated a third time in the pre-existing `TournamentLinksEditor`. All three now go through `ensureUrlScheme` in `$lib/utils/url.ts`. (One intended nuance: the old editor copy skipped prefixing for any `scheme://`; the shared helper only skips `http(s)://`.)

Verified: `npm run lint` clean, `svelte-check` 0 errors, cast integration suite 13/13.

Filed two follow-ups so they don't block this PR:
- #140 — the drop-detach can orphan a stream link if a caster changes their stored URL between casting and dropping the same match (options + a recommendation in the issue).
- #141 — the pre-existing `canonical-map-options.test.ts` failures your heads-up flagged: `b39f690` re-baked the generated Dota map options but the worker mirror wasn't updated. Confirmed still red on `main`; the fix belongs there, not here.

PR #139, opened 2026-07-17
feat(techs): science chart with Cumulative | Per Turn | Number of Techs toggle

alcaras:
The Techs tab's chart gains a view toggle — **Cumulative | Per Turn | Number of Techs** — defaulting to cumulative science.

Motivation for the toggle (community feedback on the first cut of this PR): cumulative totals get big after ~turn 60, so the effect of a single tech adoption disappears into the curve. Per Turn keeps the numbers small enough to see it; Number of Techs is the old chart, kept as a third view — including its per-point tooltip naming the tech discovered at each step.

| Mode | |
|---|---|
| Cumulative Science (default) | cumulative |
| Science per Turn | rate |
| Techs Discovered | techs |

- Default is cumulative because the science rail's markers are science events — key-tech payoffs, free techs, one-off gains — so they read against the science curve they actually affect (the rail's original pairing). The rail, "Bonus science" caption, and planner links render under all three views.
- Same compact Military-Power styling in every mode; the toggle reuses the Yields tab's `ToggleGroup` tokens. The per-turn science rate also lives on the Yields tab, but here it sits under the rail's annotations, which is the point.
- **Bug fix in `Chart.svelte` that the toggle exposed**: the option-change `$effect` called `onLayout?.()` *tracked*, so a parent whose inline handler bumps layout-tick state (the rail pattern) re-created the handler prop and re-ran the effect endlessly — `effect_update_depth_exceeded` on the first option change under a rail. The notify is now `untrack()`ed. The same latent loop was reachable on the Military tab (chart-filter change on a matchup with the rail).
- Two files (`TechsTab.svelte`, `Chart.svelte`); independent of #133 — whichever merges second gets a trivial rebase, which I'll handle.

`npm run lint` + `svelte-check` clean; all three modes + toggle cycling exercised on a local 1v1 (loop verified fixed — zero page errors).

### DIFF
--- a/src/lib/Chart.svelte
+++ b/src/lib/Chart.svelte
@@ -14,7 +14,7 @@
 </script>
 
 <script lang="ts">
-	import { onMount, tick } from "svelte";
+	import { onMount, tick, untrack } from "svelte";
 	import type { ECElementEvent, EChartsOption } from "echarts";
 
 	let {
@@ -106,7 +106,13 @@
 			chart.setOption(currentOption, true);
 			// Force resize to handle tab visibility changes
 			chart.resize();
-			onLayout?.();
+			// untrack: the notify must not add the callback prop (or anything the
+			// handler touches) to this effect's dependencies. A parent's inline
+			// `onLayout` handler is recreated whenever its fragment re-renders —
+			// e.g. from the layout-tick state the handler itself bumps — so a
+			// tracked call re-runs this effect and loops (setOption → onLayout →
+			// tick++ → new handler → setOption …: effect_update_depth_exceeded).
+			untrack(() => onLayout?.());
 		}
 	});
 </script>
--- a/src/lib/game-detail/TechsTab.svelte
+++ b/src/lib/game-detail/TechsTab.svelte
@@ -116,112 +116,76 @@
 		TECH_NAMES[tech] ?? formatEnum(tech, "TECH_");
 
 	// ─── Chart option ─────────────────────────────────────────────────
-	const techDiscoveryChartOption = $derived(
-		techDiscoveryHistory.length > 0
-			? (() => {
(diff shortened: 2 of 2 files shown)


alcaras:
Updated per the Discord feedback: the chart now has a **Cumulative | Per Turn | Number of Techs** toggle (default Cumulative). Per Turn keeps numbers small so late-game tech adoptions stay visible; Number of Techs preserves the old chart and its per-point tech tooltip. Also fixes a latent `Chart.svelte` reactive loop the toggle exposed (details in the body).

becked:
Thanks for this — the goals are right and we want them. Closing on timing, not on the work: **#133 (tech-comparison) landed after this branched and substantially restructured `TechsTab.svelte`** — it extracted `TechComparison.svelte`, moved the owtt planner links into the table headers, and removed the old pivot / unique-techs sections. That turns a "trivial rebase" into reconciling against a file that no longer holds most of what this branch carries, so the cleaner path is to re-do the same goals fresh on top of `main`.

Plan for the replacement branch:

- **Same feature:** the Techs-tab chart becomes a 3-way toggle — **Cumulative Science** (default) · **Science per Turn** · **Number of Techs** — with the science rail annotating all three, exactly as designed here.
- **Reuse `createYieldChartOption`** (the shared helper the Yields tab already uses for its rate/cumulative science chart) for the two science modes rather than a bespoke option — keeps it DRY and consistent with the Yields tab. "Number of Techs" keeps the existing tech-count chart (with its per-point tech-name tooltip) as the third mode.
- **Keep your `Chart.svelte` `untrack()` fix** — that reactive-loop fix is real and independent of the restructure, so it carries over as-is. The repro notes on the loop made it easy to keep.

Superseded by a new branch/PR to follow.

PR #155, opened 2026-07-25
feat(stats): starting-leader archetype and trait win rates

alcaras:
Adds a **Leaders** category to the stats catalog — a Leaders tab on the tournament stats page, and a Leaders subtab in the user-corpus catalog — answering "which starting leader did you get, and how did it go".

tournament

Two charts, both the wins/losses stack the nation chart already uses, so **bar length is the distribution and the split is the win rate** — one bar answers both questions:

- **Starting archetype** — all ten, on the game's own archetype glyphs.
- **Starting leader traits** — the personality traits the leader began with, capped at 15 rows like the tech/law charts (there's a long tail of one-offs). No sprite ships for these, so they're name-only rows.

Over an all-humans corpus the overall rate is ~50% by construction (a 1v1 contributes one winner and one loser), so the signal is a row's deviation from that — noted in the bundle types so a future reader doesn't misread the aggregate.

## Why archetype and not the leader's name

Old World rolls the starting leader per game: across 405 cached public games the names are generated (`NAME_BELTUM_RIMENNI`, `NAME_HITTITE_MALE10`), 90 distinct name+archetype combos appear in 150 games with most showing up once, and the same name turns up under different archetypes. Charting the identity would be a list of one-offs; the archetype (10 values, `bArchetype` in `trait.xml`) is the axis with signal.

## Fixes a latent derivation bug (the reason the trait chart has data at all)

`derivePlayerSummary` collected starting traits with `acquired_turn <= 0`. The game stamps a starting leader's traits on **turn 1** — characters are born pre-game (negative `birth_turn`), but no trait row exists before the first played turn — so the filter matched nothing and `player_summaries.starting_ruler_traits` has been NULL for **every row ever written** (verified: 0 of 1041 rows in a local corpus, and empty in a production dump). Bounded at turn 1 it captures the archetype trait plus the personality trait(s) the leader starts with, and still excludes traits won later through events. Four unit tests pin the boundary, the other-player scoping, and the succession case.

**Backfill:** existing rows repopulate through the admin reindex sweep (`POST /v1/admin/games/:id/reindex` → `buildSummaryStatements`), which rebuilds `player_summaries` from the stored blob — no migration, nothing new to run. Caveat: saves parsed before the parser captured `characters` carry no leader data in the blob at all, so those need an owner-side reparse; the charts show `n` in the tooltip and an empty state that says so.

## Shape

- **No schema change.** Archetype is already a `player_summaries` column; the traits ride the existing JSON array and are split apart by the `_ARCHETYPE` suffix (the game's own `bArchetype` flag), so the archetype isn't double-counted in the trait tally.
- Two bundle fields (`startingArchetypeWinRate`, `startingTraitWinRate`) shaped exactly like `nationWinRate`, mirrored into the frontend types.
- One registry entry per chart → StatsView renders them through the existing spec loop; the tournament stats page (which renders charts directly, no registry) gets the matching tab.
- **Extracted** the wins/losses stacked bar out of `nations.ts` into `winLossStackedOption` in `charts/helpers.ts`, so all three outcome bars are one builder rather than three copies. `nationWinLossStackedOption` keeps its name and signature; the option it produces is unchanged.
- The `_ARCHETYPE` suffix rule now lives in one place — `formatArchetype`, `archetypeSpriteKey` and `isArchetypeTrait` in `$lib/utils/formatting`, called from the leader charts, `LeaderCard` and `MilitaryTab`. That replaces the stats-local `fmtArchetype` (dead and wrong: it stripped an `ARCHETYPE_` prefix; the values are `TRAIT_X_ARCHETYPE`) along with the copies of the same rule at the game-detail call sites. The Worker keeps its own regex — `cloud/` is a separate package with no shared module.
- Bumped `BUNDLE_SCHEMA_VERSION` 5→6: entries live 24h, and the frontend dereferences the new fields directly, so a bundle cached before the deploy would break it. Amended the constant's comment with that reasoning.

## Testing

- 4 new unit tests (`cloud/src/derive-player-summary.test.ts`); `npm run lint`, `svelte-check`, and the worker suite pass.
- Exercised end-to-end locally: 68 public games imported into local dev (fresh uploads derive traits on 136 of 140 rows), 24 linked to a fixture tournament — both surfaces in the screenshots are that corpus.

Unrelated heads-up, still open from #137: `cloud/src/tournament/canonical-map-options.test.ts` fails 4 tests on `main` (b39f690 re-baked `map-option-defs.ts` without the matching cloud-mirror entry for `MAP_OPTIONS_MULTI_DOTA_INNER_TERRAIN`).

### DIFF
--- a/cloud/src/derive-player-summary.test.ts
+++ b/cloud/src/derive-player-summary.test.ts
@@ -130,5 +130,8 @@ describe("derivePlayerSummary — starting leader", () => {
 		const summary = derive(blob);
 		expect(summary.starting_ruler_archetype).toBe("TRAIT_SCHOLAR_ARCHETYPE");
 		expect(summary.succession_count).toBe(2);
+		// The successor's own trait, acquired on the turn they took the throne,
+		// must not land in the starting leader's list.
+		expect(summary.starting_ruler_traits).toBeNull();
 	});
 });
--- a/cloud/src/derive-player-summary.test.ts
+++ b/cloud/src/derive-player-summary.test.ts
@@ -1,0 +1,134 @@
+import { describe, expect, it } from "vitest";
+import type { FullGameData, PlayerRosterEntry } from "./schemas/game";
+import {
+	buildSummaryGameContext,
+	derivePlayerSummary,
+} from "./derive-player-summary";
+
+// A minimal blob carrying only what the starting-leader derivation reads.
+// The real FullGameData is far wider (the Valibot schema treats most entries as
+// unknown), so the fixture is cast the same way the module narrows its input.
+function blobWith(over: {
+	characters?: Array<Record<string, unknown>>;
+	character_traits?: Array<Record<string, unknown>>;
+}): FullGameData {
+	return {
+		match_metadata: { total_turns: 60 },
+		player_history: [],
+		city_statistics: { cities: [] },
+		families: [],
+		characters: [],
+		character_traits: [],
+		completed_techs: [],
+		law_adoption_history: [],
+		...over,
+	} as unknown as FullGameData;
+}
+
+// One leader for player 0, crowned on turn 1 (a starting leader is born
+// pre-game — birth_turn is negative — and takes the throne on the first turn).
+const LEADER = {
+	xml_id: 7,
+	player_xml_id: 0,
+	is_royal: true,
+	became_leader_turn: 1,
+	death_turn: null,
+	archetype: "TRAIT_SCHOLAR_ARCHETYPE",
+};
+
+const PLAYER = { player_index: 0 } as unknown as PlayerRosterEntry;
+
+function derive(blob: FullGameData) {
+	return derivePlayerSummary(blob, PLAYER, buildSummaryGameContext(blob));
+}
+
+describe("derivePlayerSummary — starting leader", () => {
+	it("captures the traits the leader starts with, stamped on turn 1", () => {
+		const blob = blobWith({
+			characters: [LEADER],
+			character_traits: [
+				{
+					character_xml_id: 7,
+					trait_name: "TRAIT_SCHOLAR_ARCHETYPE",
+					acquired_turn: 1,
+				},
+				{
+					character_xml_id: 7,
+					trait_name: "TRAIT_INTELLIGENT",
+					acquired_turn: 1,
+				},
+			],
+		});
+
+		const summary = derive(blob);
+
+		expect(summary.starting_ruler_archetype).toBe("TRAIT_SCHOLAR_ARCHETYPE");
+		// Both the archetype trait and the personality trait land here; splitting
+		// the two is the stats aggregator's job, not the derivation's.
+		expect(JSON.parse(summary.starting_ruler_traits ?? "[]")).toEqual([
+			"TRAIT_SCHOLAR_ARCHETYPE",
+			"TRAIT_INTELLIGENT",
+		]);
+	});
+
+	it("excludes traits acquired after the game began", () => {
+		const blob = blobWith({
+			characters: [LEADER],
+			character_traits: [
+				{
+					character_xml_id: 7,
+					trait_name: "TRAIT_INTELLIGENT",
+					acquired_turn: 1,
+				},
+				// Won through an event mid-game — not part of the starting roll.
+				{ character_xml_id: 7, trait_name: "TRAIT_WARLIKE", acquired_turn: 26 },
+			],
+		});
+
+		expect(JSON.parse(derive(blob).starting_ruler_traits ?? "[]")).toEqual([
+			"TRAIT_INTELLIGENT",
+		]);
+	});
+
+	it("ignores other players' leaders", () => {
+		const blob = blobWith({
+			characters: [
+				LEADER,
+				{
+					...LEADER,
+					xml_id: 9,
+					player_xml_id: 1,
+					archetype: "TRAIT_ZEALOT_ARCHETYPE",
+				},
+			],
+			character_traits: [
+				{ character_xml_id: 9, trait_name: "TRAIT_ROMANTIC", acquired_turn: 1 },
+			],
+		});
+
+		const summary = derive(blob);
+		expect(summary.starting_ruler_archetype).toBe("TRAIT_SCHOLAR_ARCHETYPE");
+		expect(summary.starting_ruler_traits).toBeNull();
+	});
+
+	it("takes the earliest crowned leader when a succession happened", () => {
+		const blob = blobWith({
+			characters: [
+				{
+					...LEADER,
+					xml_id: 11,
+					became_leader_turn: 34,
+					archetype: "TRAIT_HERO_ARCHETYPE",
+				},
+				LEADER,
+			],
+			character_traits: [
+				{ character_xml_id: 11, trait_name: "TRAIT_BOLD", acquired_turn: 34 },
+			],
+		});
+
+		const summary = derive(blob);
+		expect(summary.starting_ruler_archetype).toBe("TRAIT_SCHOLAR_ARCHETYPE");
+		expect(summary.succession_count).toBe(2);
+	});
+});
--- a/cloud/src/derive-player-summary.ts
+++ b/cloud/src/derive-player-summary.ts
@@ -168,7 +168,13 @@ export function derivePlayerSummary(
 		const traits = blob.character_traits as CharacterTraitRow[];
 		const startTraits: string[] = [];
 		for (const t of traits) {
-			if (t.character_xml_id === startingRuler.xml_id && t.acquired_turn <= 0) {
+			// The game stamps a starting leader's traits — their archetype plus
+			// the personality trait(s) they begin with — on turn 1, not turn 0:
+			// characters are born pre-game (birth_turn is negative) but no trait
+			// row exists before the first played turn. The original `<= 0` bound
+			// therefore matched nothing, leaving this column empty for every row.
+			// Traits acquired later (events, aging) have higher turns and stay out.
+			if (t.character_xml_id === startingRuler.xml_id && t.acquired_turn <= 1) {
 				startTraits.push(t.trait_name);
 			}
 		}
--- a/cloud/src/stats/aggregate.ts
+++ b/cloud/src/stats/aggregate.ts
@@ -46,6 +46,11 @@ export const CHUNK_SIZE = 50;
 // charts/helpers.ts — kept in sync by eye (no shared module across packages).
 const ALL_NATIONS = "__all__";
 
+// A character's archetype is itself a trait, flagged bArchetype in the game
+// data and suffixed _ARCHETYPE (TRAIT_SCHEMER_ARCHETYPE). It rides along in
+// starting_ruler_traits; the leader charts split the two apart.
+const ARCHETYPE_TRAIT = /_ARCHETYPE$/;
+
 // Succession laws (the inheritance rule, one class flagged bSuccession in the
 // game data) are a different kind of law from civic laws — they start at turn 1
 // and change mainly via forced events — so they're excluded from the law-adoption
@@ -64,6 +69,7 @@ interface BaseRow {
 	is_human: number;
 	is_uploader: number;
 	starting_ruler_archetype: string | null;
+	starting_ruler_traits: string | null; // JSON array of strings
 	final_points: number | null;
 	cities_total: number | null;
 	fifth_city_turn: number | null;
@@ -206,7 +212,7 @@ async function loadBaseRows(
 		const res = await env.SHARE_DB.prepare(
 			`SELECT ps.game_id, ps.player_index,
 			        ps.nation, ps.family_classes, ps.is_human, ps.is_uploader,
-			        ps.starting_ruler_archetype,
+			        ps.starting_ruler_archetype, ps.starting_ruler_traits,
 			        ps.final_points, ps.cities_total,
 			        ps.fifth_city_turn, ps.tenth_city_turn,
 			        ps.is_winner,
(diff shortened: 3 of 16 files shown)


becked:
## Review

Reviewed against a local checkout of the branch (deps installed, suites run).

### Verification

| Check | Result |
|---|---|
| `cloud` unit suite | **166 passed**, incl. the 4 new tests |
| `cloud` `tsc --noEmit` | clean |
| `npm run lint` (root) | clean |
| `canonical-map-options.test.ts` | fails on this branch too — pre-existing, as the description says |

Spot-checks that hold up:

- The extraction reproduces the nations option **exactly** — `labelWidth - 8` = 132 matches the old literal, icon size 20 / fontSize 14 preserved, `COMMON_GRID` + `left: 140` preserved. The tooltip switches from `row.rate` to `wins/games`, numerically identical.
- The `BUNDLE_SCHEMA_VERSION` bump covers **both** corpora — there's one `cacheKeyToString`, version-prefixed for tournament and user keys alike.
- All ten archetype glyphs exist in `sprite-manifest.ts` (`traits/TRAIT_{BUILDER,COMMANDER,DIPLOMAT,HERO,JUDGE,ORATOR,SCHEMER,SCHOLAR,TACTICIAN,ZEALOT}`), so `archetypeIconUrl` resolves for every value it can be handed.
- The 15-row cap matches `tech.ts:40` and `laws.ts:93`.

The core work is sound: the derivation fix is real and correctly bounded, the tests pin the right boundaries, the extraction is faithful, and the cache bump is correctly reasoned. Findings 1–4 are the ones I'd want addressed before merge; 1 and 2 collapse into a single small change to `WinLossRow`.

---

### 1. `rate` is now dead payload in the bundle — and the refactor is what killed it

`winLossStackedOption` takes `WinLossRow = { key, games, wins }` and recomputes the percentage in the tooltip. That removed the **only** frontend consumer of `nationWinRate.rate`, and then this PR adds `rate` to two more arrays that nothing reads:

```
src/lib/stats/charts/cities.ts:26   r.rate   ← the only .rate read in all of src/lib/stats/
```

`tech.ts`, `families.ts` and `laws.ts` read `nationWinRate` for `.nation`/`.games` only. CLAUDE.md rule 5 is explicit about no exported API without a consumer.

**Recommendation:** put `rate` on `WinLossRow` and use it in the tooltip. That keeps the server-computed value authoritative (rule 6) and fixes finding 2 for free. Leave `nationWinRate.rate` alone — dropping it is a real shape change — but the two *new* fields shouldn't ship a value nothing dereferences.

### 2. Tooltip divides by `row.games` without the guard the server bothered to write

`src/lib/stats/charts/helpers.ts` — `Math.round((row.wins / row.games) * 100)` renders `NaN%` when `games === 0`. The replaced code read `row.rate`, which the aggregator computes as `games > 0 ? wins / games : 0`. It can't fire today (`bumpOutcome` only ever creates rows with `games >= 1`), but the guard existed and the refactor discarded it. Passing `rate` through resolves this and finding 1 together.

### 3. `_ARCHETYPE` handling is now duplicated across five sites

`fmtArchetype`'s new body is **byte-identical** to two existing call sites:

```
src/lib/stats/charts/helpers.ts:17          formatEnum(value.replace(/_ARCHETYPE$/, ""), "TRAIT_")
src/lib/game-detail/LeaderCard.svelte:49    formatEnum(c.archetype.replace(/_ARCHETYPE$/, ""), "TRAIT_")
src/lib/game-detail/MilitaryTab.svelte:248  formatEnum(c.archetype.replace(/_ARCHETYPE$/, ""), "TRAIT_")
```

Same story for the sprite key (`leaders.ts:16` vs `LeaderCard.svelte:55` vs `MilitaryTab.svelte:396`), and the suffix rule appears a third way as `.endsWith("_ARCHETYPE")` in `LeaderCard.svelte:93`. The code comment ("matching the game-detail leader cards") names the duplication rather than removing it — rules 1 and 4.

**Recommendation:** one `formatArchetype()` and one `archetypeSpriteKey()` within `src/`, called from LeaderCard, MilitaryTab, and the stats helpers. The `cloud/src/stats/aggregate.ts` regex has to stay its own copy — no shared module across packages, which that file already documents.

### 4. The empty-state copy is copy-pasted four times

Identical string in `registry.ts:46`, `registry.ts:56`, and twice in `src/routes/tournaments/[slug]/stats/+page.svelte`. It also breaks register with that page's existing empty states, which are all terse ("No completed games yet.").

**Recommendation:** export one `LEADER_EMPTY_MESSAGE` from `charts/leaders.ts` and reference it from all four sites.

### 5. The succession test doesn't assert the thing it's guarding

`derive-player-summary.test.ts:113` gives the *successor* (`xml_id: 11`) a `TRAIT_BOLD` at turn 34; the point of the test is that it doesn't leak into the starting leader's list. But it only asserts archetype and `succession_count`. One line pins it:

```ts
expect(summary.starting_ruler_traits).toBeNull();
```

### 6. Minor: `barChartHeight` doesn't know about the subtitle

These are the first registry specs to set `subtitle`, and `titled()` bumps `grid.top` 64 → 92 to make room while `height` stays `rows * 34 + 90`. So the leaders bars sit ~28px tighter than the nations bars at the same row count — and tighter again than the same chart on the tournament page, which renders it with `grid.top: 40` under an HTML `<h2>`. Cosmetic; either drop the subtitles or let `barChartHeight` account for one.

### 7. Minor: "No sprite ships for these" isn't quite true

`traits/TRAIT_STRENGTH` and `traits/TRAIT_WEAKNESS` are in the manifest. `crestAxisLabel` already falls back to name-only per value, so passing a `traitIconUrl` would light up whichever traits have art without disturbing the rest. Take it or leave it — but the comment should match the manifest either way.

### 8. Note: silent truncation at 15 rows

The trait chart drops the tail with no indication ("Games played, split by outcome" reads as complete). `tech.ts` and `laws.ts` do exactly the same, so this is consistent with the repo rather than a regression — flagging only because a "Top 15 by games played" subtitle would cost nothing.

---

### Operational — worth sequencing the deploy around

The two charts on the same tab will have **asymmetric coverage** at launch:

- `starting_ruler_archetype` was already populated (it feeds the existing `top_archetype` summary), so the archetype chart has data immediately.
- `starting_ruler_traits` has been NULL for every row ever written, so the trait chart is empty for the entire existing corpus until a bulk reindex sweep runs (`/admin/reparse` → `BulkReindexModal` → `buildSummaryStatements`).

The description covers the backfill mechanism, but the practical consequence is that shipping without running the sweep gives users a Leaders tab where one chart works and the other says "No leader data yet." Worth ordering as worker → bulk reindex → announce. And per the PR's own caveat, saves parsed before the parser captured `characters` won't fill from reindex at all — those need an owner-side reparse, so the trait chart stays partial indefinitely for that tail.

### Security & performance

Nothing to flag. No new endpoints or auth surface; the new fields are aggregate counts with no PII (`online_id`/`discord_id` are nowhere near this path); visitor scope narrowing is unchanged. Cost is one extra column on an existing `SELECT`, one extra pass over `selfRows`, and one `JSON.parse` per self row — the same shape as the existing `family_classes` handling.

---

alcaras:
Thanks — that's a genuinely useful review, and the "1 and 2 collapse into one change" observation was right. All of 1–7 are in `39cb772`.

**1 + 2 — `rate` unread, guard discarded.** Fixed together as you suggested: `WinLossRow` carries `rate` and the tooltip renders it, so the server value stays authoritative and the `games > 0` guard comes back with it. You're right that the extraction is what orphaned it — I removed the last reader while adding two more producers, which is exactly the shape rule 5 is aimed at.

**3 — five copies of the `_ARCHETYPE` rule.** Now one: `formatArchetype`, `archetypeSpriteKey`, `isArchetypeTrait`, called from LeaderCard, MilitaryTab and the leader charts; the stats-local `fmtArchetype` is gone rather than left as a wrapper.

One constraint worth recording, since it decided the location: `web/src/lib/utils/` symlinks **`formatting.ts` specifically**, not the directory. A new `utils/archetype.ts` would have compiled here and broken the frozen viewer's build the next time `web/` deploys, because game-detail is symlinked wholesale. So the helpers live in `formatting.ts`. The Worker keeps its own regex, as you said and as that file already documents.

**4 — empty-state copy.** One exported `LEADER_EMPTY_MESSAGE` referenced from all four sites, shortened to sit with the tournament page's terser copy: *"No leader data in these saves yet — they were parsed before characters were captured."*

**5 — the succession test.** Added `expect(summary.starting_ruler_traits).toBeNull()`. You're right that it was asserting everything except the leak it existed to prevent.

**6 — subtitle vs `barChartHeight`.** `barChartHeight(rows, { subtitle: true })` now adds the 28px `titled()` reserves, so the leaders bars match the nations bars at equal row count. Kept the subtitles rather than dropping them, since they're carrying the "split by outcome" explanation.

**7 — the sprite comment.** Corrected, and the trait chart now passes an icon resolver: `crestAxisLabel` already falls back per value, so Strength/Weakness light up without disturbing the rest.

**8 — silent truncation.** Leaving as-is, and I want to be transparent about why rather than let it read as an oversight: alcaras specifically asked for the subtitle to stay plain after an earlier draft said "Most-drawn traits". Since it matches `tech.ts`/`laws.ts` behaviour you'd already flagged as repo-consistent, I'd rather not re-add the qualifier without his say-so. Happy to put "top 15" back if you'd prefer it explicit — it's a one-word change and I'll take your call over mine here.

**Operational.** Agreed on worker → bulk reindex → announce; the sweep is what fills `starting_ruler_traits`, and I re-confirmed that locally after this refactor: reindexing 68 games takes the column from 0 to 136 of 140 human rows. And yes — saves parsed before the parser captured `characters` can't be recovered by reindex at all, so the trait chart stays partial for that tail until owners reparse. Worth saying plainly in the release note rather than letting it look like a bug.

Re-verified after the changes: `npm run lint`, `svelte-check`, `cloud tsc`, worker suite (`canonical-map-options` still fails on `main` — b39f690 re-baked `map-option-defs.ts` without the cloud-mirror entry for `MAP_OPTIONS_MULTI_DOTA_INNER_TERRAIN`, happy to send that four-line fix separately). Both charts re-checked in the browser at tournament and user scope.

alcaras:
Finding 8 done too — alcaras is happy with it, so the trait chart's subtitle is now **"Top 15 by games played, split by outcome"** (`95f77ba`). Only that chart carries the note; the archetype chart shows all ten, so nothing is hidden there.

That's all eight findings addressed.

becked:
Re-reviewed `39cb772` + `95f77ba`. **Findings 1–6 and 8 are properly addressed** — not just claimed; I checked the substance and re-ran the suites:

| Check | Result |
|---|---|
| `npm run lint` (root) | clean |
| `cloud` `tsc --noEmit` | clean |
| frontend `tsc --noEmit` | only the pre-existing `tournament/async-action.ts:51` error, untouched here |
| `cloud` unit suite | **167 passed**; the 4 failures are all `canonical-map-options.test.ts`, and this PR touches no map-option file — pre-existing, as you said |

Two things I want to confirm because they're the kind of reasoning that gets "cleaned up" later by someone who doesn't have the context:

- **The `web/` constraint is real.** `web/src/lib/utils/` contains exactly one entry — a symlink to `formatting.ts` — while `web/src/lib/game-detail` is a wholesale directory symlink. A new `utils/archetype.ts` would have compiled here and broken the frozen viewer's next build. Right call, and the comment recording it should stay.
- **Leaving `families.ts:20`'s `CREST_ARCHETYPE_${name}` alone was correct.** That's a family-class crest — same word, different domain concept. Applying the new helper there would have been a vocabulary error.

Finding 6's arithmetic is also exactly right: with +28 the StatsView leaders plot area (`rows*34 + 118 − 152`) equals the nations plot area (`rows*34 + 90 − 124`) at equal row count, and correctly *not* applied on the tournament page, which never calls `titled()`.

---

## Retracting finding 7 — that one was my error

Your original comment — *"No sprite ships for personality traits, so these are name-only rows"* — was **correct**, and I talked you out of a true statement.

I took `traits/TRAIT_STRENGTH` and `traits/TRAIT_WEAKNESS` from the sprite manifest as evidence that art ships for plain traits. It isn't. Checked against `Reference/`:

- Those two are `HUDIconTypes` enum entries — the last two in the sprite-repository icon enum (`Reference/Source/Base/Game/ClientCore/UI/IGameSpriteRepository.cs:94-95`). They surface in XML as **icon tokens**, e.g. `text-infos.xml:1048`: `<en-US>icon(TRAIT_STRENGTH)Strength</en-US>`, and in help text as inline icon+link pairs pointing at `CONCEPT_STRENGTH`/`CONCEPT_WEAKNESS` (`HelpText.Game.cs:2363,2367`).
- There is no `TRAIT_STRENGTH` or `TRAIT_WEAKNESS` **record** in `trait.xml`. The actual classification is a pair of booleans on each trait record — `bStrength`/`bWeakness` (`InfoBase.cs:6932-6933`, parsed at `7056-7057`). Of the 306 records in `Reference/XML/Infos/trait.xml`, 72 are strengths (Affable, Eloquent, Warlike, Brave, Pious, Legendary, …) and 23 are weaknesses (Cruel, Slothful, Corrupt, Timid, Drunk, Sickly, Inexperienced, …); none sets both. DLC traits live in the same file gated by `GameContentRequired` — no `trait-<suffix>.xml` — and none of the scenario mods' `trait-add.xml` files add a strength or weakness.

So no starting trait resolves art, and `traitIconUrl` is a resolver that can never resolve.

**What the change actually did — narrower than it looks.** In `winLossStackedOption` the branch is chosen by whether `iconUrl` was *passed*, not whether it resolves. With nothing resolving, `crestAxisLabel` degenerates to `{ interval: 0, align: "left", margin: 152, fontSize: 14, formatter: name }` plus an empty `rich: {}` — byte-for-byte the behavior of the plain branch it replaced. There is **no rendering change at all**, which is why it passed a visual re-check. What's left is a dead resolver plus a comment that now states something false.

**Suggested revert** in `src/lib/stats/charts/leaders.ts`: drop `traitIconUrl`, drop `iconUrl:` from `startingTraitWinLossOption`, and restore the original comment — worth sharpening with the above, since the manifest is exactly the trap:

> Only archetype glyphs ship as trait art. `traits/TRAIT_STRENGTH` and `traits/TRAIT_WEAKNESS` in the sprite manifest are `HUDIconTypes` entries, not trait records — a trait's strength/weakness classification is `bStrength`/`bWeakness` on the record itself (72 and 23 of the 306 in `trait.xml`). So these rows are name-only.

One check before writing "name-only" as an absolute: the manifest also carries bare `traits/TRAIT_{BUILDER,COMMANDER,DIPLOMAT,HERO,JUDGE,ORATOR,SCHEMER,SCHOLAR,TACTICIAN,ZEALOT}` as the archetype glyphs. If `trait.xml` defines a plain `TRAIT_JUDGE` record distinct from `TRAIT_JUDGE_ARCHETYPE`, that one could resolve.

---

## Two small items still open

**A. Finding 8's fix landed on only one of the two surfaces.** The registry subtitle now reads "Top 15 by games played, split by outcome", but the tournament page renders `<h2>Starting leader traits</h2>` with no subtitle mechanism — so the tournament Leaders tab still truncates silently at the same cap. Same shape as finding 4, same rule (wire all N parallel surfaces).

**B. Stale comment.** `src/lib/stats/charts/helpers.ts:129` still reads `// Display name for a row key (fmtNation, fmtArchetype, …)` — `fmtArchetype` no longer exists.

## Separate from this PR

The root cause of my finding-7 mistake looks like a bake issue worth its own ticket: the pipeline ships HUD icons inside the `traits/` sprite category, whose other members are all trait-id-keyed. That's what makes `SPRITE_MANIFEST["traits/" + traitName]` read like a valid lookup when it isn't. `TRAIT_STRENGTH`, `TRAIT_WEAKNESS` and probably `TRAIT_CONNECTION_TACTICAL_ADVANTAGE` are miscategorized.

Findings 1–6 and 8 are good to go; A and B are a five-minute pass, and 7 wants reverting.

---

Issue #143, opened 2026-07-19
refactor(stats): ScienceBreakdown.buildings holds all improvement tile yield, not buildings


becked:
## Summary

`ScienceBreakdown.buildings` (`src/lib/game-detail/science-techs.ts:439`) does not hold buildings. It holds **every improvement's unstaffed tile science**, rural and urban alike. The name also collides with an unrelated `buildings` field in the same file.

## Evidence

Checked the science-yielding improvements against `Reference/XML/Infos/improvement.xml`:

| Improvement | Science | `bUrban` |
| --- | --- | --- |
| Monastery | +2 | `1` — urban |
| Shrine | +1 | `1` — urban |
| Windmill | +2 | absent — **rural** |
| Watermill | +2 | absent — **rural** |
| Grove (on luxury) | +2 | absent — **rural** |

Three of the five sources in the row are rural tile improvements. Libraries and the Musaeum are not in this bucket at all — they have `flat: 0, pct: > 0`, so they route to `modifiers`.

In Old World's vocabulary a player *builds* an *improvement*; "building" is not a separate category from "improvement". The split this bucket actually encodes is **tile yield vs. the specialist working the tile**, which the code's own comment states at `science-techs.ts:645-650`: the improvement's unstaffed science lands here, while a specialist's boost to that tile (a Gardener doubling a Grove) lands on the specialist's row.

## Name collision

`buildings` already means something different 265 lines earlier — `ScienceTechUsage.buildings: NamedCount[]` (`science-techs.ts:174`, populated at `:309` via `buildingName()`) is the list of improvements in a tech line, used for tooltip rendering. Two unrelated meanings for one identifier in one file.

## Impact

No incorrect numbers. The arithmetic is right; only the naming is wrong. The cost is comprehension: a reader looking for where rural improvements are accounted for will not look in `buildings`.

## Suggested fix

Rename the `ScienceBreakdown` field to something describing the actual axis (e.g. `improvementYield` / `tileYield`). Blast radius is contained — roughly 10 sites in `science-techs.ts` plus the key and label in `src/lib/game-detail/TechsTab.svelte:558`; nothing crosses beyond those two files.

The user-visible label is a separate decision. It currently reads "Buildings & resources", which is already vaguer than the field name and arguably closer to correct.

Reported by a user (Konstant) while discussing the Science Sources table.

PR #157, opened 2026-07-25
feat(stats): wonders — build timing, builder win rate, and eligibility-aware build rate

alcaras:
Adds a **Wonders** category to the stats catalog — a Wonders tab on the tournament stats page and a Wonders subtab in the user-corpus catalog.

wonders

One row per wonder on a **turn axis**, so the chart reads as the arc of a game: the bar is the P25–P75 span of when the wonder gets built, the marker is the median, and everything else hangs off that.

- **The marker carries the builders' outcome** in three buckets — solid copper (mostly won, ≥60%), solid muted (mixed), hollow ring (mostly lost, <40%). Buckets rather than a gradient because per-wonder samples are small, and shape backs up color so the states survive a dark background.
- **The trailing label is the builders' win rate with its sample** (`83% won (5/6)`) — the denominator matters when a wonder has one build.
- **Build rate is in the tooltip** along with the culture tier and exact turn stats.

In the screenshot the Ziggurat is the story: a wide early span, hollow marker, `0% won (0/6)` — built often, never by a winner — directly above the Yazilikaya's `83% won (5/6)`.

## The denominator takes two gates

"Nobody built it" and "nobody could build it" are different findings, and separating them needs both:

1. **The wonder has to be in the game.** Old World enables only a subset of the wonders per game — a base save disables **15 of the 28**, and across a 68-game sample the enabled pool averages **13.3**. Nothing captured that list before, so a culture-only denominator counted players as "eligible" for wonders that were never on the board (it inflated The Jerwan Aqueduct's denominator from 26 to 46).
2. **The player has to have the culture for it.** Wonders carry **no tech prereq** — verified across all 28 base-game entries — only a `<CulturePrereq>`, so eligibility means a city that reached that level.

## Shape

- **Parser 2.12.0** records the save's game-level `<ImprovementDisabled>` block as `game_details.disabled_improvements`; `game_wonder_pool` (migration 0034) stores the enabled complement per game. **Blobs older than 2.12.0 carry no pool and contribute no eligibility at all** rather than a wrong one — coverage grows as saves are reparsed.
- `player_summaries.best_culture_level` is the highest level any city the player held reached (culture only climbs, so the end-state snapshot is also their best). Keyed on `owner_player_xml_id` so a mirror match doesn't credit one side with the other's cities.
- `wonder_events` indexes the blob's existing `player_wonders`, keyed `(game_id, wonder)` since a wonder is unique per game. Both tables are written on upload and rebuilt by the **admin reindex sweep** — the same backfill path the other derived tables use.
- `bake:wonders` emits the wonder → culture-prereq map and the ascending culture ranking to **both** the frontend and the Worker (the law-classes dual-output pattern), so eligibility is computed from the game's own data rather than a hand-kept list.
- Wonders that render another entry's art (The Acropolis draws the Parthenon, the Via Recta Souk the Grand Bazaar) resolve through the existing baked `IMPROVEMENT_ICON` alias map.
- `BUNDLE_SCHEMA_VERSION` 5→6: entries live 24h and the frontend dereferences the new field directly.

**Deploy note:** this bumps `PARSER_VERSION`, so Worker before frontend as usual, and existing saves need a reparse before their wonder pool exists.

## Testing

- 3 new unit tests for `best_culture_level` (highest across cities, other players' cities ignored, null when none).
- `npm run lint`, `svelte-check`, `tsc`, and the worker suite pass.
- Exercised end-to-end locally on 68 imported public games: 904 pool rows (13.3 enabled/game), 203 wonder events, and the admin reindex sweep run over all of them — the screenshots are that corpus.

Independent of #155 (leaders); they touch neighbouring lines in the stats registry, so whichever lands second gets a small rebase, which I'll handle.

Unrelated heads-up, still open: `cloud/src/tournament/canonical-map-options.test.ts` fails 4 tests on `main` (b39f690 re-baked `map-option-defs.ts` without the matching cloud-mirror entry).


alcaras:
Self-review pass before you look at this — two fixes pushed in 4abf681:

**Builder misattribution (real bug).** `derivePlayerWonders` resolves a wonder's builder by finding its improvement on the map and reading the tile owner; when that fails it falls back to `player_id: 0` with a null nation. Indexing those credited whoever holds player index 0 with someone else's wonder *and its win*. Measured across the 407-game public sample: 26 of 1304 wonder rows (2.0%), in 16 games, every one landing on index 0 — systematic, not noise. Those rows are now skipped at index time (the null nation is the parser's own marker).

**Test coverage for the eligibility math**, which is the part of this PR worth pinning. New integration tests drive the real upload pipeline and read back through the stats endpoint: both gates open, an enabled wonder above the player's culture level, a disabled wonder absent entirely, an eligible wonder nobody built (the denominator case), a pre-2.12.0 blob contributing no eligibility, and the unresolved-builder case above. The upload fixture gained wonder/city/disabled-improvement inputs to drive them.

Also made the culture-prereq lookup explicit in the eligibility loop instead of relying on `cultureRank`'s `-1` for an unknown wonder, which would have read as "eligible for everyone" if a mod wonder ever reached the pool.

Worker suite green (the 4 `canonical-map-options` failures are the pre-existing mirror drift on `main`).

alcaras:
Carried your #155 review forward — `14aaca5`. Two of those findings were patterns rather than one-offs, and this branch had both:

- **Finding 4** — the empty-state copy was typed twice (registry spec + tournament page). One exported `WONDER_EMPTY_MESSAGE` now, referenced from both.
- **Finding 6** — the wonder spec sets a subtitle, and `titled()` reserves 28px for one while `barChartHeight` didn't know about it, so the bars drew tighter than a subtitle-less chart at the same row count. `barChartHeight` takes `{ subtitle }`, using the **same signature and wording as the #155 change**, so whichever of the two lands second resolves to identical text rather than a real conflict.

Findings 1–2 (`rate` passthrough) don't apply here — the wonder chart is a timeline, not a `winLossStackedOption` caller, and it already reads the server's `rate` and `win_rate` rather than recomputing them. Finding 3's archetype helper is leader-only.

One dependency to flag: if #155 merges first, this branch picks up its `barChartHeight` change and my version drops out on rebase — no action needed from you either way.

becked:
## Review

Reviewed at `14aaca5` against base `b30fc31`.

### Verification

| Check | Result |
|---|---|
| `npx svelte-check` | 0 errors, 0 warnings |
| `npm run lint` | clean |
| `cloud/ npx tsc --noEmit` | clean |
| `cloud/ npm test` | 467 passed, 4 failed |
| New tests only | 3 unit + 4 integration, all pass |

The 4 failures are `src/tournament/canonical-map-options.test.ts` — pre-existing, and this PR touches neither `canonical-map-options.ts` nor `map-option-defs.ts`. Matches the heads-up in the description.

### What holds up well

- **The migration mirrors `tech_events`/`law_events` exactly** — `REFERENCES games(game_id) ON DELETE CASCADE`, so the re-import path's `INSERT OR REPLACE INTO games` cascades with no new delete needed, and `handleAdminReindex` adds only the explicit deletes it actually requires.
- **`BUNDLE_SCHEMA_VERSION` 5→6 with the rule rewritten.** "Adding a field counts" fixes a genuinely misleading comment — the old wording would have told the next contributor to skip the bump.
- **The null-nation filter** in `buildWonderEventStatements` is well-targeted: `derivePlayerWonders` falls back to `player_id: 0` when the tile lookup fails, so indexing those would credit whoever holds index 0. Documented and tested.
- **Dual-output bake matches the law-classes pattern** — diffed the two generated files, byte-identical.
- **Vocabulary is clean** — `improvement`/`wonder`/`CulturePrereq`/`bWonder` all match `Reference/XML`; sidesteps the #143 `building` trap.
- `loadWonderEvents` mirrors `loadTechEvents` verbatim, down to the `is_human = 1` join.

---

### Findings

Ranked by whether a number comes out wrong.

#### 1. `rate` is wrong corpus-wide at launch, not just the tooltip (High)

`eligible` conflates two states: "zero players qualified" and "we have no pool data for this game." When it's the second, `aggregate.ts`

```js
rate: eligible > 0 ? turns.length / eligible : 0,
```

emits **0** for wonders that were built. That's the number the feature exists to produce.

It happens because `wonderStats` is keyed off `eligibleByWonder ∪ wonderTurns`, so a game with no `game_wonder_pool` row still contributes builds — and the corpus isn't filtered by parser version (`resolve.ts` has no such filter), so pre-2.12.0 games stay in and contribute `wonder_events` (`player_wonders` predates this PR) with no pool behind them.

Since `PARSER_VERSION` bumps, **every existing save is in that state until reparsed**, so this is the default rendering on deploy day rather than a corner case.

You'd notice it via the tooltip, which reads self-contradictingly in that state — `src/lib/stats/charts/wonders.ts`:

```js
row.eligible === 0
  ? "Not available to anyone in these games"
  : `Built by ${row.built} of ${row.eligible} eligible (${pct(row.rate)}%)`,
```

then unconditionally appends `Median turn 30 …` / `Builders won 1/1 (100%)` when `built > 0`. The `excludes games whose blob predates the parser` test asserts exactly this row shape (`eligible: 0, built: 1`).

Suggest making `eligible` `Nullable<number>` (null = no pool coverage, distinct from a real zero) so `rate` can be null rather than 0, with a third tooltip branch for it.

#### 2. `win_rate` attributes outcomes to the wrong player (High)

`src/lib/parser/derive/player-wonders.ts` resolves the builder from `t.ownerPlayerXmlId` on the end-state map, so a **captured** wonder is attributed to the capturer. `wins/built` is then computed over the wrong set of players, and winners capture more — so the bias runs in the same direction the chart is measuring. The "built often, never by a winner" read is precisely the shape a capture artifact would produce.

I checked whether the blob carries a better signal: `WONDER_ACTIVITY` fires for *every* player (which is why the parser goes to the tile at all), so `EventLog.playerXmlId` can't identify the builder either. There's no cheap correct fix available — which leaves stating the limitation in the subtitle/tooltip, or relabeling toward "held by". That's a caveat the chart needs, not a wording preference.

This is pre-existing parser semantics, but this PR is the first thing to build a win-rate claim on top of it.

#### 3. The eligibility denominator is survivor-weighted (Medium)

`derive-player-summary.ts` reads `c.owner_player_xml_id === idx` against the end-state city list. A player eliminated on turn 40 of an 80-turn game holds no cities, gets `null`, hits `rank < 0`, and contributes **zero** eligibility for every wonder. Those are disproportionately the players who didn't build wonders, so dropping them shrinks the denominator and inflates every build rate. In a 1v1 tournament corpus — the tab this PR adds — conquest finishes are common enough that this isn't a tail case.

The comment is also load-bearing and not quite right:

> culture only ever climbs, so the end-state snapshot is also the best level the player ever held

True per *city*, not per *player* — cities change hands. It also makes `built > eligible` (and `rate > 1`, rendered as e.g. "150%") reachable. Falling back to `first_owner_player_xml_id`, or taking the max over cities the player ever held, would cover most of it; at minimum the comment shouldn't promise a guarantee that isn't there.

#### 4. Two raw hexes in a chart module (Low)

`wonders.ts` uses `#9b8f7d` for the mixed bucket and `#cfc9bd` for the ring border. `src/lib/stats/charts/` otherwise has zero inline hexes — the two outcome colors are named in `helpers.ts` (`WIN_COLOR`/`LOSS_COLOR`), and `#cfc9bd` already exists as `TOOLTIP_TEXT` in `src/lib/game-detail/EventRail.svelte`. Worth promoting both next to the existing pair.

#### 5. The tournament page drops the subtitle that explains the encoding (Low)

`registry.ts` gives the user-stats chart "P25–P75 turn span, median colored by the builders' outcome". `src/routes/tournaments/[slug]/stats/+page.svelte` renders a bare `<h2>Wonders</h2>` — same chart, same three-bucket legend, no key to it. (The `barChartHeight(wonders.length)` call there is correctly *without* `{ subtitle: true }`; the fix is the subtitle text, not the height.)

#### 6. Stale comment (Nit)

`maxTurn * 1.02` is commented "Axis headroom for the trailing build-rate labels", but those labels sit outside the grid (`grid.right: 110`) and carry the *win* rate, not the build rate.

---

### Deploy

Worker-before-frontend is right, and the `BUNDLE_SCHEMA_VERSION` bump covers the 24h cache window. `LineChart` and `LegendComponent` are both registered in `src/lib/echarts.ts`, so the new series type survives the recent tree-shake (c1676f8) — worth noting since this is the first stats chart to use a line series for markers. `disabled_improvements` is capped at `maxLength(200)` against 129 baked improvements, so base-game saves have room.

Finding #1 is the one I'd want resolved before merge — it isn't a rare edge, it's what every user sees until the reparse sweep completes.


---

PR #159, opened 2026-07-25
feat(stats): capital family, family city footprint, and families on the tournament page

alcaras:
Three things about families, all on the same panel — which one runs the capital, how much of the empire each ends up holding and where it fell in the founding order, and (finally) any of this on the tournament stats page.

families

## Capital family

Which of a player's three families runs the capital is a different question from which three they have: the capital is the city every early bonus compounds through. The existing panel only answered the latter.

New `player_summaries.capital_family_class` (migration 0035), derived from the blob's `is_capital` city and keyed on `owner_player_xml_id` so a mirror match can't credit one side with the other's capital, plus a stacked wins/losses bar — length is how often that class held the capital, the split how those games ended.

Across a 407-game public sample (805 human capitals) the spread is real: Traders **57%** (n=138) and Riders 52% at the top, Clerics **37%** (n=65) and Artisans 40% at the bottom.

## Family city footprint, and founding order

Knowing a player ran Hunters says nothing about whether they leaned on them. `player_family_cities` (migration 0036) records, per player and family class, how many end-of-game cities it held and when its first was founded. The bundle turns those into the mean share of the empire that class ran (on the bar) and, by ranking each player's families on that first-founding turn, how often it was their **1st / 2nd / 3rd** family (in the tooltip).

tooltip

That split separates roles the pick count can't: Maurya's Sages are the 1st family 80% of the time, while Maurya's Riders are almost never first (0% / 75% / 25%).

## Two consistency fixes

- **The per-nation chart plotted pick rate and win rate as two independent percentage bars** — the only outcome chart in the catalog not using the shared stacked encoding (length = games, split = outcome). It now matches its siblings; the normalized pick rate moved into the tooltip beside the counts it already carried.
- **The tournament stats page never rendered families at all**, even though `familyByNation` has been in the bundle it already fetches (47 rows over 36 games on the live tournament). Rather than fork a second chart, `FamilyStatsPanel` and its option builders are typed against `ChartBundleCore` and the tournament page renders the same component the player page does. That also sets up the general rule that a stat on one stats surface should exist on the other.

## Testing

- 3 new unit tests for the capital-family derivation (reads the capital's class, ignores another player's capital, null when there's no capital or no family on it).
- `npm run lint`, `svelte-check`, `tsc` and the worker suite pass.
- Exercised end-to-end on 68 imported public games: 395 `player_family_cities` rows across 67 games, then the admin reindex sweep over all of them — the screenshots are that corpus.

## Notes for review

- Both new tables/columns backfill through the existing admin reindex sweep; no new operational surface.
- `BUNDLE_SCHEMA_VERSION` 5→6 — entries live 24h and the frontend dereferences the new fields directly.
- **Known limitation, stated in the code:** share and founding order come from the end-of-game city snapshot, so a family whose cities were all captured or razed contributes nothing, and "first founded" is the earliest *surviving* city. The save carries no intermediate ownership history to do better.
- Independent of #155 (leaders) and #157 (wonders); all three touch the stats registry, so whichever lands last takes a small rebase, which I'll handle.

Unrelated heads-up, still open: `cloud/src/tournament/canonical-map-options.test.ts` fails 4 tests on `main` (b39f690 re-baked `map-option-defs.ts` without the matching cloud-mirror entry).

### DIFF
--- a/cloud/migrations/0037_capital_family_class.sql
+++ b/cloud/migrations/0037_capital_family_class.sql
@@ -1,0 +1,18 @@
+-- The family class that holds the player's capital.
+--
+-- A player's three families are already recorded (player_summaries
+-- .family_classes), but which of them runs the capital is a distinct and
+-- earlier decision: the capital is the city every early bonus compounds
+-- through, so its family's class shapes the opening in a way "we had Traders
+-- somewhere" does not.
+--
+-- Derived from the blob's city_statistics: the player's is_capital city, read
+-- through owner_player_xml_id so a mirror match can't credit one side with the
+-- other's capital. NULL when the player ended with no capital (eliminated, or
+-- a pre-2.6.0 blob whose cities carry no family class). Backfilled for
+-- existing rows by the admin reindex sweep, which rebuilds player_summaries
+-- from the stored blob.
+ALTER TABLE player_summaries ADD COLUMN capital_family_class TEXT;
+
+CREATE INDEX idx_summaries_capital_family
+    ON player_summaries(capital_family_class);
--- a/cloud/migrations/0037_capital_family_class.sql
+++ b/cloud/migrations/0037_capital_family_class.sql
@@ -12,7 +12,8 @@
 -- a pre-2.6.0 blob whose cities carry no family class). Backfilled for
 -- existing rows by the admin reindex sweep, which rebuilds player_summaries
 -- from the stored blob.
+--
+-- Deliberately unindexed: the stats layer selects player_summaries by game_id
+-- and buckets by this column in JS (loadBaseRows / buildChartBundle), so an
+-- index on it would be write cost on every upload and reindex for no read.
 ALTER TABLE player_summaries ADD COLUMN capital_family_class TEXT;
-
-CREATE INDEX idx_summaries_capital_family
-    ON player_summaries(capital_family_class);
--- a/cloud/migrations/0038_player_family_cities.sql
+++ b/cloud/migrations/0038_player_family_cities.sql
@@ -1,0 +1,29 @@
+-- Per-player, per-family-class city footprint.
+--
+-- player_summaries.family_classes already records *which* three classes a
+-- player ran, but not how much of the empire each ended up holding, or how
+-- early it arrived. Those are the two questions that separate a family you
+-- leaned on from one you merely had: a class holding half your cities from
+-- turn 20 played a different game than one that showed up with a single late
+-- conquest.
+--
+-- One row per (game, player, family class) — at most three per player, so this
+-- stays small. Derived from the blob's city_statistics at index time, keyed on
+-- owner_player_xml_id so a mirror match can't credit one side with the other's
+-- cities. `cities` counts the player's end-of-game holdings of that class (the
+-- save is an end-state snapshot, so a razed or lost city isn't in it), and
+-- `first_founded_turn` is the earliest founding among them, which the stats
+-- layer turns into "how much of the game it was present for".
+--
+-- Backfilled for existing games by the admin reindex sweep, which rebuilds the
+-- derived tables from the stored blob.
+CREATE TABLE player_family_cities (
+    game_id TEXT NOT NULL REFERENCES games(game_id) ON DELETE CASCADE,
+    player_index INTEGER NOT NULL,
+    family_class TEXT NOT NULL,
+    cities INTEGER NOT NULL,
+    first_founded_turn INTEGER,
+    PRIMARY KEY (game_id, player_index, family_class)
+);
+
+CREATE INDEX idx_family_cities_class ON player_family_cities(family_class);
--- a/cloud/migrations/0038_player_family_cities.sql
+++ b/cloud/migrations/0038_player_family_cities.sql
@@ -17,6 +17,11 @@
 --
 -- Backfilled for existing games by the admin reindex sweep, which rebuilds the
 -- derived tables from the stored blob.
+--
+-- No secondary index: the only reader (loadFamilyCities in
+-- stats/aggregate.ts) selects by game_id, which the primary key's leading
+-- column already serves. One on family_class would cost every upload and
+-- reindex a write for a lookup nothing performs.
 CREATE TABLE player_family_cities (
     game_id TEXT NOT NULL REFERENCES games(game_id) ON DELETE CASCADE,
     player_index INTEGER NOT NULL,
@@ -25,5 +30,3 @@ CREATE TABLE player_family_cities (
     first_founded_turn INTEGER,
     PRIMARY KEY (game_id, player_index, family_class)
 );
-
-CREATE INDEX idx_family_cities_class ON player_family_cities(family_class);
--- a/cloud/src/CLAUDE.md
+++ b/cloud/src/CLAUDE.md
@@ -32,6 +32,10 @@ Run from `cloud/`: `npm test` (both projects), or filter with `--project unit` /
 
 When you bump `PARSER_VERSION` in `src/lib/parser/types.ts`, also add the new string to `KNOWN_PARSER_VERSIONS` in `cloud/src/schemas/game.ts` with a one-line changelog entry above the set. The Worker rejects unknown versions with `INVALID_BLOB: Unknown parser_version`, so a frontend that ships ahead of the Worker breaks all uploads. **Deploy ordering: Worker first, frontend second.**
 
+## Bumping `BUNDLE_SCHEMA_VERSION`
+
+The stats bundle's cache-flush lever (`cloud/src/stats/cache.ts`). **Don't edit the constant** — it's derived as the highest key of the `BUNDLE_SCHEMA_CHANGELOG` table right above it, so you bump it by adding an entry saying what changed. Bump whenever the `ChartBundle` gains or loses a field: entries live 24h and consumers dereference every field directly, so a frontend deployed behind the Worker would read a pre-field bundle and blow up on it. A chart drawn from fields the bundle already carries needs no bump. The value is deliberately **not** restated in `docs/aggregate-statistics.md` — it sat four versions stale there across four stats PRs, so don't reintroduce it.
+
 ## Events retention
 
 A nightly cron (03:47 UTC, `[triggers]` in `cloud/wrangler.toml`) prunes the `events` table per the policy in `cloud/src/retention.ts`: 24h for rate-limit counters, 90d for general audit, never for tournament audit. Two rules when touching events:
--- a/cloud/src/derive-player-summary.test.ts
+++ b/cloud/src/derive-player-summary.test.ts
@@ -26,7 +26,10 @@ function blobWith(over: {
 	} as unknown as FullGameData;
 }
 
-const PLAYER = { player_index: 0 } as unknown as PlayerRosterEntry;
+const PLAYER = {
+	player_index: 0,
+	nation: "NATION_ROME",
+} as unknown as PlayerRosterEntry;
 
 function derive(blob: FullGameData) {
 	return derivePlayerSummary(blob, PLAYER, buildSummaryGameContext(blob));
@@ -139,7 +142,13 @@ describe("derivePlayerSummary — starting leader", () => {
 
 // A city at the given culture level, owned by player 0 unless stated. `held`
 // defaults to the sole owner — pass it to describe a city that changed hands.
-function city(culture: string | null, owner = 0, held = [owner]) {
+// `family` is the class running it; `capital()` wraps the capital case.
+function city(
+	culture: string | null,
+	owner = 0,
+	held = [owner],
+	family: string | null = "FAMILYCLASS_TRADERS",
+) {
 	return {
 		city_id: 1,
 		owner_nation: "NATION_ROME",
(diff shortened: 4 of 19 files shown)


alcaras:
Self-review pass before you look at this — two fixes pushed in 24dbbad.

**Captured cities were backdating founding order (real bug).** A captured city keeps the turn its *original* owner founded it, and I was folding those into each family's first-founded turn — so a family could be dated to a turn the player had nothing to do with, and the 1st/2nd/3rd split in the tooltip is exactly what that feeds. Measured across a 250-game public sample: it moves the first turn for **0.8% of player-class rows** and flips the founding order outright for **1.4% of players**. `first_founded_turn` now counts only cities the player founded; a family that arrived purely by conquest has no founding turn and contributes no slot. Captured cities still count toward the family's share of the empire, which is the right split of the two questions.

I also checked the adjacent worry — whether captures can introduce a *fourth* family class and push a real family out of the three slots. They can't: across the local corpus no player holds more than three classes (127 players at three, 7 at two), so a captured city is reassigned to one of the player's own families.

**Share weighting.** The cross-nation mean weighted each nation by its pick count, but the per-nation mean's own denominator is the picks that carried city data. Now weighted by `share_samples`, so a nation with sparse city data can't pull the aggregate.

**Tests:** 4 new integration tests through the real upload pipeline — share splitting across classes, founding order from own foundings, a captured city failing to backdate a family, and the capital's class. The upload fixture gained city seeding (with `capturedFrom`) and mirrors seeded cities into the blob's families list.

Worker suite green (the 4 `canonical-map-options` failures are the pre-existing mirror drift on `main`).

alcaras:
Carried your #155 review forward — `24d5367`.

**Finding 3, in miniature.** The capital-family and by-nation charts each hand-rolled the same wins/losses stack. One local builder now, with the optional trailing label as its only variation.

I deliberately kept it **inside `families.ts`** rather than lifting it to `charts/helpers.ts`: #155 introduces a catalog-wide `winLossStackedOption`, and both of these route through it on rebase once that lands. Extracting into `helpers.ts` here would collide with #155 for no gain — happy to do it that way instead if you'd rather this PR own the shared version.

Findings 4 and 6 don't bite here: the capital chart is panel-rendered rather than a registry spec, so no subtitle sizing is involved, and the empty states are per-condition rather than one repeated string. Finding 2's guard is intact — the capital chart reads the server's `rate`, and the by-nation win rate keeps its `count > 0` check.

Re-verified after the refactor: lint, svelte-check, worker suite, and both charts in the browser at tournament and player scope.

becked (commented):
## Overview

Three related additions on the Families panel, plus two consistency fixes:

- **`player_summaries.capital_family_class`** (migration 0035) — which family class holds the player's capital, derived from the blob's `is_capital` city, keyed on `owner_player_xml_id`. Surfaced as a new stacked wins/losses bar.
- **`player_family_cities`** (migration 0036) — per (game, player, family class): end-of-game city count and earliest *self-founded* turn. Feeds `avg_share` (bar label) and `slot_counts` 1st/2nd/3rd founding order (tooltip).
- **Families on the tournament stats page** — `FamilyStatsPanel` and its builders retyped to `ChartBundleCore`, new `families` tab.
- The per-nation family chart moves from two independent percentage bars to the catalog's shared wins/losses stacked encoding.

The design work here is strong: the mirror-match keying on `owner_player_xml_id`, the "captured cities don't backdate a founding" rule (with a measured justification and a test), the `share_samples` field so cross-nation recombination weights correctly, the `BUNDLE_SCHEMA_VERSION` bump reasoning, and the stated end-state-snapshot limitation are all better than typical.

Verified locally: `npm run lint` clean, `svelte-check` 0 errors, cloud suite **466 passed** (the one failing file, `canonical-map-options.test.ts`, is the pre-existing `main` breakage the PR already calls out).

There is one release-blocking bug.

---

## 🔴 Blocking: the family-cities insert overshoots D1's parameter cap

`cloud/src/games.ts:697` reuses `TECH_LAW_ROWS_PER_INSERT` — a constant sized for **4 columns** (`Math.floor(99/4) = 24`, `games.ts:105`) — for an insert with **5 columns**:

```ts
const columns = ["game_id","player_index","family_class","cities","first_founded_turn"]; // 5
for (const chunk of chunked([...tally.values()], TECH_LAW_ROWS_PER_INSERT))  // 24 rows
```

24 rows × 5 = 120 bindings, past D1's cap. This is exactly the hazard the comment at `games.ts:98-101` warns about for `GPT_COLS`.

**Reproduced against Miniflare** with this PR's own fixture (uploads seeding N `(owner, family_class)` pairs):

```
row count -> upload status: { 18: 201, 19: 201, 20: 201, 21: 500, 24: 500 }
POST /v1/games → 500 {"error":"Database write failed","code":"D1_FAILED"}
```

**≥21 rows fails the whole upload**, not just the new table — `allStatements` is one atomic `batch()`. `buildFamilyCityStatements` tallies *every* owner including AI, so 7 players × 3 families = 21 rows. That's an ordinary vs-AI or large-MP save, and the same builder runs in `handleAdminReindex`, so the backfill sweep 500s on those games too. The end-to-end validation in the PR description (395 rows / 67 games ≈ 5.9 rows/game) was all 1v1 saves, which is why it didn't surface.

**Fix** — mirror the existing `GPT_COLS` pattern:

```ts
const FAMILY_CITY_COLS = 5;
const FAMILY_CITY_ROWS_PER_INSERT = Math.floor(D1_MAX_PARAMS / FAMILY_CITY_COLS); // 19
```

Worth pairing with: filter the tally to human roster entries at write time. `loadFamilyCities` already joins `WHERE ps.is_human = 1`, so every AI row written is dead storage — and it's what drives the row count up.

An integration test with ≥21 rows would lock this down; the fixture supports it today (`cities: [{owner, familyClass, foundedTurn}]` with owner 0..7).

---

## Correctness & data modeling

- **Both new indexes are unused.** No query in the diff (or repo) filters or sorts on `capital_family_class` or `player_family_cities.family_class` — `loadBaseRows` selects by `game_id` and buckets in JS; `loadFamilyCities` hits the PK prefix. Per YAGNI / "no speculative code", drop `idx_summaries_capital_family` and `idx_family_cities_class` — they're write cost on every upload and reindex for no read.
- **Migration numbering skips 0034.** `main` ends at 0033. Intentional (reserved for #155/#157)? A gap is functionally harmless but invites a same-number collision on rebase.
- **Multiple capitals: last-wins, no tie-break.** `derive-player-summary.ts:236` overwrites `capital_family_class` on each match rather than taking the first. `is_capital` is a presence check on `<Capital/>` (`parsers/cities.ts:137`), so if a captured capital ever retains the element while owned by the capturer, attribution silently depends on array order. Cheap to make deterministic (`if (capital_family_class === null && …)`), or note why it can't happen.
- **`avg_share` denominator is family-assigned cities, not all cities.** Cities with `family_class == null` are skipped entirely, so `ownCities` excludes them. The `types.ts` comment ("share of the player's end-of-game cities") and the bar label ("X% of cities") both overstate slightly. Also note it's a different denominator than `player_summaries.cities_total`, which still matches by `owner_nation` — so on a mirror match the new stat is right and `cities_total` isn't. Worth a sentence in the comment.

## Project-fit / conventions

- **`outcomeSeries` is a third copy of the wins/losses stack.** `nationWinLossStackedOption` (`charts/nations.ts:53-69`) already inlines exactly this shape. The comment defers extraction to #155's `winLossStackedOption` — but CLAUDE.md's rules 1 and 4 say extract now: put it in `charts/helpers.ts` and route `nations.ts` through it in this PR. Whichever of the three PRs lands first should own the helper; deferring means #155 rebases onto a duplicate it then has to delete.
- **Stale doc comment on `familyNationPicksOption`** (`charts/families.ts:76-79`): still says *"paired horizontal bars of pick rate and win rate per pool class, sorted by pick rate"* — the whole point of the change is that it's now a stacked wins/losses bar sorted by pick count.
- **Tooltip wording is wrong in the All-nations case.** `Picked in ${r.count} games (${pct(r.pickRate)} of this nation's)` — when `nation === ALL_NATIONS` the denominator is total corpus games across every nation. The pre-PR wording (`Picked: X% (N games)`) didn't have this problem.
- **Duplicate import** in `FamilyStatsPanel.svelte:15-16` — two `import … from "./charts/helpers"` statements. No lint rule catches it; merge them.
- **Height inconsistency in the same panel:** the capital chart uses `barChartHeight(...)`, the per-nation chart is still hardcoded `height="400px"` despite having a comparable per-class row count and crest labels. `barChartHeight` is already imported.
- **`BUNDLE_SCHEMA_VERSION` changelog line** says only "6: capital family class win rate" — v6 also adds three fields to `familyByNation`.
- **Empty-state divergence on the tournament tab:** the outer guard renders `text-tan opacity-60` "No completed games yet.", while `FamilyStatsPanel`'s own guard renders `text-brown` "Not enough data." — reachable together (capital data present, `familyByNation` empty). The outer guard largely duplicates the panel's own handling.

## Tests

Good coverage of the interesting semantics — the captured-city founding test is the right test to have written. Two notes on `test/helpers/save-blob.ts`:

- Every seeded city gets `owner_nation: "NATION_EGYPT"` regardless of `owner`, so a `{owner: 1}` city claims player 0's nation. Harmless for these tests (they key on xml_id) but it models an impossible save and would mislead any future `cities_total` test. Derive it from the owner index.
- `cities: UploadFixtureOpts["cities"] & object` reads oddly; `NonNullable<UploadFixtureOpts["cities"]>` says the same thing directly.

Missing: a case at the param-cap boundary (see blocking issue), and one asserting `capitalFamilyWinRate` is `[]` when no city carries a family (pre-2.6.0 blob) — the graceful-degradation path the whole backfill story rests on.

## Security / performance

No concerns. No new PII surface; `player_family_cities` holds only game-derived aggregates. Cache keys embed `BUNDLE_SCHEMA_VERSION`, so the bump invalidates the tournament bundle too — verified `handleTournamentGamesStats` goes through the same `getCached`/`putCached`. `loadFamilyCities` chunks at `CHUNK_SIZE` like its siblings and is genuinely small (≤3 rows/human/game). FK `ON DELETE CASCADE` matches `tech_events`/`law_events`, and game deletion relies on the same cascade — consistent, nothing to add there.

## Verdict

Blocking on the D1 parameter bug — it breaks uploads that work on `main` today, for a common save shape. Everything else is small: two dead indexes, the `outcomeSeries` extraction, and a handful of comment/wording fixes. Once the chunk size is fixed and covered by a test, this is a clean merge.

---

becked:
Rebased onto current `main` (`3013fdf`) and pushed — `6a85a68`. #155 and #157 both landed, so this is the PR that takes the rebase you called out.

Eight files conflicted: `derive-player-summary.ts` + its test (add/add), `games.ts`, `stats/aggregate.ts`, `stats/cache.ts`, `test/helpers/save-blob.ts`, `src/lib/stats/types.ts`, and the tournament stats page. The review findings are **not** addressed yet — this is the rebase only, so the blocking D1 parameter bug is still there.

## Resolutions that touch code `main` owns

**The `cities` upload fixture had to become one option.** `main`'s wonder-stats seed (`{owner_player_xml_id, culture_level}`) and this PR's family seed (`{owner, familyClass, foundedTurn, isCapital, capturedFrom}`) both populate `city_statistics.cities` — one array, so they can't coexist as two options. Merged into a single seed with every field but `owner` optional, and renamed the field at `main`'s five wonder-stats call sites (`owner_player_xml_id: 0` → `owner: 0`). The unified type now mixes `culture_level` (mirrors the blob field) with camelCase family fields; worth settling one way while the other fixture note is being addressed.

**`derive-player-summary.test.ts` was add/add with two different `city()` helpers.** Kept `main`'s positional builder, extended it with a `family` param, and added a small `capital()` wrapper. The three capital tests now go through `main`'s general `blobWith({ city_statistics: … })` rather than this PR's array-only variant. `PLAYER` picked up `nation: "NATION_ROME"` from this side.

**Ordering.** `capitalFamilyWinRate` sits after `wonderStats` and before `familyByNation` in `aggregate.ts`, both `types.ts` mirrors, and both return objects — which keeps the "distinct from familyByNation below" comment accurate. The tournament `families` tab goes after `wonders`, matching `CATEGORIES` in `charts/registry.ts`.

## Two things git couldn't flag

- **Migrations renumbered** `0035`/`0036` → `0037_capital_family_class.sql` / `0038_player_family_cities.sql`. `main` now occupies 0034–0036 (`tournament_match_casters`, `match_occupant_snapshot_indexes`, `wonder_events`). Because the filenames differ, git merged both sets without a conflict — the branch would otherwise have carried two `0035_*` and two `0036_*`.
- **`BUNDLE_SCHEMA_VERSION` 6 → 8**, below `main`'s v7 wonders line.

## Verification

`npm run lint` clean, `svelte-check` 0 errors, cloud `tsc` 0 errors, worker suite **503 passed / 4 failed** — the 4 are the pre-existing `canonical-map-options` DOTA drift on `main`; this branch touches neither that test nor `map-option-defs.ts`.

## One finding the rebase shrank

`winLossStackedOption` is now on `main` at `src/lib/stats/charts/helpers.ts:129`, with `nations.ts` and `leaders.ts` already routed through it. So the third-copy finding is no longer "extract into `helpers.ts` and rewire `nations.ts`" — it's delete `outcomeSeries` in `families.ts:24` and call the shared helper. Your `24d5367` comment already points at exactly that.

Everything else from the review stands as written: the D1 parameter cap and its ≥21-row test, the two dead indexes, the capital tie-break, the `avg_share` denominator comment, the stale `familyNationPicksOption` doc comment, the All-nations tooltip wording, the duplicate `helpers` import in `FamilyStatsPanel.svelte`, `barChartHeight` on the per-nation chart, the empty-state divergence, and the fixture's `owner_nation`. The v8 changelog line also still names only the capital family class, not the three new `familyByNation` fields.

---

Issue #150, opened 2026-07-20
R2 blob latency for non-US users — there is no edge cache tier today


becked:
## Context

R2 bucket `per-ankh-shares` is in `ENAM` (verified live: 1,047 objects, 242 MB). Unlike D1, **R2 cannot be replicated** — a bucket's region is fixed at creation and Cloudflare has no cross-region replication product. Caching is the only available lever for non-US latency on blob reads. (D1 replication is tracked separately in issue #149.)

## The finding

**No Worker response in this codebase is stored in Cloudflare's edge cache.**

Cloudflare Workers run *before* the cache. A response generated in a Worker via `new Response()` is never automatically cached at the edge — only `fetch()` subrequests auto-populate the CDN cache, or an explicit `caches.default.put()`. This app generates every response in the Worker, and the only `caches.default` use in the repo is a rate-limit counter on a synthetic key (`share-legacy.ts:149-158`).

Consequence: the `s-maxage` directives already in the code are **inert at the edge** — they constrain browser and intermediary caches only.

- `games.ts:2255` — `public, max-age=3600, s-maxage=60`
- `games.ts:2096` — `public, max-age=300, s-maxage=60`
- `channels.ts:325` — `public, max-age=0, s-maxage=60`

Every `SHARE_BUCKET.get()` therefore does a full ENAM round trip on every request, for every user, worldwide. This is not a TTL-tuning job; it is introducing a cache tier that does not exist.

The comment at `games.ts:2247-2252` documents `s-maxage=60` as a deliberate trade so a Make Private toggle propagates in ~1 min. That reasoning is sound but currently moot.

**Worth confirming empirically before building:** check `cf-cache-status` on a cold vs. warm request to `/v1/games/public-recent`. That hits prod, so it needs an explicit go-ahead.

## Why this shape is safe here

Because Workers run *before* the cache, a Cache API tier lives **inside** the Worker. Auth checks, rate-limit counting, and audit writes all still execute on a cache hit — the cache elides only the R2 round trip, not enforcement. That makes "authorize at origin, then serve bytes from cache" architecturally clean rather than a security hole.

## R2 read inventory

| file:line | Key | Route | Auth | Cache-Control |
|---|---|---|---|---|
| `games.ts:2195` | `games/{id}.json.gz` | `GET /v1/games/:id` | public → anon OK; private → owner | owner `private, no-store`; other `public, max-age=3600, s-maxage=60` |
[...]

Issue #149, opened 2026-07-20
D1 read replication for non-US latency — gated on a rate-limit staleness decision


becked:
## Context

Per-Ankh has a substantial non-US player base, and all of our stateful backends sit in one region. Verified against the live account:

- `SHARE_DB` (`per-ankh-share-index`) — `running_in_region: ENAM`, `read_replication: { mode: "disabled" }`, ~10.8k reads/day, ~1.56M rows read/day.
- `per-ankh-shares` (R2) — `location: ENAM`.

Every D1 query from a Worker running in, say, Sydney or Frankfurt crosses to Eastern North America and back. Handlers issuing several sequential queries pay that RTT per query, not per request.

Of our four backing services, the Worker and KV are already globally distributed, and R2 **cannot** be replicated (bucket region is fixed at creation; no cross-region replication product exists — that's issue #150). D1 is the only one where we can move data closer to users.

## Cost

**Zero.** Cloudflare bills D1 solely on `rows_read` / `rows_written` / storage, identically whether replication is on or off; replicas incur no extra storage or compute charge, and replicating writes isn't billed. Workers bill CPU time rather than wall-clock, so latency saved waiting on queries doesn't show up as a saving there either.

The constraint on this work is correctness risk, not money.

## The blocking discovery

Read replication only takes effect via the Sessions API (`env.SHARE_DB.withSession(...)`); without it all queries continue to hit primary regardless of the setting. Replica reads may be **stale**.

The obvious candidates — tournament standings, bracket, public tournament views, public game detail — turn out **not to be read-only**. The `events` table serves as both audit log and rate-limit counter, so these endpoints run:

```sql
SELECT COUNT(*) FROM events WHERE ip_address = ? AND event_type IN (...) AND created_at > datetime('now','-1 hour')
-- decide allow/deny
INSERT INTO events (event_type, ip_address) VALUES ('tournament_view', ?)
```

Disqualified this way:

- All ten public tournament routes, via one shared preamble — `enforceTournamentViewRateLimit` (`cloud/src/tournament/public.ts:85`, count at `:93`, INSERT at `:107`), reached through `loadViewableTournament` (`public.ts:732`).
- `GET /v1/games/public-recent` — count `games.ts:1888`, INSERT `games.ts:1903`.
- `GET /v1/games/:id` — count `games.ts:2164`, INSERT `games.ts:2181`, on the **non-owner branch only** (`games.ts:2160`).
[...]

PR #164, opened 2026-07-26
perf(cloud): cache parsed-game blobs per-POP in front of R2

becked:
Addresses the non-US blob-read latency in #150.

## The problem

`per-ankh-shares` is in ENAM, R2 buckets are single-region, and Cloudflare has no cross-region replication — so caching is the only lever. But there was nothing to tune: **Workers run before the edge cache**, so a response built with `new Response()` is never stored, and the `s-maxage` directives already on these routes only ever constrained browsers. Every `SHARE_BUCKET.get()` was a full ENAM round trip, for every viewer, worldwide.

## The approach

Cache the **R2 object bytes**, not the HTTP response. That choice is what keeps the invalidation surface to one path instead of the seven #150 enumerated.

Both public readers take a fresh D1 row read *before* the R2 read, and the response body is the blob with D1 metadata (`display_name`, `collection_id`, `user_nation`, …) merged in *after* it, with `stripOnlineIds` running per request. So the cache holds bytes and never a decision:

- **Visibility flips** — `is_public` PATCH, `linkTournamentMatch()` forcing public (including the dedup-hit branch that writes no R2 at all) — need no invalidation. The gate re-evaluates and 403/401s before the cache is consulted.
- **Deletes** likewise, including the admin CLI's out-of-band `wrangler r2 object delete`: it drops the games row alongside the object, so the cached bytes become unreachable.
- **Metadata edits** never stale an entry, because they're merged post-read.

That leaves the re-import overwrite, and the cache key handles it structurally.

## Key drift instead of purge

The key carries the game's `parser_version`, mirroring the `PARSER_VERSION` segment in `stats/cache.ts`. A re-import is the only path that overwrites a blob in place, it only runs on a strictly-newer version (`compareSemver > 0`), and that version lands on the games row — so the key drifts exactly when the bytes change, and **every POP misses at once** rather than only the colo that ran the upload.

This matters because Cache API entries are per-POP and don't replicate, and zone purge can't reach these keys at all (synthetic, so they belong to no zone). Key drift makes reparse propagation global, which is what lets the TTL be **24h** — and at this request volume, a long TTL is what gives a distant colo any chance of a second reader landing on a warm entry.

One explicit eviction survives, keyed on the *outgoing* version: if the D1 batch rolls back after the R2 put, `handleGameUpload` deliberately keeps the new bytes with the old row and intends the old row to render them. Without it a warm entry would shadow those bytes for the full TTL.

## Deliberate exclusions

- **Owners bypass the cache.** Their response is `private, no-store` specifically so a reload right after a reparse shows the new bytes. Side benefit: cached bytes are only ever read on the path that strips `online_id`.
- **`saves/{id}.zip` is not cached.** It streams straight from R2 today; caching means buffering or teeing, which costs that memory-safety guarantee at the 50 MB ceiling — for an explicit, rate-limited user action rather than the repeat traffic the cache exists for. Recorded at the read site.

## Not covered by this PR

#150's other opportunities are untouched, so this **refs** rather than closes it: the ZIP path (#1), `GET /v1/games/public-recent` (#3), and the legacy `/v1/share/:id` viewer (#4).

Also worth flagging: there's still no runtime instrumentation separating the R2 hop from the D1 queries that run serially ahead of it (D1 is ENAM-primary too — #149). `setLogField()` plus the `cf_ray` colo suffix already in the access log would answer that, and would show real hit/miss rates. Not in scope here.

## Testing

7 new integration tests in `cloud/test/integration/games/blob-cache.test.ts`. Since the tier is invisible from the response body, each proves itself by deleting the R2 object while leaving the games row intact — a cached read still 200s, an owner read 404s. Three pin the load-bearing design claim directly: an `is_public` flip 401s, a deleted game 404s, and a restored blob is visible immediately, all with **zero** cache invalidation anywhere. One pins key drift: bumping `parser_version` on the row orphans the warm entry.

Full `cloud` suite 337/337 across 45 files, `tsc --noEmit` clean, lint and prettier clean.

### DIFF
--- a/.claude/skills/admin-cli/SKILL.md
+++ b/.claude/skills/admin-cli/SKILL.md
@@ -54,7 +54,9 @@ Build a full local fixture (Swiss + championship via the real planner) with `./p
 ./per-ankh admin cache clear videos                                # Force a YouTube refetch
 ```
 
-Unlike the Worker's `invalidateStatsCache` (which walks only the current `BUNDLE_SCHEMA_VERSION`), `clear` sweeps every schema version under the prefix — that's what reaches entries orphaned by a version bump. Clearing `videos` costs YouTube Data API quota on the next playlist view. Edge-cached responses (`Cache-Control: s-maxage`) are a separate layer and are **not** touched.
+Unlike the Worker's `invalidateStatsCache` (which walks only the current `BUNDLE_SCHEMA_VERSION`), `clear` sweeps every schema version under the prefix — that's what reaches entries orphaned by a version bump. Clearing `videos` costs YouTube Data API quota on the next playlist view.
+
+**The parsed-game blob cache is a different layer and this command cannot reach it.** It caches R2 bytes for `games/{id}.json.gz` in each POP's Cache API storage (`cloud/src/blob-cache.ts`), which has no CLI surface — `wrangler` has no cache command, and zone purge can't address the keys because they're synthetic and belong to no zone. It also needs no operator action: the key carries the game's `parser_version`, so a reparse drifts the key and every POP misses at once. If a game somehow serves stale bytes anyway, that's the D1-batch-rollback path in `handleGameUpload` (the Worker evicts locally there) and the entry expires on its own within 24h. The `Cache-Control: s-maxage` directives on game responses are inert at the edge — Workers run before the cache, so nothing stores them (issue #150).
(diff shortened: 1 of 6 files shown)


becked:
@claude review, be particularly careful on caching

PR #168, opened 2026-07-27
refactor(game-detail): make BuildComparison generic over its subject

alcaras:
`BuildComparison` is the app's only diverging-bar module, and it was hard-wired to units: the `SpriteIcon` category, the `UNIT_` label prefix, and `unitType` on every item. The same head-to-head shape wants reusing for anything counted per player.

It now takes `iconCategory` and `labelOf`, both defaulting to the unit behaviour, so the Military tab's call sites are unchanged apart from `unitTypes` → `keys`. `BuildItem.unitType` becomes the neutral `key`. There's also a new optional `showDiff` column — the ±N gap between the two sides on each row and on the total — which is off by default and unused here.

**No user-visible change.** The Military tab renders exactly as before; this is groundwork, sent on its own so the regression surface is reviewable in isolation.

It's the first of four — Economy tab, Families tab and a Specialists rework follow, and each uses this. I'll send them as this lands rather than all at once.

### DIFF
--- a/src/lib/game-detail/BuildComparison.svelte
+++ b/src/lib/game-detail/BuildComparison.svelte
@@ -1,10 +1,3 @@
-<script lang="ts" module>
-	import type { SpriteCategory } from "./helpers";
-
-	/** One row's subject and how many of it a side has. */
-	export type BuildItem = { key: string; count: number };
-</script>
-
 <script lang="ts">
 	// Head-to-head-by-type module, ported from owglick's H2H card. Every type is
 	// a center-split diverging-bar row (player A grows left, B grows right),
@@ -15,6 +8,7 @@
 	// caller gates it to two players.
 	import SpriteIcon from "./SpriteIcon.svelte";
 	import { formatEnum } from "$lib/utils/formatting";
+	import type { BuildItem, SpriteCategory } from "./helpers";
 
 	let {
 		title,
@@ -60,10 +54,10 @@
 		ca: number;
 		cb: number;
 	};
-	// One row per unit type, listed alphabetically by display name. A side that
-	// never built a type carries a 0 and renders a blank bar/count on its half.
-	// Rows come from the caller's shared `keys` order when given (so panels
-	// line up); otherwise from this panel's own union of both rosters.
+	// One row per key. A side that has none of a key carries a 0 and renders a
+	// blank bar/count on its half. Rows come from the caller's shared `keys`
+	// order when given (so panels line up); otherwise from this panel's own
+	// union of both sides, alphabetically by display name.
 	const rows = $derived<Row[]>(
 		(
 			keys ??
@@ -78,7 +72,7 @@
 	);
 	// Bar scale: longest single-side count across all rows.
 	const max = $derived(Math.max(1, ...rows.map((r) => Math.max(r.ca, r.cb))));
-	// Per-side unit totals, shown in a footer row aligned under the count columns.
+	// Per-side totals, shown in a footer row aligned under the count columns.
 	const totalA = $derived([...aM.values()].reduce((t, n) => t + n, 0));
 	const totalB = $derived([...bM.values()].reduce((t, n) => t + n, 0));
 </script>
--- a/src/lib/game-detail/BuildComparison.svelte
+++ b/src/lib/game-detail/BuildComparison.svelte
@@ -27,7 +27,6 @@
 		keys,
 		iconCategory = "units",
 		labelOf = (key: string) => formatEnum(key, "UNIT_"),
-		showDiff = false,
 	}: {
 		title: string;
 		statA?: string;
@@ -45,8 +44,6 @@
 		iconCategory?: SpriteCategory;
 		// eslint-disable-next-line no-unused-vars -- callback type signature
 		labelOf?: (key: string) => string;
-		// Trailing ±N column: how far ahead the leading side is on that row.
-		showDiff?: boolean;
 	} = $props();
 
 	function byType(items: BuildItem[]): Map<string, number> {
@@ -144,19 +141,6 @@
 							class="w-5 flex-none text-center font-mono text-[11px] text-white"
 							>{r.cb || ""}</span
 						>
-						{#if showDiff}
-							<span
-								class="w-8 flex-none text-right font-mono text-[10px]"
-								style="color:{r.ca === r.cb
-									? 'inherit'
-									: r.ca > r.cb
-										? ca
-										: cb}"
-								class:text-muted={r.ca === r.cb}
-							>
-								{r.ca === r.cb ? "" : `+${Math.abs(r.ca - r.cb)}`}
-							</span>
-						{/if}
 					</div>
 				</div>
 			{/each}
@@ -178,19 +162,6 @@
 						class="w-5 flex-none text-center font-mono text-[10px]"
 						style="color:{cb}">{totalB}</span
 					>
-					{#if showDiff}
-						<span
-							class="w-8 flex-none text-right font-mono text-[10px] font-bold"
-							style="color:{totalA === totalB
-								? 'inherit'
-								: totalA > totalB
-									? ca
-									: cb}"
-							class:text-muted={totalA === totalB}
-						>
-							{totalA === totalB ? "" : `+${Math.abs(totalA - totalB)}`}
-						</span>
-					{/if}
 				</div>
 			</div>
 		</div>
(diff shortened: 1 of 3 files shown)


alcaras:
Self-review pass.

**Verified:** every `BuildItem` call site in MilitaryTab is renamed — the `unitType` identifiers still in that file belong to the event-rail's own marker types (tech-unlock/bonus-card tooltips), not to this component. Defaults (`iconCategory="units"`, `labelOf` = `formatEnum(key, "UNIT_")`) preserve the old rendering exactly; the only call-site change is `unitTypes` → `keys`. `npm run lint` + `svelte-check` clean at this commit alone.

**Disclosed:** `showDiff` has no consumer inside this PR — #169 is the consumer. Kept here so the shared component is touched once in the stack rather than in two PRs; if you'd rather it ride with its first caller I'll move it.

One stale comment (still said `unitTypes`) fixed in 2e01d14.

becked:
## To move this forward

**1. Widen the scope: extract the aggregation, not just the component.**

This generalizes the rendering half of the diverging-bar module but leaves the feeding half to each caller. Downstream, three PRs each write the same thing independently:

- `EconomyTab.svelte:477` — `panels()`
- `SpecialistsTab.svelte:278` — `comparisonPanels`
- `EconomyTab.svelte:574` (#172) — `projectPanel`

All three are the same shape: build a per-side `Map<group, Map<key, count>>`, union the keys across sides, sort, emit `BuildItem[][]`. That belongs here, alongside the component it feeds — otherwise the reuse this PR is for only lands half the time. `panels()` in #169 is already the general form (it takes `bucket`/`keyOf`/`sortKey`/`weightOf`); lift that into a helper here and let #169/#171/#172 call it.

**2. `showDiff` has no consumer in this PR.**

`BuildComparison.svelte:147-157` and `:181-193` add the ±N column, and nothing in this diff turns it on — it's exercised first in #169. As sent, this is speculative API (`CLAUDE.md` rule 5). Either fold in the call site or drop it and add it with #169.

---

becked:
## Pushed two commits to this branch

Rather than send you back and forth, I've implemented both review points on `pr1-build-comparison` (`3b7c416`, `eb3aaa0`). Revert or rework freely — but read the second one first, because I did **not** do what my review asked for.

### `3b7c416` — drop `showDiff` until its first caller

Straightforward: it rides with #169 now. Worth noting on the way out that its equal-case styling was unreachable — the `'inherit'` colour and the `text-muted` class only apply when `r.ca === r.cb`, which is exactly when the cell renders `""`. When you re-add it with the call site, that ternary can collapse to the two-branch form.

### `eb3aaa0` — `comparisonRowKeys()` in `helpers.ts`, not a lifted `panels()`

My review said to lift #169's `panels()` up here. **That was wrong and I'm not asking for it.** `panels()` closes over `improvementData.improvements` and `orderedPlayers`; making it also serve #172 (`projectsProduced`, keyed `player_xml_id`, ungrouped) and this tab (two different owner rules across its two blocks) needs six callbacks and a generic row type. Worse than the duplication it removes.

Two things narrowed the target:

1. **The zero-fill in the callers is redundant.** `BuildComparison` already re-aggregates (`byType`) and fills gaps via `aM.get(t) ?? 0`. So `items: sides.map(s => keys.map(key => ({ key, count: s.get(...) ?? 0 })))` is work the component repeats. Callers only ever needed to hand over raw per-side rows plus a shared key order.
2. **The bucketing loops genuinely differ** — different source tables, different owner rules, different group axes. They read fine locally. It's the tail that's common.

So the extracted piece is just the tail:

```ts
export function comparisonRowKeys(
  sides: BuildItem[][],
  compare?: (a: string, b: string) => number,  // default: total desc, then key
): string[]
```

`MilitaryTab`'s `sharedBuildUnitTypes` is now its first caller, so nothing in this PR is speculative any more. Row order is unchanged — same traversal order into the map as into the `Set` it replaces, same comparator, stable sort.

`BuildItem` moved to `helpers.ts` alongside `SpriteCategory` and `CityColumn`. A `.ts` helper importing a type out of a `.svelte` module block is backwards, and that block existed only to export it. `MilitaryTab` imports it from `./helpers` now.

### The drift this was actually about

In #172's tree, `EconomyTab.svelte:524` sorts ties by insertion order and `:588` sorts them alphabetically — same tab, two panels, two orderings of the same "combined total, biggest first" idea. The default in `comparisonRowKeys` breaks ties by key, so there's one implementation of that order.

Also worth naming, since it's the sharper argument for extracting at all: in #171's tree, `EconomyTab.svelte:477` has `panels()` **and** `SpecialistsTab.svelte:278` has a hand-rolled duplicate of it. Not two branches that couldn't see each other — one tree. `panels()` just wasn't reachable, being a component-local function.

### What's left for you

Downstream rebases (all five are based on `99429f6`, not this tip, so they were rebasing regardless):

- **#169** (×2) and **#172** (×3) pass `showDiff` — those call sites need it re-added with its first consumer.
- `panels()` (`EconomyTab.svelte:477`), `comparisonPanels` (`SpecialistsTab.svelte:278`) and `projectPanel` (`EconomyTab.svelte:574`) can each drop their union/sort/zero-fill tail onto `comparisonRowKeys` and pass raw per-side rows through.

One heads-up unrelated to this PR: `web/src/lib/game-detail` is a symlink to `src/lib/game-detail`, and the legacy share viewer renders `GameDetailView`. Everything in this stack lands there too on its next redeploy, against blobs parsed by older `PARSER_VERSION`s. This PR is behaviour-preserving so it's a non-event here, but #172's `projects_produced` at 2.13.0 is worth a thought for old blobs.

`npm run lint`, `prettier --check` and `npm run check` (0 errors, 0 warnings, 1283 files) all clean at `eb3aaa0`. I did not render the Military tab against a real save — the equivalence argument above is static.

---

becked:
## Correction, and one more commit

### Correcting my `eb3aaa0` note

I wrote:

> A `.ts` helper importing a type out of a `.svelte` module block is backwards, and that block existed only to export it.

There was no module block. `BuildComparison.svelte` has exactly one `<script lang="ts">` on `main` — line 1, closing at line 74 — and `export type BuildItem` was declared inside it at line 11. So I described a construct your file never had, and implied the move deleted a block when in fact one type declaration moved and nothing else changed.

The move is still right, for a slightly different reason than the one I gave: an instance `<script>` runs per component instance and isn't a module surface at all, so exporting a type from it is the odder thing. `helpers.ts` is where it belongs. But the reasoning as written was wrong and I'd rather say so than leave it in the thread.

### `005a3a3` — require `comparisonRowKeys`' comparator

The `compare` parameter was optional, defaulting to combined-total descending then key. Nothing in this PR takes that default: `MilitaryTab` passes its own display-name comparator. That's precisely the speculative-API objection that sent `showDiff` away in `3b7c416`, and I missed that it applied to my own commit one file over.

So `compare` is required now, and the totals map went with it — with a comparator always supplied, only `totals.keys()` was ever read, so summing counts was work no caller consumed:

```ts
export function comparisonRowKeys(
	sides: BuildItem[][],
	compare: (a: string, b: string) => number,
): string[] {
	const keys = new Set<string>();
	for (const side of sides) {
		for (const it of side) keys.add(it.key);
	}
	return [...keys].sort(compare);
}
```

Row order in this PR is unchanged — same keys, same comparator, same stable sort.

### What this changes for the stack

Smaller than what I described last time. `comparisonRowKeys` now covers the union and the sort; the zero-fill is still redundant (`BuildComparison` re-aggregates and fills with `?? 0`, and the footer totals sum the raw side maps, so padding wouldn't move them either). But a panel that wants total-descending order builds its own totals and hands over the comparator:

```ts
const total = (key: string) =>
	sides.reduce((sum, s) => sum + (s.get(group)?.get(key) ?? 0), 0);
const keys = comparisonRowKeys(rows, sortKey
	? (a, b) => sortKey(a) - sortKey(b)
	: (a, b) => total(b) - total(a) || a.localeCompare(b));
```

Which means the tie-order drift I pointed at — `EconomyTab.svelte:524` breaking ties by insertion order, `:588` breaking them alphabetically — isn't fixed by this PR any more. It gets fixed in #169, and if #172's `projectPanel` wants the same order, that comparator is the thing to lift into `helpers.ts` at that point, with two real consumers behind it. #171's `comparisonPanels` passes `specialistSortKey`, so it's unaffected either way.

Downstream call sites still need `showDiff` re-added with its first consumer, as before.

`npm run lint`, `prettier --check` and `npm run check` (0 errors, 0 warnings, 1283 files) clean at `005a3a3`. Still a static equivalence argument — I haven't rendered the Military tab against a real save.

---

PR #169, opened 2026-07-27
feat(game-detail): Economy tab

alcaras:
Replaces the Improvements tab with an **Economy** tab: one switchable economy curve with an annotation rail, a worker-turn ledger, and the improvements pivot table the old tab already had.

> **Stacked on #168.** Its commit leads this diff and disappears once #168 lands.

## The curve

The switch — GDP · Maintenance · Territory · Cities founded · Workers built — picks the curve and nothing else, following the Techs tab, whose switch picks `chartOption` while everything below renders regardless. Under the plot sits the Military tab's DOM event rail, carrying the economy milestones the save actually dates: **city foundings, wonder completions and calamities**. There is no per-turn improvement history anywhere in the save, so those three are what's datable.

**GDP** is money income plus food/wood/stone/iron income valued at that turn's market price. The prices are the game's own: `yield_price_history` has been in the blob all along with no consumer. Forward-filled per turn and divided by the save's ×10,000 fixed point, which `yield.xml`'s `<iPrice>4</iPrice>` pins against a turn-1 raw value of ~40,200. The tooltip breaks each side's total into its five parts with the unit price, since a food-heavy economy and a stone-heavy one at the same GDP played very differently. Across 394 cached duels the player with the higher cumulative GDP won **68%** — real signal, and unlike ending military power (90%) not close to tautological.

**Maintenance** duplicates a chart the Yields tab already has. Kept deliberately: the rail annotates it, which is the point of having it here. Flagging it rather than hiding it.

**Territory** comes from `tile_ownership_history` — also previously unused, and it tracks losses as well as gains, unlike the other two curves.

## The ledger

A worker on a tile adds one turn of progress per turn, so an improvement's build time *is* what it cost the workforce. New bake reads `iBuildTurns` from `improvement.xml` plus its event adds.

Improvements that **mature** inherit their source's cost — a Hamlet (4 turns) becomes a Village after 20 and a Town after 20 more, and neither of those is `<bBuild>`, so they'd otherwise price at zero. That's 1,332 tiles and 5,328 worker-turns across the cached corpus, ~1.7% of all work done.

I checked the base costs against the save's own `<ImprovementBuildTurnsOriginal>` over 7,671 improved tiles: exact match 62% of the time, within ±1 for 96%. The gaps are the game's per-city modifiers (Builder −1, Steward +1, Artisans seat −2 urban). Reading that per-tile value is the natural follow-up once the parser carries it.

## Smaller things

- **National wealth** is a panel, not a curve: the save records `YieldStockpile` only for the final turn, so wealth over time isn't recoverable.
- `#improvements` has been a shareable link for as long as that tab existed, so old hashes map forward rather than rendering an empty pane.
- The side-by-side comparison is 1v1 like the Military tab's, and says so in FFA instead of silently vanishing.
- `EventRail` now exports the surface/border colours it already used, so chart tooltips can sit on the same background instead of ECharts' white default.

Two of the four; Families tab and a Specialists rework follow.


alcaras:
Self-review pass — things worth knowing beyond the description:

**Verified:**
- `#improvements` has been a shareable deep link since the tab existed; the hash migration maps it to `#economy` (checked it reads the hash unconditionally, so the effect tracks).
- Zero references to the deleted `ImprovementsTab` anywhere, including the frozen `web/` tree that symlinks `game-detail/` wholesale.
- The pivot table keeps the `tables.improvements` state key, so search/sort state behaves exactly as the old tab.
- `src/lib/generated/improvement-builds.ts` is pure bake output from the script in this PR — no hand edits.

**Deliberate trade-offs, flagged rather than hidden:**
- The Maintenance view duplicates a chart the Yields tab already has (`YIELD_CHART_CONFIG` includes `YIELD_MAINTENANCE`). Kept because the annotation rail under it is the point; happy to drop the mode if you'd rather not carry the duplicate.
- `STOCKPILE_SCALE = 10` (national-wealth panel) follows the save's universal ×10 fixed point for yield quantities, but unlike `PRICE_SCALE` I couldn't pin it against an independent bound. If an in-game stockpile number ever disagrees by ×10, that constant is the suspect — it's commented as such.
- Side-by-side comparisons are 1v1 (Military-tab framing) and say so in FFA instead of silently vanishing.

Sanity check on GDP: across 394 cached duels, the player with the higher cumulative GDP won 68% — signal, not tautology (ending military power is 90%).

becked:
## To move this forward

**1. `National wealth` renders a bare heading on legacy blobs.**

`EconomyTab.svelte:629` — the `<h2>` sits outside its `{#if wealth.some(...)}`:

```svelte
<section class="mb-6">
  <h2 …>National wealth</h2>
  {#if wealth.some((w) => w.components.length > 0)}
```

With no `player_resources` (older blobs, and the frozen `web/` viewer, which is what the `= []` default is for) the section renders as a heading with nothing under it. Every other conditional block on this tab hides heading and body together — move the `<h2>` inside the guard.

**2. `panels()` should be the shared extraction, not a local.**

`EconomyTab.svelte:477` is the general form of an aggregation #171 and #172 each rewrite. Lift it next to `BuildComparison` in #168 and call it from here. See the note on that PR.

**3. Stale comments — the switch grew to five modes.**

- `EconomyTab.svelte:125` — "Four views of the same question"; `EMPIRE_MODES` has five (`gdp`, `maintenance`, `territory`, `cities`, `workers`).
- `EconomyTab.svelte:566` — "three growth curves the blob can answer".
- `EconomyTab.svelte:373` — "Three rows per nation: cities founded, wonders completed, calamities. **Both** are worker-turn events" — three rows, and wonders/calamities aren't worker-turn events.


---

becked:
## Retracting review point 2, and what #168 leaves you

### Point 2 is withdrawn

I said:

> `panels()` should be the shared extraction, not a local. `EconomyTab.svelte:477` is the general form of an aggregation #171 and #172 each rewrite. Lift it next to `BuildComparison` in #168 and call it from here.

Don't do that. Making `panels()` also serve #172 (`projectsProduced`, keyed `player_xml_id`, ungrouped) and #171 needs six callbacks and a generic row type — worse than the duplication it removes. The bucketing loops genuinely differ: different source tables, different owner rules, different group axes. Only the tail is common.

Full reasoning is on #168; the extracted piece ended up being just that tail:

```ts
export function comparisonRowKeys(
	sides: BuildItem[][],
	compare: (a: string, b: string) => number,
): string[]
```

Points **1** (National wealth `<h2>` outside its guard, `EconomyTab.svelte:629`) and **3** (three stale comments at `:125`, `:566`, `:373`) still stand as written.

### The rebase will apply cleanly and still not compile

Worth knowing before you run it. `pr2-economy-tab` replays two commits over #168's tip, and they touch files #168's later commits didn't — so no conflict markers, and three breaks that only `npm run check` will surface:

1. **`EconomyTab.svelte:37`** — `import BuildComparison, { type BuildItem } from "./BuildComparison.svelte"`. That named export is gone; `BuildItem` lives in `./helpers` now, which this file already imports from twenty lines down.

2. **`showDiff` at `:754` and `:775`** — the prop was removed in `3b7c416` since #168 had no consumer for it. This PR is the first consumer, so it re-adds it. One thing to fix on the way back in: the equal-case styling was unreachable. `'inherit'` and `class:text-muted` only applied when `r.ca === r.cb`, which is exactly when the cell rendered `""`, so that ternary collapses to the two-branch form. Same on the totals row.

3. **`panels()` at `:477`** — `comparisonRowKeys` takes over the union and the sort. Its comparator is required (an optional total-descending default had no caller in #168, which was the same speculative-API objection that sent `showDiff` away), so `panels()` passes its own:

```ts
const keys = comparisonRowKeys(rows, sortKey
	? (a, b) => sortKey(a) - sortKey(b)
	: (a, b) => total(b) - total(a) || a.localeCompare(b));
```

The zero-fill in `items` can go with it — `BuildComparison` re-aggregates via `byType` and gap-fills with `?? 0`, and the footer totals sum the raw side maps, so padding each side to the full key set was never doing anything. Hand over raw per-side rows.

### The tie-break is now yours to decide

This is the drift the extraction was actually for. `panels()` currently sorts `total(b) - total(a)` with no tiebreak, so equal rows fall back on whichever side mentioned them first; #172's `projectPanel` sorts the same idea as `total(b) - total(a) || a.localeCompare(b)`. Same tab, two orderings.

Adding the `localeCompare` tiebreak here settles it, and does visibly reorder ties — fine, that's the point, but it's a behaviour change inside this PR rather than a pure adoption. If #172 wants the same order, that comparator is the thing to lift into `helpers.ts` at that point, with two real consumers behind it rather than none.

### One question, unrelated to the above

`player_resources` and `yield_price_history` reach the blob through the trailing `looseObject` on the Worker's `FullGameDataSchema`, so your "in the blob all along with no consumer" reading is right and nothing server-side needs to change. But neither carries a `maxLength`, where the sibling `tile_ownership_history` does (`cloud/src/schemas/game.ts:302`). The schema's comment scopes bounds to fields "searched/aggregated server-side" and these stay render-only, so the omission is consistent — but `yield_price_history` is per-turn and is about to have a consumer. Fine to leave; flagging it so it's a decision rather than an oversight.

---

becked:
## Rebased and pushed, plus one new finding to pick up separately

Branch is rebased onto current `main` and the four follow-ups are pushed. `#168`'s commit has dropped out of the diff as expected. `npm run check` is clean at every commit in the series, not just the tip; `npm run lint` and Prettier pass.

What landed:

- **`1354b03`** — `showDiff` restored on `BuildComparison`, in the corrected two-branch form. The `'inherit'` colour and `class:text-muted` only applied when `r.ca === r.cb`, which is exactly when the cell renders `""`, so the ternary collapses. Same on the totals row. `BuildItem`'s move to the `./helpers` import rides in this commit rather than the next, so the tree never sits uncompilable between the two.
- **`faec997`** — `comparisonRowKeys` takes over the union and the sort in `panels()`; the zero-fill is gone. Tie-break decided, see below.
- **`b8d3fed`** — the `National wealth` `<h2>` moved inside its guard.
- **`db243e8`** — the three stale comments.

### On the tie-break

Adopted, and on the **displayed** name rather than the raw key: `improvementDisplayName(a).localeCompare(improvementDisplayName(b))`. Two reasons over the raw-key form. `improvementDisplayName` resolves through `IMPROVEMENT_NAMES` and appends a shrine domain, so raw-key order and on-screen order genuinely diverge — a raw-key tie-break would look arbitrary in the rendered column. And display-name ordering is what both existing comparators already do: `MilitaryTab`'s shared union sorts `formatEnum(p, "UNIT_")`, and `BuildComparison`'s own default union sorts `labelOf(p)`.

It applies to both comparator paths, not just the total-descending one — `IMPROVEMENT_UNLOCK_COST` is a coarse integer with many ties, so the counts panel had the same insertion-order dependence. Ties visibly reorder; that's the intended effect. If #172 wants the same order, the comparator is the thing to lift into `helpers.ts` then, with two real consumers behind it.

### New: the view switch can strand the user

Not touched here — flagging it for a separate pass, since the fix is a judgement call rather than fallout from #168.

`availableModes` (`EconomyTab.svelte:160`) gates `territory` on `tileOwnershipHistory.length` and `gdp` on `yieldPrices.length`, but `workers` stays selectable no matter what `units` holds. Meanwhile `empireChartOption` drops any all-zero series (`:262`) and returns `null` once none survive (`:272`) — and `{#if empireChartOption}` at `:571` wraps **the switch as well as the plot**.

So selecting "Workers built" with no unit data unmounts the whole block, switch included. The `$effect` at `:168` won't recover it: it only re-homes `empireMode` when the mode is missing from `availableModes`, and `workers` is still there. No way back without a reload.

That is the standing state of the frozen `web/` viewer, which never passes `units` — `GameDetailView.svelte:73` defaults it to `[]`, and `web/src/routes/share/[id]/+page.svelte` doesn't pass it. Any blob predating unit data does the same on the main app.

Gating `workers` on `units.length > 0` alongside the other two is the change that matches the existing idiom; lifting the switch out of the `empireChartOption` guard is the other option, and arguably the more robust one, since it also covers a mode going empty for reasons nobody predicted.

PR #170, opened 2026-07-27
feat(game-detail): Families tab

alcaras:
Families get their own tab. They were only visible inside Economy, which had grown to five sections and buried them — and they're a first-class Old World subsystem with a direct economic price.

> **Stacked on #168 and #169.** Their commits lead this diff and disappear as they land.

## What the XML says

A family's opinion of its player falls in one of six bands, and the band applies an `EffectCity` to **every city that family holds**. Five of the six move upkeep:

| Band | Opinion | Upkeep | Also |
|---|---|---|---|
| Furious | ≤ −200 | **+50%** | −20% civics & training, 10% rebel chance, no hurrying |
| Angry | −199…−100 | **+20%** | −10% civics & training, 5% rebel |
| Upset | −99…−1 | **+10%** | |
| Cautious | 0…99 | — | |
| Pleased | 100…199 | **−10%** | |
| Friendly | ≥ 200 | **−20%** | |

New bake reads the bands and their modifiers out of `opinionFamily.xml` + `effectCity.xml`, so the thresholds and percentages aren't hardcoded.

## The tab

**Family opinion over time** — one chart per nation (six lines in two colours was unreadable), families coloured from the shared series palette, nation named in the title. Uses `family_opinion_history`, which was in the blob with no consumer.

**Turns by family opinion** — each nation's turns stacked across the bands on a ruinous-to-ideal scale, then each of its families underneath in the order they entered play, with the turn-weighted upkeep modifier labelled off the end. The per-family rows are where it earns itself: in one reference game both nations average +0.1%, and underneath that each had one family at +6.5% and another at −4.5%. Bars are legitimately different lengths, since a family only exists from the turn its first city is founded — the tooltip gives that turn so a short bar isn't misread as a lost family.

**Worker-turns by family** — the economic footprint behind the politics, moved out of Economy's Workforce section.

## What I did not build

A per-family split of the maintenance bill. Upkeep is a single player-level yield, and `yield.xml`'s `<iPerImprovement>` of 2 money reproduces barely a fifth of it — the ratio runs 2.6–6.8 per improvement across the corpus, so distance-from-capital and the rest dominate. Splitting that between families would have been mostly invention.

Three of four; a Specialists rework follows.


alcaras:
Self-review pass.

**Chart design decisions, so they don't read as accidents:**
- Opinion is one chart **per nation** (revised after the first cut — six lines in two nation colours was unreadable). Families take `getChartColor(i)` from the shared palette; the nation is named in the title.
- `BAND_COLORS` are deliberate literals, not series colours: the six bands are a semantic scale that means the same thing in every game (same reasoning as the Specialists tab's tier ramp), and nation colour is spent on the axis labels.
- Per-family bars are legitimately different lengths — a family only exists from the turn its first city is founded, and the tooltip names that turn so a short bar isn't misread as a lost family.

**Verified:** the bake fails loudly if `opinionFamily.xml`'s band structure changes (asserts exactly one open-ended band); `EconomyTab.svelte` is byte-identical to #169's head, so nothing leaked across the split; `FAMILY_DASH` and the combined chart are fully gone.

**What I deliberately did not build:** a per-family split of the money bill. Upkeep is one player-level yield, and `<iPerImprovement>` reproduces only ~1/5 of the observed charge (ratio 2.6–6.8 per improvement across the cached corpus) — a split would be mostly invention. The band modifiers are the honest per-family statement the data supports.

becked:
## To move this forward

**1. Two different bar-height formulas for the same job.**

`FamiliesTab.svelte` defines `barHeight(rows) => rows * 26 + 110` and uses it for the worker-turns chart, but the band chart inlines `bandTallies.length * 30 + 130` at its `ChartContainer`. Pick one — if the band rows genuinely need more room, make that a parameter rather than a second formula that will drift.

**2. `BAND_COLORS` should live with the other semantic scale.**

The inline hex table is the right call for a scale that means the same thing in every game — but the repo's precedent for exactly that is `UNIT_CLASS_COLORS` in `helpers.ts`, not a component-local. Move it there so the next consumer of the bands finds it.


---

Issue #115, opened 2026-07-07
DOTA map-option schema change: sync cloud canonical mirror & unblock adding DOTA maps to live tournaments


becked:
## Summary

Old World reworked the **DOTA** map-script options. `Reference/XML` was refreshed on 2026-07-01, and `b39f690` re-baked the generated tables (deployed 2026-07-07), so the SvelteKit generated modules now carry the new DOTA schema. The **cloud-side canonical mirror was not updated to match**, which:

1. leaves `cloud/src/tournament/canonical-map-options.test.ts` red (4 failing assertions), and
2. means admins **cannot add a DOTA map** to a tournament whose pool already contains DOTA maps on the old schema — the deployed maps UI emits new-schema DOTA options that the worker's old-schema validation rejects (`400 MAP_OPTIONS_INVALID`).

This is a genuine game data change (confirmed from `Reference/XML/Infos/mapOptionsMulti.xml` — the old keys exist nowhere in current game data), **not** a bad bake, so reverting the generated data is not the answer.

## The schema change

| | Old (stored in live pools) | New (current game / generated) |
|---|---|---|
| keys | `MAP_OPTIONS_MULTI_DOTA_BOUNDARY_TERRAIN`, `MAP_OPTIONS_MULTI_DOTA_RIVER_WIDTH` | `MAP_OPTIONS_MULTI_DOTA_INNER_TERRAIN`, `MAP_OPTIONS_MULTI_DOTA_OUTER_TERRAIN`, `MAP_OPTIONS_MULTI_DOTA_PATH_WIDTH` |
| values | `MAP_OPTION_TERRAIN_JUNGLE/SAND/WATER…`, `MAP_OPTION_RIVER_NARROW` | `MAP_OPTION_TERRAIN_INNER_*`, `MAP_OPTION_TERRAIN_OUTER_*`, `MAP_OPTION_PATH_WIDE/NARROW` |

Note it's a restructure, not a rename: 2 options → 3, with a different value set.

## Current impact

- **Running tournaments are unaffected.** Matches, Swiss pairing/progression, standings, and display of already-assigned maps never touch the canonical mirror.
- **Adding non-DOTA maps still works** (only DOTA's schema changed).
[...]

PR #171, opened 2026-07-27
feat(game-detail): specialist coverage split by slot kind, and a side-by-side view

alcaras:
Two changes to the Specialists tab, both consequences of the Economy tab landing next door.

> **Stacked on #168, #169 and #170.** Their commits lead this diff and disappear as they land.

## Coverage was measuring two different things at once

`coverage` was the staffed share of *every* specialist-eligible improvement, which mixes two economies that behave nothing alike: rural specialists sit one-per-tile on farms and mines, urban ones fill graded buildings you have to build first. The denominator now splits on the eligible improvement's own kind — `ELIGIBLE_IMPROVEMENTS` already carried the `urban` flag — so the headline reads **rural filled a/b** and **urban filled a/b**, and the chart plots both.

The combined number was hiding the interesting half. One reference game:

| | overall | rural | urban |
|---|---|---|---|
| Kush | 27% | 25% | 37% |
| Persia | 46% | 41% | 68% |

Persia staffed two-thirds of their buildings against Kush's third — a completely different specialist economy than "27% vs 46%" suggests.

## The side-by-side view belongs here

`BuildComparison` (generalised in #168) now renders specialists as rural/urban diverging bars on this tab. Sorted the way the Techs tab sorts specialists — unlock cost, class contiguous, then tier — so a reader moving between the two tabs sees the same order. That rule was inline in `science-techs.ts`; it's now an exported `specialistSortKey` shared by both rather than a second copy.

1v1 like the Military tab's comparisons, and it says so in FFA instead of silently rendering nothing.

## Also

`SpecialistSummary.coverage` is removed rather than left computed-for-nobody, and the coverage chart's rural/urban colours now derive from the `SEGMENTS` table above it instead of repeating the same two hexes as literals.

Last of the four.

### DIFF
--- a/scripts/bake-specialists.ts
+++ b/scripts/bake-specialists.ts
@@ -17,9 +17,15 @@
 //                                          (EFFECTCITY_SPECIALIST_RURAL for the
 //                                          8 rural lines; the urban tiers carry
 //                                          a _1/_2/_3 suffix on the zType).
-//   Reference/XML/Infos/improvement.xml  — <Entry> with <zType> (IMPROVEMENT_*),
+//   Reference/XML/Infos/improvement*.xml — <Entry> with <zType> (IMPROVEMENT_*),
 //                                          a <Specialist> tag iff the improvement
 //                                          can hold a specialist, and <bUrban>1.
+//                                          improvement-event*.xml carry the
+//                                          event/DLC entries (LAURION_MINE takes
+//                                          a Miner) and must be read too: an
+//                                          eligible improvement missing here is
+//                                          missing from both halves of the
+//                                          coverage ratio, not just one.
 //   Reference/XML/Infos/text-*.xml       — <Entry> with <zType>TEXT_*</> and
 //                                          <en-US>. text-specialistClass.xml has
 //                                          the class line names ("Priest"); the
@@ -50,6 +56,10 @@ const REPO_ROOT = resolve(__dirname, "..");
 const SIDECAR = resolve(REPO_ROOT, ".bake/specialists.json");
 
 const RURAL_MARKER = "EFFECTCITY_SPECIALIST_RURAL";
+
+// The base improvement catalogue plus its event/DLC adds, mirroring
+// bake-improvement-builds.ts so both tables are built from the same files.
+const IMPROVEMENT_FILE = /^improvement(-event.*)?\.xml$/;
 
 interface SpecialistEntry {
 	zType?: string;
@@ -120,13 +130,20 @@ async function main(): Promise<void> {
 	const textFiles = allFiles
 		.filter((f) => /^text-.*\.xml$/.test(f))
 		.map((f) => resolve(infosDir, f));
+	const improvementFiles = allFiles
+		.filter((f) => IMPROVEMENT_FILE.test(f))
+		.sort()
+		.map((f) => resolve(infosDir, f));
 
-	const [specialistEntries, improvementEntries, ...textFileEntries] =
+	const [specialistEntries, improvementEntryLists, textFileEntries] =
 		await Promise.all([
 			loadEntries<SpecialistEntry>(resolve(infosDir, "specialist.xml")),
-			loadEntries<ImprovementEntry>(resolve(infosDir, "improvement.xml")),
-			...textFiles.map((p) => loadEntries<TextEntry>(p)),
+			Promise.all(
+				improvementFiles.map((p) => loadEntries<ImprovementEntry>(p)),
+			),
+			Promise.all(textFiles.map((p) => loadEntries<TextEntry>(p))),
 		]);
+	const improvementEntries = improvementEntryLists.flat();
 
 	const textByKey = new Map<string, string>();
 	for (const entries of textFileEntries) {
--- a/src/lib/game-detail/BuildComparison.svelte
+++ b/src/lib/game-detail/BuildComparison.svelte
@@ -2,10 +2,10 @@
 	// Head-to-head-by-type module, ported from owglick's H2H card. Every type is
 	// a center-split diverging-bar row (player A grows left, B grows right),
 	// with the absent side left blank when only one player has a type. Used on
-	// the Military tab for the Ending Army / Military Built comparisons and on
-	// the Economy tab for improvements and specialists — hence the icon
-	// category and label being the caller's to choose. A 1v1 framing; the
-	// caller gates it to two players.
+	// the Military tab for the Ending Army / Military Built comparisons, on the
+	// Economy tab for improvements and worker-turns, and on the Specialists tab
+	// for rural / urban specialists — hence the icon category and label being
+	// the caller's to choose. A 1v1 framing; the caller gates it to two players.
 	import SpriteIcon from "./SpriteIcon.svelte";
 	import { formatEnum } from "$lib/utils/formatting";
 	import type { BuildItem, SpriteCategory } from "./helpers";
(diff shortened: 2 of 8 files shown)


alcaras:
Self-review pass.

**Verified:** the old combined `coverage` field is removed with zero remaining references (rather than left computed-for-nobody — the #155 lesson); `specialistSortKey` is consumed by both the Techs tab and this one, so there's exactly one specialist ordering; `COVERAGE_SERIES` colours derive from the `SEGMENTS` table above rather than repeating the hexes.

**Two judgment calls to sanity-check:**
- The side-by-side comparison is fixed to the game's two sides and ignores the tab's nation filter — the filter governs the pivot table and charts. Same spirit as the Military tab's H2H (which has no filters at all), but it is an inconsistency within this tab; one-line change if you'd rather the filter hide it.
- `KIND_LABELS` here repeats two label strings that #169's `PANEL_LABELS` also carries. Left inline as below the extraction threshold for a two-entry record; happy to consolidate if you disagree.

Coverage split rationale in one line: the combined number averaged two different economies — in the reference game Kush's 27% hides 25% rural vs 37% urban, and Persia's 46% hides 41% vs 68%.

becked:
## To move this forward

**1. `SEGMENTS[2]` is "Urban Level 2", not urban.**

`SpecialistsTab.svelte:228-229`:

```ts
{ key: "rural", name: "Rural", color: SEGMENTS[0].color },
{ key: "urban", name: "Urban", color: SEGMENTS[2].color },
```

`SEGMENTS` is `[Rural, Urban Level 1, Urban Level 2, Urban Level 3]`, so index 2 is the middle tier's copper (`#C87941`). The comment says the change makes "the two charts on this tab always agree on what rural and urban look like" — they agree on rural, and on urban the coverage chart matches only tier 2 of the ramp. Two positional indexes into a table above are also exactly what breaks silently when `SEGMENTS` is reordered. Look the colours up by `key`, and decide deliberately which point on the urban ramp represents urban as a whole.

**2. The comparison ignores the tab's nation filter.**

`SpecialistsTab.svelte:268` adds `orderedPlayers` from `orderPlayersUploaderFirst(players, userNation)`, while the cards and both charts run off `displayedPlayers`, which applies `selectedNations` from `tableState.filters`. With a nation filter active the two disagree about who is on screen, and the left/right sides of the comparison can order differently from the cards above them. Derive the comparison from the same filtered list.

**3. `comparisonPanels` duplicates `panels()` from #169.**

`SpecialistsTab.svelte:278` is the third independent copy of the same per-side `Map<group, Map<key, count>>` → union → sort → `BuildItem[][]` aggregation (`EconomyTab.svelte:477` and `:574` are the others). It should call the shared helper — see the note on #168.

**4. `ELIGIBLE_IMPROVEMENTS` misses event improvements.**

The new denominator splits on `ELIGIBLE_IMPROVEMENTS[…].urban`, but `bake-specialists.ts:127` reads only `improvement.xml`, while this stack's own `bake-improvement-builds.ts` reads `improvement.xml` **plus** `improvement-event*.xml`. Any event/DLC improvement that can hold a specialist is missing from both `built` and `staffed`, so the coverage figures are quietly computed over a narrower catalogue than the ledger next door uses. Either widen `bake-specialists.ts` to the same file set or say why the narrow set is correct here.

**5. Redundant guard.**

`SpecialistsTab.svelte:392` — `{#if !matchup && placedCount > 0}` is inside the `{:else}` of `{#if placedCount === 0}`, so the second condition can't be false.

---

becked:
## Rebased onto `main`, review points addressed

Force-pushed as `01bdab1`. #168–#170 have landed, so their commits are gone from this diff and it's down to the single commit. Since the merged versions picked up review follow-ups of their own, a couple of things here had to adapt to APIs that changed on the way in.

### Rebase adaptation

`BuildItem` moved to `helpers.ts` as part of #168's follow-ups, so `import BuildComparison, { type BuildItem } from "./BuildComparison.svelte"` no longer resolved — it now comes from `./helpers`. That was the only thing `svelte-check` caught after the rebase.

### 1. `SEGMENTS[2]` is "Urban Level 2", not urban

Both colours are now looked up by `key` through a `SEGMENT_COLOR` map, so reordering the ramp can't silently repaint them. Urban keeps `#C87941` — the middle tier's copper — but deliberately rather than incidentally: this chart draws one bar for urban against the ramp's three, and the midpoint reads as "urban" where either end would read as one particular tier. The comment says so.

### 2. The comparison ignores the tab's nation filter

`orderedPlayers` now derives from `displayedPlayers`, so the filter governs every surface on the tab.

One consequence worth flagging, since it goes slightly beyond the filter case: `displayedPlayers` descends from `columnPlayers`, which already excludes players who placed no specialists. So a two-player game where only one side placed any now shows the "needs exactly two nations" note instead of rendering the comparison with one empty half. That's consistent with the cards above it, which already listed only the one player.

### 3. `comparisonPanels` duplicates `panels()`

`comparisonRowKeys` landed in `helpers.ts` while this branch was out, and it's the narrow version — union plus a caller-supplied comparator, with each tab keeping its own aggregation. The hand-rolled union+sort here is replaced with a call to it, passing `specialistSortKey` with a display-name tiebreak, matching how `EconomyTab.panels()` calls it. Per-side rows are now handed over raw for `BuildComparison` to gap-fill, rather than pre-padded.

### 4. `ELIGIBLE_IMPROVEMENTS` misses event improvements

Real, and there's exactly one instance. `bake-specialists.ts` now reads `/^improvement(-event.*)?\.xml$/` — the same file set as `bake-improvement-builds.ts` — which adds `IMPROVEMENT_LAURION_MINE: { urban: false }`, taking the table from 97 eligible improvements to 98. It declares `<Specialist>SPECIALIST_MINER</Specialist>` and lives only in `improvement-event.xml`, so a staffed Laurion Mine was previously missing from *both* `built` and `staffed`.

Two notes on scope. The three files define disjoint `zType` sets, so the merge needs no precedence rule and I didn't port `bake-improvement-builds.ts`'s first-wins guard, which would be dead code here. And the file set stops at the event adds: `Mods/Carthage{2,3,4}/Infos/improvement-add.xml` also declare specialist slots (`BALEARIC_RANGE`, `MONASTERY_CARTHAGE`, `TEMPLE_CARTHAGE`, all urban), but those are campaign-mod improvements and no generated table covers `Mods/` for this purpose today. Deliberately left out.

One correction to my own review note: I said the ledger next door already used a wider catalogue. It reads the wider file set, but `LAURION_MINE` has no `<iBuildTurns>` — it's event-spawned, never built — so `bake-improvement-builds.ts` drops it anyway. No generated table knew about it before this change.

### 5. Redundant guard

Dropped.

### Also

Two comments on `main` claimed a specialists comparison lives on the Economy tab, which it doesn't — `EconomyTab.svelte`'s "Side-by-side comparisons" header and `BuildComparison.svelte`'s module comment. This PR is where specialists actually land, so both are corrected here.

### On the two judgment calls from the self-review

The nation-filter one is resolved above. `KIND_LABELS` is left inline — `PANEL_LABELS` is still local to `EconomyTab.svelte` on `main`, so consolidating would mean promoting a three-entry record to a shared module to save two strings. Agreed it's below the threshold.

### Verification

- `svelte-check` — 0 errors, 0 warnings
- `npm run lint` — clean
- `prettier --check` on the changed files — clean
- `cloud/` tests — 578 passed, 4 failed, all four in `canonical-map-options.test.ts`. That's the known DOTA schema drift (#115/#141), already red on `main`; neither input to the failing assertion is touched by this branch.

PR #172, opened 2026-07-27
feat(parser): projects_produced — PARSER_VERSION 2.13.0

alcaras:
Projects are the third thing a city builds besides units and (via workers) improvements, and the blob carried nothing about them. This adds the Player node's `ProjectsProduced` map — the whole-game record of every project completed, with counts, surviving city loss — and renders it as a Projects panel in the Economy tab's Built section.

> **Stacked on #168 + #169** (it renders into the Economy tab). Different risk profile from the rest of the stack: this one bumps `PARSER_VERSION`.

## Parser

`parseProjectsProduced` sits beside `parsePlayerResources` and parses the same way (`collectStrictNamedInts` over a name-keyed int map). Wire shape `{player_xml_id, project, count}`, matching the newer per-player arrays. `PARSER_VERSION` → **2.13.0**.

Median 20 projects per player across the 400-game cached corpus; most-built are Festival, Treasury I, Forum I, Inquiry, Archive I, Walls.

## Worker

One change: `2.13.0` added to `KNOWN_PARSER_VERSIONS` + `CURRENT_PARSER_VERSION`, with a history comment. The field itself flows through the existing `looseObject` — no schema shape change, no migration, no new derived tables.

**Deploy order: worker first, frontend after** — otherwise uploads carrying `parser_version: 2.13.0` are rejected as unknown. And only new uploads get the field: reindex rebuilds from the blob, and old blobs never had it, so the panel appears gradually. It hides on older games rather than implying nobody built projects.

## One display decision worth a look

`formatEnum` deliberately strips trailing digits ("Garrison 1" → "Garrison") — correct where tiers aggregate, wrong here where Council 1/2/3 are distinct accomplishments on distinct rows. A local `projectLabel` reuses `formatEnum` and puts the tier back. First cut rendered three identical "Council" rows, which is how this was caught.

No 2D project art is baked, so rows render label-only (SpriteIcon degrades to nothing for missing sprites; alignment is preserved by the fixed icon slot).

## Verification

- New parser run against a real save: 33 project rows extracted, spot-checked against the raw XML.
- Full pipeline test: save parsed with 2.13.0 and uploaded through the local worker — accepted, reimported `2.10.0 → 2.13.0`, panel renders with per-tier rows.
- Cloud suite: 516 passed / 4 failed — the 4 are the pre-existing `canonical-map-options` drift on `main`, untouched here.
- `npm run lint`, `svelte-check`, cloud `tsc` clean.

Follow-ups deliberately out of scope: per-city `ProjectCount` (would enable projects-by-family), civics pricing via `project.xml` `iCost`.


alcaras:
Self-review pass.

**Verified:** every new symbol has consumers (`parseProjectsProduced` → `extractAllGameData`, `ProjectProducedInfo` → wire + tab props, `projectPanel`/`projectLabel` → the Built section); the parser mirrors `parsePlayerResources` beside it, same `collectStrictNamedInts` strictness — a malformed map fails the parse the same way a malformed `YieldStockpile` would, which is the sibling's existing behaviour rather than a new failure mode.

**The bug the end-to-end test caught:** first render showed three identical "Council" rows — `formatEnum` strips trailing digits by design, wrong here where Council 1/2/3 are distinct rows. `projectLabel` reuses `formatEnum` and puts the tier back.

**Disclosed:**
- The panel sits inside the same 1v1 gate as the other Built comparisons; the FFA fallback text mentions the improvements table but projects have no equivalent table below — an FFA game's projects are simply not shown anywhere yet.
- No project art is baked, so rows are label-only (SpriteIcon degrades to nothing; the fixed icon slot keeps alignment).
- Deploy order and the new-uploads-only caveat are in the description — the panel appears gradually as saves are re-uploaded, and hides otherwise rather than implying nobody built projects.

becked:
## To move this forward

**1. The Projects panel is invisible in FFA and solo games.**

`EconomyTab.svelte:822` puts the panel inside the `{#if matchup}` block, so `projects_produced` renders only for exactly two nations. Unlike improvements — which fall back to the "Every improvement" pivot table, and whose `!matchup` note explicitly says so — projects have no other surface on the tab, so in an FFA the data is parsed, shipped and then shown nowhere. Either give it a per-player fallback (the ledger cards' shape would work) or extend the existing note to say projects need 1v1 too. A panel that silently disappears is the case the tab's other empty-states were written to avoid.

**2. `iconCategory="improvements"` for `PROJECT_*` keys.**

`EconomyTab.svelte:838` passes the improvements category for project zTypes. It renders only because `SpriteIcon` degrades to nothing on a miss — but it's a lookup into the wrong manifest, and it'll start resolving wrong the day a `PROJECT_*` and `IMPROVEMENT_*` key collide. `SpriteCategory` has no `projects` member, so the honest fix is making `iconCategory` nullable in `BuildComparison` (render the fixed icon slot empty) rather than naming a category that isn't the right one.

**3. `projectPanel` is the third copy of the aggregation.**

`EconomyTab.svelte:574` rewrites the same per-side counts → union keys → sort → `BuildItem[][]` shape as `panels()` at `:477` and `comparisonPanels` in #171. It differs only in being flat (no group axis) and summing `row.count` instead of 1. Call the shared helper — see #168.

**Not blocking, for the record:** the version bump itself is wired correctly — `KNOWN_PARSER_VERSIONS` + `CURRENT_PARSER_VERSION` both updated, additive field riding `looseObject`, no migration. Worker-before-frontend deploy order is the operative constraint at merge time.

---

becked:
## All three addressed, and rebased onto current `main`

The branch now carries only its own change — #169 and #171 landed, and their two commits here were patch-identical to what merged, so the rebase dropped them. Three commits: the parser bump as it was, then one commit per review point group.

**1. Invisible in FFA and solo — fixed, and it drove the bigger change here.**

`BuildComparison` grows a second mode rather than the tab growing a second way to draw a row. Omit `b`/`cb` and it renders a ledger: one bar per row anchored left with the count trailing it, instead of a center split. The duel branch is the markup it always was, so the Military, Specialists and improvement panels render unchanged. `title` went optional alongside it — the ledgers head themselves with a crest and the nation's colour, and an empty title bar is dead height.

Two things had to become shareable for a wall of ledgers to read as one picture instead of four unrelated charts:

- `max` — one bar scale across every nation, the busiest single nation-and-project cell in the game. Per-panel scaling drew a nation with two projects exactly like one with twenty.
- `labelWidth` — the label column was a hardcoded `110px`, sized for unit and improvement names. Project names run well past it ("Suppress Dissent Christianity") and were being truncated, so the Economy tab now sizes the column off the longest name in the game. Both projects panels take it: **the duel panel was clipping the same names.**

Each ledger lists only what its own nation built, biggest first — handing every card the shared row order would pad it with a blank row for each project its neighbours built and it didn't. They lay out in columns rather than a grid, because nations build wildly different numbers of projects and a grid row is as tall as its tallest cell, so a short nation beside a long one inherits its neighbour's gap.

**2. `iconCategory="improvements"` for `PROJECT_*` keys — fixed.**

`iconCategory` is nullable and the projects panels pass `null`. The fixed `w-3.5` slot still reserves its width, so label-only rows line up with rows that have icons, and there's no lookup into a manifest that would start resolving wrong on a `PROJECT_*`/`IMPROVEMENT_*` collision.

**3. Third copy of the aggregation — gone, using the helper from #168.**

`comparisonRowKeys()` had landed in `helpers.ts` by the time this rebased, so `projectPanel` is deleted. `projectsByPlayer` is the one per-nation source, `projectItems` hands over raw rows for the component to gap-fill itself, and `projectKeys` calls the shared helper — with a comparator, because ties have to break on the displayed name the way the improvement panels do. On raw zTypes the underscore sorts ahead of a letter, which would put Councillor above Council 1.

---

**Verification.** `npm run build`, `npm run lint`, `svelte-check` (0/0) and Prettier clean. Cloud suite 578 passed / 4 failed, the four being the `canonical-map-options` DOTA drift that is red on `main` and untouched here. The ledgers were rendered against a real four-nation FFA save — that pass is what caught the truncated labels and the blank placeholder rows. The column layout went in after that render and has not been eyeballed since.

**Legacy `web/` viewer.** Its `src/lib/game-detail` symlinks to this one, so these changes do reach it. Its `check` and `build` currently fail — but they fail on `main` too: `GameDetailView.svelte` there already imports `$lib/map-settings`, and `web/src/lib` carries no `map-settings`, `parser`, `echarts` or `ui` symlink at all. Nothing in the failure names `BuildComparison`, and this PR touches no file under `web/`. Flagging it as a thing that exists, not as a thing this PR did.

**Still true from the description:** only new uploads carry `projects_produced`, so the panel appears gradually rather than implying nobody built projects on older games.

---

becked:
Added the project assets on top of this. A new `npm run bake:project-icons` bakes the project art into the sprite atlas and generates the `PROJECT_ICON` name table at `src/lib/generated/project-icons.ts`, so the Economy tab's project ledgers draw real icons instead of bare labels.

Also on the branch: per-nation project ledgers for FFA games (no side-by-side pivot table to fall back on there, so the parsed data would otherwise ship and show nowhere), and the panel now owns its own icons and row order.

PR #173, opened 2026-07-27
feat(game-detail): Wonders tab — the catalogue with this game painted on

alcaras:
A Wonders tab on game detail: one row per culture tier, every wonder in the catalogue, with this game's reality painted on — wonders the save **disabled are faded** ("Not in this game"), wonders still open are plain, and a **built** wonder carries its builder's colour, nation and completion turn. One glance answers what was in the pool and who got what.

> **Stacked on #168 → #169 → #172.** Their commits lead this diff and disappear as they land.

## No new data

Everything already ships: `WONDER_CULTURE_PREREQ` + `CULTURE_LEVELS` from the wonders bake (#157), `disabled_improvements` from 2.12.0 blobs (#157), `player_wonders` for builds, and the 2D wonder art. `disabled_improvements` sits on `gameDetails`, so there's no new prop threading at the page level.

## Two semantics worth checking

- **Unknown ≠ nothing-disabled.** Pre-2.12.0 blobs carry `disabled_improvements: null` — the pool is unknown, which is a different claim from the empty list. Nothing fades on those saves, and a one-line note says why.
- **Built beats disabled.** A wonder that demonstrably completed was in the pool, whatever the end-state list says (captures/edge cases). The card shows the build.

Builder resolution goes through the mirror-match-safe list (`findByPlayer`: id match, nation fallback), with the wonder row's own nation/name covering blobs where neither resolves.

Wonders also **leave the Economy tab's Built comparisons** — they were the odd group out among rural/urban worker economy, and they have a home now.

`npm run lint` and `svelte-check` clean; screenshot-verified against a live game (Pyramids → Egypt T21 bordered orange, Yazilikaya/Musaeum → Maurya T84/T58 purple, disabled pool faded per tier).

Heads-up on merge order: this and #170 both touch the `GameDetailView` tab list on adjacent lines, so whichever lands second will need a trivial rebase.

### DIFF
--- a/src/lib/game-detail/EconomyTab.svelte
+++ b/src/lib/game-detail/EconomyTab.svelte
@@ -548,7 +548,7 @@
 
 	const improvementPanels = $derived(
 		panels(
-			["rural", "urban", "wonder"],
+			["rural", "urban"],
 			(imp) => IMPROVEMENT_BUILDS[imp.improvement]?.kind ?? null,
 			(imp) => (IMPROVEMENT_BUILDS[imp.improvement] ? imp.improvement : null),
 			(key) => IMPROVEMENT_UNLOCK_COST[key] ?? 0,
@@ -561,7 +561,7 @@
 	// cost, since that's the question this panel answers.
 	const workerTurnPanels = $derived(
 		panels(
-			["rural", "urban", "wonder"],
+			["rural", "urban"],
 			(imp) => IMPROVEMENT_BUILDS[imp.improvement]?.kind ?? null,
 			(imp) => (IMPROVEMENT_BUILDS[imp.improvement] ? imp.improvement : null),
 			null,
@@ -664,10 +664,10 @@
 		})),
 	);
 
+	// Wonders have their own tab; the Built panels carry only worker economy.
 	const PANEL_LABELS: Record<string, string> = {
 		rural: "Rural",
 		urban: "Urban",
-		wonder: "Wonders",
 	};
 </script>
 
--- a/src/lib/game-detail/EconomyTab.svelte
+++ b/src/lib/game-detail/EconomyTab.svelte
@@ -470,9 +470,9 @@
 
 	// ─── Side-by-side comparisons ─────────────────────────────────────
 	// The Military tab's diverging-bar module, pointed at what this tab counts.
-	// One panel per kind — rural / urban / wonders — because that split is the
-	// strategic axis: a wide farm empire and a tall urban one show up as
-	// different blocks rather than one undifferentiated list. 1v1 like the
+	// One panel per kind — rural / urban — because that split is the strategic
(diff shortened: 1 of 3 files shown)


alcaras:
Self-review pass.

**Verified:** the four culture-tier icons all exist in the sprite manifest (`icons/CULTURE_WEAK`…`LEGENDARY`); the "wonder" group is fully gone from both Economy Built panel lists; the ledger cards keep their wonders count as a summary number, which is deliberate — the detail moved, the headline stayed; builder resolution uses `findByPlayer` (id match, nation fallback — the shared idiom), with the wonder row's own nation/name covering blobs where neither resolves.

**Disclosed:**
- A built wonder whose builder can't be resolved to the player list renders with a transparent border and the nation name in tan — information preserved, colour dropped.
- Cards fade with `opacity + grayscale` while keeping layout (transparent border placeholder), so a row doesn't reflow between games with different pools.
- The bake's `CULTURE_LEVELS` carries the four wonder-prereq tiers; `CULTURE_ESTABLISHED` exists for cities but no wonder requires it, so it has no row.
- Merge-order note from the description stands: this and #170 touch adjacent tab-list lines; second one in needs a trivial rebase.

becked:
## To move this forward

**1. A built wonder missing from the catalogue vanishes from the tab.**

`WondersTab.svelte:51` builds every row from `Object.keys(WONDER_CULTURE_PREREQ)`, and `:85` drops tiers with no cards. A wonder that isn't in that table is not rendered at all — including one this game demonstrably completed, which is the opposite of the "built beats disabled" rule the PR body sets out.

This is reachable, not hypothetical: `bake-wonders.ts:85-86` reads only `improvement.xml`, while this stack's own `bake-improvement-builds.ts` reads `improvement.xml` **plus** `improvement-event*.xml`. So an event/DLC wonder is priced as `kind: "wonder"` in the Economy ledger and absent from the Wonders tab. Either widen `bake-wonders.ts` to the same file set, or reconcile `player_wonders` against the catalogue and surface the leftovers (an "Other" row) so a completed wonder can never silently disappear.

**2. Pulling wonders out of the Built panels leaves the numbers unreconcilable.**

This PR narrows both panel groups to `["rural", "urban"]`, but `playerEconomies` is unchanged, so on the same screen:

- `EconomyTab.svelte:745` — the headline `{eco.workerTurns}` still includes wonder turns
- `EconomyTab.svelte:760` — the card still shows `countByKind.wonder` as a "wonders" stat

…above panels that now total rural + urban only. A reader adding up the panels won't reach the headline. Either drop wonder turns from the ledger totals when they move out, or keep the wonders stat and say in the card that the panels exclude it.

**3. Stale comment.**

`EconomyTab.svelte:470` still reads "One panel per kind — rural / urban / wonders" after this PR removes the wonder group.


---

PR #174, opened 2026-07-28
perf: cut the blocking D1 round trips on the anonymous game-detail path

becked:
Implements items 1–3 of the closing plan on #150, plus test coverage for the one that reorders a gate.

Static reading establishes which queries are *independent*; it cannot establish payoff. No latency claim here is measured — #150's instrumentation item is still blocked on a Logpush destination. What these changes do is remove sequencing that was never required, which is verifiable by reading.

## The changes

**1. `getGame` + `getGameTournamentLink` run concurrently** (`src/routes/games/[id]/+page.ts`)

`getGameTournamentLink` takes only `params.id` and reads nothing from the game payload, so awaiting it after `getGame` was a false dependency. The `.catch()` is attached where the promise is created rather than at the await, so a `getGame` throw can't leave it rejecting unhandled. Error semantics are otherwise unchanged.

**2. `handleMe`: 4 D1 round trips → 2** (`cloud/src/auth.ts`)

- `isSiteAdmin()` re-selected `discord_id` for a user_id the handler already holds on `row`. Comparing inline deletes that round trip. `isSiteAdmin` is unchanged and keeps its six other call sites.
- `claimTournamentSlots` and the beta lookup both need only the users row and neither reads the other, so they run concurrently. `Promise.all`, not `db.batch()` — batch is transactional, and a claim failure must not take the beta read with it. Deferring the claim to `waitUntil` is also not an option: `+layout.ts` calls `getMyTournaments` straight after, and that handler reads exactly what the claim writes.

No behavior change.

**3. `handleGameDetail`: the anon read-limit count overlaps the games-row read** (`cloud/src/games.ts`)

A sessionless request cannot be the owner, so non-ownership is certain before either query runs — exact, not speculative. Signed-in requests keep the sequential path (they may be the owner, and owners aren't counted), falling back to the query at the gate.

Only the read-only `COUNT` moved. Gate order is untouched — 404, then 401/403, then 429, then the audit insert — and the insert still sits behind the visibility gate, so a request that was never served still spends no budget.

## Tests

`cloud/test/integration/games/anon-read-rate-limit.test.ts` — 9 tests. The 429 path on this route had no coverage at all, and change 3 reorders the queries behind it, so these pin the *order*, not just the cap.

Both load-bearing properties were confirmed by mutation:

| Mutation | Tests red |
|---|---|
| Drop the signed-in fallback count | 1 |
| Move the audit insert above the visibility gate | 6 |

The game seeder was extracted from `blob-cache.test.ts` into `cloud/test/helpers/games.ts` rather than copy-pasted (`blob-cache.test.ts` is a pure move, asserting exactly what it did before).

Seeding fills the bucket with one recursive-CTE statement rather than per-row inserts. The row-at-a-time version put enough extra work through the shared test runtime to tip already-marginal timing-sensitive tests in *other* files into timeouts; the CTE version measures at the clean-tree baseline over 5 full-suite runs.

## Not in scope

- The `countEventsSince` → per-POP counter swap (#150 plan item 2) — a real accuracy-for-latency trade.
- KV-caching `/v1/games/public-recent` (plan item 4).
- Instrumentation and Smart Placement — operator/config decisions.

## Verification

- `cloud/` `tsc --noEmit` — clean
- `npm run lint`, `svelte-check` (1284 files, 0 errors) — clean
- `prettier --check` on all changed files — clean
- `cloud/` `npm test` — 525 passing

Two pre-existing failures on `main` are unaffected and unrelated: the 4 `canonical-map-options` tests (DOTA schema drift, #115/#141), and `users/search.test.ts > does NOT count q<2 searches`, which flakes at 2/5 runs on a clean tree and at the same rate here.

Refs #150.

https://claude.ai/code/session_01AxG2y1YZXmKYKQMSBR3Rhw

### DIFF
--- a/cloud/src/auth.ts
+++ b/cloud/src/auth.ts
@@ -36,7 +36,6 @@ import {
 	sessionFromRequest,
 } from "./session";
 import type { SessionEnv } from "./session";
-import { isSiteAdmin } from "./admin";
 import { displayNameSql } from "./identity";
 import { logError, setSecurityReason } from "./log";
 import type { QueryableD1, EventsEnv } from "./d1";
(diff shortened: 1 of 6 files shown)


becked:
Ready to merge from my side. Nothing has been pushed since this PR was opened — the five commits below are the whole branch. Posting a commit-level walkthrough since the description is organized by change rather than by commit.

| Commit | What it does |
|---|---|
| `86f0489` `perf(games)` | Starts the tournament-link read before awaiting `getGame`. `getGameTournamentLink` takes only `params.id`, so the sequencing was a false dependency. `.catch()` is attached at promise creation, not at the await, so a `getGame` throw can't leave it rejecting unhandled. |
| `26772a6` `perf(auth)` | `handleMe` 4 D1 round trips → 2. Inlines the admin comparison against `row.discord_id` (which `isSiteAdmin` was re-selecting), and runs the slot claim + beta lookup concurrently. `isSiteAdmin` unchanged, keeps its six other call sites. |
| `57cc813` `perf(games)` | `handleGameDetail` issues the `anon_read` count alongside the games-row read when `session === null`. Only the read-only `COUNT` moved; the audit insert stays behind the visibility gate. |
| `7c10081` `test(games)` | Pure move of `makeBlob`/`putBlob`/`seedGame` out of `blob-cache.test.ts` into `cloud/test/helpers/games.ts`, so the new test doesn't copy ~45 lines of seed SQL. `blob-cache.test.ts` asserts exactly what it did before. |
| `9ebfb5c` `test(games)` | The 9 gate-order tests, plus exporting `ANON_READS_PER_HOUR` for them to pin against. |

Each commit is independently valid — the `ANON_READS_PER_HOUR` export sits with the test that consumes it, and `57cc813` keeps the constant private with no test file present, so `tsc` and the suite pass at every point in the branch.

**Two notes for review.**

`26772a6` deliberately uses `Promise.all` rather than `db.batch()`: `batch()` is transactional, and a slot-claim failure must not take the beta read with it — `claimTournamentSlots` swallows its own errors precisely so a claim hiccup can't break `/v1/auth/me`. Moving the claim to `waitUntil` is likewise not an option, because `+layout.ts` calls `getMyTournaments` immediately after and that handler reads exactly what the claim writes.

`57cc813` is the one that reorders a gate, which is why `9ebfb5c` pins the order rather than just the cap. Both load-bearing properties were confirmed by mutation — dropping the signed-in fallback count turns one test red, moving the audit insert above the visibility gate turns six red — and both mutations were reverted before committing.

**Verification** is the local run: `cloud/` `tsc --noEmit`, `npm run lint`, `svelte-check` (1284 files, 0 errors), `prettier --check` on every changed file, and `cloud/` `npm test` at 525 passing. CI here is CodeQL only, so the green checks above are not the test suite.

Two failures are pre-existing on `main` and untouched by this branch: the four `canonical-map-options` tests (DOTA schema drift, #115/#141), and `users/search.test.ts > does NOT count q<2 searches`, which flakes at 2/5 full-suite runs on a clean tree and at the same rate here. I measured that baseline explicitly because an earlier revision of the new test seeded its rate-limit bucket row-by-row and did add load-induced flakiness elsewhere; seeding via a single recursive-CTE statement brought it back to the clean-tree rate.

No migrations in this branch, so the deploy carries no schema step. Plan and follow-ups are in becked/per-ankh#150.

PR #175, opened 2026-07-28
feat: Momentum — a per-turn win-probability curve for duels

alcaras:
## What

Ports the momentum model from owglick (`docs/momentum-model.md` there) to per-ankh. For finished 1v1 games:

- **Overview tab** gets a **Momentum** chart above the nation cards: P(uploader-side wins) per turn, the area between the line and the 50% midline filled in whichever nation currently leads. Below it, a detail panel — hover explores any turn, click pins it, defaults to the final turn — showing each side's raw key stats, the per-dimension decomposition of the current lead, what moved since the previous turn, and events from that window (log events plus battles derived from military-power drops).
- **Home feed** duel cards swap the VP sparkline for the same momentum curve (VP stays as the fallback for everything else, and for duels not yet scored).

## Screenshots

**Overview tab** — the curve with the leader's fill, detail panel pinned on the final turn (the derived battle event and the tech from that window on the right):

momentum panel on the Overview tab

**Hovering a mid-war turn** (T86) — level decomposition of Carthage's 84%, what moved since T85, and the trade both armies paid for:

hovering turn 86

**Home feed** — duel cards swap the VP sparkline for the momentum curve; non-duels keep VP:

feed cards

## The model — background and how the number is computed

(Context, since owglick — where this was designed and first validated — is a private repo. The in-repo authority is the header of `scripts/bake-momentum.ts`; this is the prose version.)

**The question it answers.** For a finished 1v1, "who was winning at turn T, and by how much?" — a single 0–100% number per turn, plus an exact breakdown of *why*. It's a retrospective reading of a finished match, not a forecast: the weights are fitted on a corpus of finished duels, and game progress (T / final turn) is an input, so the curve only exists once the match is over.

**Six dimensions, stored as differences.** At each turn we take player A − player B on: **cities** (owned city centres, reconstructed from `tile_ownership_history`), **growth** (`YIELD_GROWTH` rate), **orders**, **science**, **eco** (a −5…+5 majority vote across the five resource incomes — money, food, iron, stone, wood — each counting +1/0/−1 for who earns more, which resists one silly market price dominating), and **military** (power difference divided by the pair's mean power, because absolute power grows ~20× over a match and a 200-power lead means everything at T20 and nothing at T90). Differences rather than raw values: almost everything in a 4X correlates with "empire is bigger", and differencing kills that collinearity.

**Standardise, then score.** Each difference is divided by the corpus standard deviation *at that turn* (smoothed ±7 turns), so "a 2-city lead" is measured against how big city leads typically are at that point of a game. The standardised vector **z** is scored by a logistic regression with **no intercept** — that one omission makes the model antisymmetric by construction: swap the players and you get exactly 1 − p, and a perfectly tied position is exactly 50%.

**Weights change over the game.** One fixed weighting misreads both ends of every match — growth/cities front-load, military back-loads. So the logistic is fitted separately in five game-progress buckets (0–30%, 30–50%, 50–70%, 70–85%, 85–100%; L2-regularised, Newton/IRLS). In the fitted weights you can watch growth fade (0.34 early → 0.57 mid) while military climbs to dominance (1.46 in the endgame).

**The decomposition is exact, not a heuristic.** p = σ(w·z), so each dimension's contribution w_j·z_j sums to the log-odds — that's what the panel's "behind the lead" bars are. The change bars hold the weight and scale fixed at the later turn and multiply by the z-delta, so a stat that didn't move contributes exactly zero (differencing the level contributions instead would fold normaliser drift into phantom "changes").

**Fit and validation.** Fitted on 384 finished duels from per-ankh's own corpus via `npm run bake:momentum` (needs a local blob directory; deliberately not in `bake:all`) — deduped on the save's `xml_game_id` so a match both players uploaded counts once, all balance eras kept (a held-out test showed the extra data beats era purity; see the model-update comment below). The bake fails hard on its validation suite — tile-index integrity, early curve coverage, and the growth-peaks-before-military shape — and this fit independently reproduced the weights from the original design within noise. Discrimination on the corpus: AUC 0.71 / 0.77 / 0.82 at 30% / 50% / 70% of game progress — i.e. genuinely uncertain early, quite sure late, which is what you'd want from an honest "who's winning" line.

**Why the weights are static.** `MOMENTUM_MODEL_VERSION` is baked next to the numbers. A finished game's curve must never drift between visits because someone re-ran a fit; refits are an explicit act (an Old World balance patch would be the occasion), and bumping the version documents it.

## Mechanics

- Scoring is pure arithmetic over existing blob fields: `src/lib/game-detail/momentum.ts`, byte-mirrored at `cloud/src/momentum.ts` with a drift test (`momentum-mirror.test.ts`, the canonical-map-options pattern) that also pins the spec's invariants — antisymmetry, Σ contributions = log-odds, unchanged stat contributes exactly 0. That last one caught a real IEEE −0 display bug during development.
- The Worker scores duels at derive time (upload + reindex) into a new `game_player_turn.momentum` column (**migration 0041** — renumbered after upstream took 0039/0040 for profile slugs); the public-recent feed reads it as `momentum_series` per player. Two humans + known winner + ≥10 turns gates scoring; everything else is untouched.
- Cities per turn are reconstructed from `tile_ownership_history` against end-state `map_tiles` (positional index = `tile_xml_id`, validated at bake time), so `GameDetailView` grows `tileOwnershipHistory`/`rawMapTiles` props — end-state tiles, not the map-slider-mutated ones.

## Deploy order

1. Migration 0041 (`game_player_turn.momentum REAL`)
2. Worker (writes momentum at derive time; feed exposes `momentum_series`)
3. Frontend (Overview panel + feed cards; both no-op gracefully while the column is empty)
4. Admin reindex sweep — backfills the whole corpus, no re-uploads needed (verified locally: 387 games scored)

## Testing

- `cloud/`: mirror + invariant tests green; full suite green except the 4 pre-existing `canonical-map-options` failures on `main` (missing `MAP_OPTIONS_MULTI_DOTA_INNER_TERRAIN` in the cloud mirror — unrelated, happy to fix separately).
- `npm run lint`, `svelte-check`: clean.
- Verified live locally on real games: fill colors, hover/pin, decomposition arithmetic cross-checked against D1 by hand, feed fallback for non-duels.


alcaras:
Self-review (correctness + fit against CLAUDE.md's contributing checks), findings fixed in the two `refactor:` commits on the branch:

**Found and fixed**

- **Dead export** — `momentumDrama` (drama/comeback metrics) shipped with no consumer; it was speculative for future feed badges. Removed from both copies and the mirror test; it returns with whatever feature actually reads it.
- **Duplicated hex literal** — the 50% midline colour `#6b6459` appeared verbatim in both `MomentumPanel` and `RecentSaveCard`. Extracted to `CHART_REFERENCE_LINE_COLOR` in `$lib/config` next to the rest of the chart palette.
- **Stale comments** — two comments still described the decomposition as living in "the tooltip" (it moved to the pinned detail panel during development). Refreshed, mirror kept byte-identical (drift test green).
- **Bake registration** — `bake:momentum` added to `package.json` alongside the other bake scripts, and `MOMENTUM_CORPUS_DIR` documented in `.env.example`. Deliberately not in `bake:all`: refitting silently on someone's local corpus would drift every finished game's curve.

**Checked, no finding**

- *Parallel surfaces*: `RecentSaveCard` has exactly one render call site (home feed); the other references are comments. `momentum_series` is typed on `PublicRecentPlayer` only and the worker populates it in `handlePublicRecentGames` only, so type and data agree.
- *Colors*: player series come from `getNationChartColor`/`resolveDetailPlayers`; panel chrome uses CSS variables; no other literals.
- *Null handling*: `??`/`!= null` throughout the scoring lib; the one `||` (`stripMarkup(...) || formatEnum(...)`) is a display fallback.
- *Writers/readers*: both writers of `game_player_turn.momentum` (upload + reindex) share `buildGamePlayerTurnStatements`, so the duel gate can't diverge between paths.
- *Runes*: no conditional-read `$effect` traps; chart wiring reads state unconditionally.
- *Event kinds*: all nine `log_type` keys in the panel's KIND map verified against real blob data (including the rare `TEAM_DIPLOMACY`/`OCCURRENCE`).
- *Migration*: `0039` numbered after current `main`'s `0038`, forward-only, additive.

- The tournament `MatchPopover` still draws its own VP sparkline from the blob — a candidate for the same momentum-for-duels treatment in a follow-up, but it reads different data (client-side blob, not the feed endpoint) so it's real scope.
- The full `cloud/` suite has 4 pre-existing `canonical-map-options` failures on `main` (cloud mirror missing `MAP_OPTIONS_MULTI_DOTA_INNER_TERRAIN`) — unrelated to this PR; happy to fix separately.

alcaras:
## Model update: corpus fix (fefeb1e) + rebase on current main

Your review of the model doc (thank you — both passes were excellent) triggered a full experiment round, and it caught something real in **this PR's bake**, so the branch is updated.

**What was wrong.** `bake-momentum.ts` fitted on every blob in the corpus directory: no dedupe — a match uploaded by both players entered twice, double-weighting tournament games and leaking same-match pairs into any held-out evaluation — and no era bound, silently mixing saves back to 2022.

**What the evidence said, which surprised us.** We first "fixed" the era mixing with a 2026-05-07 patch cutoff — and a held-out 2×2 (corpus × SD window, tested only on current-era games, grouped folds pinned across variants) reversed it: **all-era training beats the cutoff on the modern era itself** — AUC 0.782 vs 0.773, and the confident-and-wrong rate (>80% shown for the eventual loser) at 50–85% progress **halves** (2.7% vs 5.4%). The per-turn-standardised A−B features are era-robust, and at ~200 modern duels, sample size binds harder than balance drift — the late buckets, where the model makes its boldest claims, gain the most. The bake header documents revisiting the cutoff once the modern corpus alone reaches ~350 duels.

**What shipped** (one commit, `fefeb1e`): dedupe on `xml_game_id` keeping the longer upload (384 duels from 458 blobs), all eras kept deliberately, SD smoothing widened ±3 → ±7 (cuts the phantom-change residual's p95 ~13% at no CV cost). Fitted weights moved modestly; the late military weight eases 1.55 → 1.46.

One game shows the stakes better than the tables — a 91-turn tournament duel where Carthage dominates economically for 40 turns and Hittite wins the closing war by conquest. The endpoint under three fits:

| fit | final-turn reading | actual |
|---|---|---|
| original (dupes, all eras, ±3) | Hittite 68% | Hittite won |
| patch-cutoff only (~210 duels) | **Carthage 52%** — wrong side | |
| **shipped** (deduped all eras, ±7) | Hittite 63% | ✓ |

original fit
shipped fit

Also in this push: rebased onto today's `main` (the Economy/Families merges landed on the same game-detail files; `tileOwnershipHistory` is now shared by both consumers and the union resolution keeps your merged comments). Full doc-side answers to both reviews — calibration (reliability hugs the diagonal, 0.9+ bin wins 97%), bootstrap CIs, the absolute-turn bucket refit, the `YieldTotalHistory` stockpile confirmation — live in owglick's `momentum-model.md` §11.

Checks: lint / svelte-check clean, cloud 720 passing (the 4 `canonical-map-options` failures remain the local Reference-XML drift).

becked:
Rebased onto current `main` (was 162 commits behind) and force-pushed to `feat/momentum-pr`, now `5cb800e`. Two trivial import/script collisions, plus one that needed a decision: `main` has since taken 0041 (`featured_videos`) and 0042 (`drop_legacy_share_tables`), so the momentum migration is now **`0043_momentum.sql`** — deploy step 1 in the description needs that number. `lint`, `svelte-check`, and the cloud suite are green apart from the four pre-existing `canonical-map-options` failures on `main`, which this PR doesn't touch.

Five findings. The first two are user-visible today.

## 1. The curve has four structural discontinuities in every game

Weights switch abruptly at progress 0.30 / 0.50 / 0.70 / 0.85 and nothing interpolates. I ran a synthetic 100-turn duel where every lead is held constant — steady relative army lead, steady yield gaps, no cities changing hands. The line still jumps:

```
T29 → T30   52.8% → 55.2%    (+2.4)
T49 → T50   52.8% → 58.9%    (+6.1)
T69 → T70   56.9% → 69.6%   (+12.7)
T84 → T85   67.7% → 77.3%    (+9.6)
```

Nothing happened in the game. `mil` goes from `-0.003` to `+0.7305` crossing bucket 2→3, so anyone holding an army lead gets a free ~13-point leap at exactly 70% of the way through — and every reader will interpret that as a battle.

The bucket-fitting decision itself is right; it's the hard boundary that isn't. Interpolating the weight vector between bucket centres (or making progress a continuous interaction term) preserves the front-load/back-load behaviour without the cliffs.

## 2. …and the detail panel contradicts itself at those turns

`ch` holds weight and scale fixed at the later turn: `ch_j = w_new × (z_t − z_{t−1})`. The actual move is `Σw_new·z_t − Σw_old·z_{t−1}`. The difference — `Σ(w_new − w_old)·z_{t−1}` — is exactly the bucket jump, and it appears in no bar.

The panel header uses the real delta (`MomentumPanel.svelte:148`, `pt.p - prev.p`) while the bars come from `pt.ch`, filtered at `|v| ≥ 0.03` (`:139`). At a boundary all six `ch` are `0.00`, the filter empties, and the fallback string renders. Actual output from the run above:

```
T70:  header → "T69 → T70: +13 pts"
      ch → cities=0.00 growth=0.00 orders=0.00 science=0.00 eco=0.00 mil=0.00
      bars → (none) → renders "no dimension moved much"
```

"+13 pts" directly above "no dimension moved much", at four predictable turns in every duel.

`momentum-mirror.test.ts` pins `Σlv = log-odds` but has no invariant for `Σch ≈ Δlog-odds`, which is why this passes. The bake header already concedes a residual ("cuts the Σch−Δlog-odds residual p95 ~13%") — worth pinning that relationship in the test so the boundary case can't hide inside it.

## 3. `cities` costs far more than it earns — and dropping it takes the razed-city bug with it

The fitted weights are `-0.0391, 0.1717, 0.1286, -0.0094, -0.1597`. Sign-flipping, near-zero, and negative in the endgame. The bake comment keeps it as "independently interpretable" while conceding r ≈ +0.65 with growth — but a coefficient that flips sign across buckets is acting as a suppressor for growth, not measuring anything. In the panel a negative contribution renders under the *"against {leader}"* heading, so a late-game leader with more cities gets a bar telling the reader their city lead is working against them.

That's the visible half. The structural half is what that dimension costs:

- `momentumCurve` returns **null** if `citySeries()` returns null — no `map_tiles` or no `tile_ownership_history` means no curve at all for that game (`momentum.ts:221-222`)
- `featsAt` returns null when either side lacks a city count at turn T, dropping that turn from the curve (`:154-156`)
- `citySeries` is the sole consumer of both `mapTiles` and `tileOwnership` in the scorer
- it's why `rawMapTiles` is threaded as a new prop through `GameDetailView` → `OverviewTab`
- it's why the bake carries the Gotcha-5 tile-index validation at all

And it has a silent correctness bug: `citySeries` only walks tiles that are `is_city_center` **in the end state** (`momentum.ts:123`). A city razed mid-game never existed as far as the curve is concerned — the player who held it is undercounted for the entire period they held it, and losing it produces no momentum swing at all. In a duel corpus that's the event the chart most needs to show.

So the whole most-fragile input path, plus a hard gate on the entire feature, serves the dimension with the smallest weights. Dropping `cities` deletes the reconstruction, the new prop, the bake validation, and the razed-city bug in one move. If you'd rather keep it for interpretability, it at least needs a non-negativity constraint and the end-state bug fixed.

## 4. The AUC cited for "honest uncertainty" is in-sample

The generated file says it plainly: `// In-sample AUC at 30/50/70% of game: 0.706 / 0.768 / 0.822`. The description promotes these to "Discrimination on the corpus … genuinely uncertain early, quite sure late, which is what you'd want from an honest 'who's winning' line." Citing in-sample numbers as evidence the model knows its own uncertainty is the one place the framing outruns the evidence.

Cross-validation should group **by game**, not by row: every turn of a game is a training row carrying the same label, so the fit sees tens of thousands of heavily autocorrelated rows backed by 384 independent matches, and `L2 = 2.0` is calibrated against the inflated count. (`aucAt` already samples one point per game — it's only the fit that doesn't group.)

Related, and more important for what ships: this renders a **probability** as a percentage, and nothing in the bake measures calibration — no Brier score, no log loss, no reliability curve. AUC is rank-only; a model can discriminate at 0.82 and still be systematically overconfident. Calibration is the only property the UI actually claims. A reliability check would sit naturally alongside the three existing hard-fail validations.

This matters most at the front of the curve. At 30% progress the model is barely better than a coin flip, but the chart renders that stretch with exactly the same visual authority as the confident endgame, and a filled win-probability curve is the visual idiom of live win-probability graphics — which are forecasts. The code header is scrupulous about "who was winning, not who would have won"; nothing in the UI is.

## 5. Generate the Worker mirror instead of hand-copying it

Of the four momentum files, two are generated and safe — `bake-momentum.ts` writes both `src/lib/generated/momentum.ts` and `cloud/src/generated/momentum.ts` from one `OUTPUTS` array in a single run, so the weights table can't drift. That's the established dual-emit pattern (law-classes, nation-names, wonders) and it's the right call.

The exposure is the other pair. `cloud/src/momentum.ts` is a hand-made copy of the 280-line scorer — nothing in `scripts/` or `package.json` emits it, and the files say so: *"change one, copy to the other."* This is the repo's first mirror duplicating executable arithmetic rather than data.

The drift test is weaker than it reads:

```ts
expect(cloudCurve.toString()).toBe(frontCurve.toString());
```

`Function.prototype.toString()` returns the source of `momentumCurve` only. Every module-level helper — `yieldSeries`, `powerSeries`, `citySeries`, `featsAt`, `sdAt`, `zOf`, `weightsAt`, and the `ECO5` constant — is outside that string, covered only by the one 31-turn synthetic fixture. That fixture supplies just `YIELD_MONEY` of the five eco resources, so the other four tie at 0 on both sides: **drop `YIELD_IRON` from one copy's `ECO5` and every assertion still passes.** Same for `citySeries` (two centres, ownership set once at turn 2, never changed) and the `sdAt` clamp (`finalTurn: 31` never reaches the table edge).

The fix is cheap, because the entire intentional delta between the two files is two lines:

```
9a10
> // THIS IS THE MIRROR — the source of truth is src/lib/game-detail/momentum.ts.
21c22
< } from "../generated/momentum";
---
> } from "./generated/momentum";
```

So: read the frontend scorer, insert the header line, rewrite one import path, write `cloud/src/momentum.ts`, stamp it `// AUTO-GENERATED … Do not edit.` Then replace the drift test with a regeneration check — re-run the transform, assert the on-disk file matches. That gives the scorer the same guarantee the weights table already has, with no architectural change.

(The alternative — deleting the mirror and importing across the package boundary — does work; I confirmed esbuild bundles `cloud/src/… → ../../src/lib/game-detail/momentum` cleanly and `tsc -p cloud/tsconfig.json` passes, and three cloud test files already import across. But no cloud *runtime* file does, and it would undo the dual-emit convention for one of four bakes. That's a repo-wide decision, not something to settle inside this PR.)

With the mirror generated, the behavioural half of `momentum-mirror.test.ts` becomes redundant. Keep the model invariants there — antisymmetry, `Σlv = log-odds`, unchanged-stat-contributes-zero — and that's the natural home for the `Σch ≈ Δlog-odds` assertion from #2.

---

**Minor, but nearly free now and unrecoverable later:** migration 0043 adds a bare `momentum REAL`. The stated invariant is that a finished game's curve must never drift between visits — but `MOMENTUM_MODEL_VERSION` bumps only on *form* changes (a refit deliberately keeps it), and the admin reindex sweep rewrites every row. After the first refit the column silently mixes vintages with no way to tell which is which. A `momentum_version` column costs one line now; the provenance can't be reconstructed later.

alcaras:
Thanks for the rebase and for the depth here — I verified all five against `5cb800e` and agree with all five. Plan below; fixes will land on this branch.

**1 — bucket cliffs.** Agreed. I'll interpolate the weight vector piecewise-linearly between bucket centres (clamped flat before the first centre and after the last), which keeps the front-load/back-load shape and kills the four jumps. That's a form change, so `MOMENTUM_MODEL_VERSION` → 2 along with the refit below.

**2 — the decomposition.** With interpolated weights the per-turn weight delta becomes small and continuous, and the "phantom changes" worry that motivated holding the weight fixed at the later turn mostly evaporates with it. So rather than patching the fixed-weight form, I'll make `ch` exact — difference `lv` directly, so `Σch = Δlog-odds` identically — and pin that identity in the test as you suggest. The unchanged-stat-contributes-zero invariant weakens to "≈ 0 within per-turn normaliser drift", which the 0.03 display filter already absorbs. Header and bars then can't disagree by construction.

**3 — cities.** Dropping it. Your cost/benefit is right, and the razed-city bug settles it — the event the dimension most needs to show is the one it structurally can't see. Refit as five dims; `citySeries`, the `rawMapTiles` prop threading through `GameDetailView` → `OverviewTab`, and the Gotcha-5 bake validation all go with it (as does the cities row in the key-stats table, since the reconstruction is what fed it). The curve then no longer hard-gates on `tile_ownership_history`, which also widens which games get one.

**4 — in-sample AUC and calibration.** Agreed on all three parts. The refit will cross-validate grouped by game and pick L2 against that rather than the inflated row count, and the generated header will cite the out-of-sample numbers. I'll add a reliability check (Brier + a coarse reliability curve) as a fourth hard-fail validation, since as you say calibration is the only property the UI actually claims. On the UI itself: the "retrospective, not a forecast" framing living only in a code header is the wrong place — I'll surface it in the panel and de-emphasise the low-confidence early stretch rather than render 30% progress with endgame authority. Exact treatment I'll experiment with and you can judge on the result.

**5 — generated mirror.** Doing it exactly as you describe: the bake reads the frontend scorer, inserts the mirror header, rewrites the one import, stamps it generated; the drift test becomes a regeneration check. The behavioural invariants — antisymmetry, `Σlv = log-odds`, and the new `Σch = Δlog-odds` — stay in `momentum-mirror.test.ts`. Agreed the cross-import question is out of scope here.

**Migration.** Adding `momentum_version INTEGER` to 0043 while it's still unmerged, written at derive time next to the score. You're right that the provenance can't be reconstructed after the first refit.

Sequencing: the mirror generation and the migration column are independent of the model and will come first; 1–4 all touch the fit, so they'll land together as one refit commit (needs a local corpus run) plus the panel changes.

alcaras:
All five landed, in the order promised — four commits on top of your `5cb800e`, PR description updated to match (including deploy step 1 → 0043).

- `90e6063` — **#5**: the bake now generates `cloud/src/momentum.ts` through `scripts/momentum-mirror.ts` (`--mirror-only` regenerates without a corpus); the drift test is a byte-for-byte regeneration check.
- `31c3695` — **minor**: `momentum_version INTEGER` added to 0043, written at derive time.
- `f002242` — **#1–#4** as one refit, `MOMENTUM_MODEL_VERSION` → 2:
  - weights interpolate piecewise-linearly between bucket centres (your constant-lead synthetic is now a regression test: no single turn may move p by ≥3 points);
  - `ch` is the exact difference of `lv`, so `Σch = Δlog-odds` identically — pinned in the test next to `Σlv`;
  - cities is gone, and with it `citySeries`, the `rawMapTiles` threading, the Gotcha-5 validation, and the hard gate on `tile_ownership_history` (more games get a curve);
  - L2 by 5-fold CV grouped by game (flat across the grid, 8 wins by a hair — consistent with the earlier experiments), header cites pooled held-out numbers, and a calibration hard-fail joins the suite.
- `9e8a20f` — **#4's UI half**: the retrospective note is a caption under the chart, and the line/fills fade in over the first ~30% of the game instead of rendering the coin-flip stretch with endgame authority.

The corpus had grown to 471 deduped duels since the v1 fit, so the numbers moved for two reasons at once. Held-out (grouped) AUC@30/50/70%: **0.668 / 0.755 / 0.812** — a touch under the old in-sample 0.706/0.768/0.822, which is the honest gap you predicted. Calibration: Brier **0.198** (bound < 0.25), 10-bin ECE **0.034** (bound ≤ 0.08).

One judgement call to flag on #2: rather than adding the bucket-jump term to the fixed-weight form, `ch` now differences `lv` directly. With interpolated weights the per-turn weight/scale drift is small and continuous, so the "phantom changes" the fixed-weight form existed to avoid stay below the 0.03 display filter — and in exchange the header/bars identity is exact rather than approximate. The tied-stat-contributes-zero invariant still holds exactly and is still tested.

Fresh v2 screenshot is in the description. Suite state unchanged: everything green except the four pre-existing `canonical-map-options` failures.

becked:
The momentum panel has a second side now: an ⓘ in the card's top right flips it — the same crossfade as this repo's bracket/standings switch — and the back documents each part of the panel. The front prints log-odds contributions, a standardised-lead split and inferred battles, none of which announce their own arithmetic.

@alcaras — could you review the text? You know the model far better than the panel does, and I'd rather a wrong or overstated sentence get caught before it becomes the explanation everyone reads. The two I'd most like a second opinion on are the battle thresholds and the "not a forecast" framing. (*Player A* / *Player B* are the duel's two players; in the app they're the real names, in their colours.)

---

**The curve** — One point per turn: the model's probability that *Player A* was ahead, with the gap to the 50% midline filled in whoever leads — *Player A* above it, *Player B* below. Hover a turn to read it out below the chart, click to pin it. It is a retrospective reading of a finished game, not a forecast: a turn is weighted by how far through the match it falls, which needs the final turn to be known.

**What is scored** — Five differences between the two players at each turn: growth, orders and science (yield rates); eco — which side has the higher rate for money, food, iron, stone and wood, ±1 apiece; and military, the power gap relative to the two sides' average power, since absolute power grows many times over across a match. Each difference is divided by how far apart duellists typically are at that turn, multiplied by a fitted weight and summed into log-odds, which the logistic curve turns into the percentage. The weights are fitted separately for successive stages of a game and interpolated between them, so the same lead is worth different amounts at T20 and at T80.

**At T… — the numbers** — Each side's own figures at the shown turn — growth, orders and science yield rates, and military power — with the higher of the pair in that player's colour. The model reads only the difference between the two; these are here raw so the arithmetic can be checked against the game.

**Behind the lead** — The current log-odds split by dimension — weight × standardised lead, one bar each — signed so that a positive bar helps the leader and the bars add up to the whole lead. Dimensions pulling the other way are listed separately, under a heading naming the leader they work against. A dimension worth less than 0.03 is dropped as noise.

**The swing** — The same split applied to the move since the previous scored turn: the exact difference between the two turns' contributions, so the bars always add up to the heading's swing, and positive is toward whoever gained. A tied dimension contributes nothing; a steady lead still drifts a little as the weights shift with game progress.

**Events** — Battles are not in the save's event log, so they are read from military power across the window: both sides losing more than 25 power is a trade, named for whoever lost less; one side alone losing more than 60 is an army destroyed. The rest are logged events — cities, wonders, laws, techs, religions, successions, wars — for the turns since the previous scored point. A turn missing either side's power, orders or science is not scored, so that window can span several turns. They are context, not causes: nothing here is attributed to the swing above.

CleanShot 2026-08-30 at 21 51 38@2x

becked:
Five commits landed after [the "all five landed" comment](https://github.com/becked/per-ankh/pull/175#issuecomment-5432332847) without one of their own — a presentation pass over the panel, plus the docs note. No model, migration or Worker code moved in any of them: the only `cloud/` file touched is a `CLAUDE.md`, so the v2 numbers, migration 0043 and the derive-time write all stand as described.

- **`e7be0c8`** — *docs*. Notes in `cloud/src/CLAUDE.md` and `src/lib/game-detail/CLAUDE.md` so a future session knows `cloud/src/momentum.ts` is generated (edit the frontend scorer, regenerate with `--mirror-only`, never hand-edit) and that a refit is an explicit act outside `bake:all`, versioned by `MOMENTUM_MODEL_VERSION` and the per-row `momentum_version`.

- **`c748e6c`** — *drops the early-game fade and the honesty caption*, and closes a header/bars contradiction. This reverses the UI half of `83102dc` (announced here as `9e8a20f`), so it deserves the reasoning: the ramp was keyed to x position rather than to anything the model measures, so it dimmed the opening of every game by the same amount whether the curve was pinned to 50% or already falling — and alpha reads as "no data here", not "less certain", which made a thin early band look like the fill simply started partway through the game. Where the model doesn't know, the curve sits on the 50% midline, and that is legible on its own. The caption explaining the ramp went with it; the panel's contract is that it carries no commentary.
  Separately, the 0.03 display floor is per-dimension while the header quotes their sum, so five dimensions each drifting just under it still add to a move the header prints — a flat-lead 100-turn duel rendered "+1 pts" over "no dimension moved much" on four separate turns. `bars()` takes its floor as a parameter now and lowers it when the header has points to explain; a turn that genuinely didn't move still gets the fallback.

- **`0ee8a7d`** — *a solid 50% midline*. It specified no width, so it inherited the series' 2px and drew as heavily as the data line. Solid at 1px carries less ink while staying continuous. Both momentum surfaces move together — the panel and the feed sparkline in `RecentSaveCard.svelte`.

- **`45ead09`** — *the hover marks the turn, and nothing else*. The axis pointer arrived with three marks the panel never wanted: a dashed vertical line, ECharts' blue turn label, and a second dot where the clamped fill series cross the midline. `type: "none"` drops the pointer line while keeping the component live — it still snaps and still emits `updateAxisPointer`, which is what drives the detail panel, because ECharts builds the label outside the pointer-graphic branch. The midline dot needs `symbol: "none"` on the two fill series rather than `emphasis: { disabled: true }`: `LineView.highlight` builds its temporary symbol without consulting that flag, but a "none" symbol draws no path. The turn label takes the app's orange as `CHART_ACCENT_COLOR` in `config/charts.ts`, beside the reference-line colour rather than inlined — hex because canvas can't resolve `rgb(var(--color-orange))`.

- **`2a8f83a`** — *the events heading names its own turns*. "Logged this window" named the scoring window, machinery the panel never puts on screen. It now reads the same `eventsFrom` bound the filter does — "At T39" when the previous scored point is adjacent, "At T37–T39" when a gap in the scored series widens it — so the heading can't disagree with the rows underneath it, and "At" matches the stats column. It stays source-neutral on purpose: battles in that list are derived from power drops, not read from the log. Also drops the "click to pin" caption; the interaction is unchanged, only the label goes.

(The sixth, `ea038fe`, is the info side in the comment above.)

**Corrections to the PR text.** The description described the fade and the caption as shipped v2 behaviour in three places — the v2 screenshot caption, the "question it answers" paragraph, and the testing bullet. Those are fixed now, and the Overview bullet mentions the info side. The v2 overview screenshot still shows the fade and the caption: it predates this pass, and its caption now says so.

Issue #152, opened 2026-07-20
Investigate: per-player difficulty science and EFFECTCITY_BASE — two unmodelled flat sources (~20% of Other), blocked on a difficulty gating question


becked:
Two flat science sources that the Techs breakdown (`src/lib/game-detail/science-techs.ts`) models nowhere, and that together measure **~20% of the signed `Other` remainder** — larger than #148 and #131 individually. Both are exact and, unusually, need **no parser change, no `PARSER_VERSION` bump, and no Worker deploy**: the data is already on the blob.

**They are not ready to implement.** The difficulty half fails an arithmetic sanity check against real saves, and the root cause is a consumption-site question in the game source that this issue exists to answer. Filing as an investigation, not a build ticket.

### 1. Per-player difficulty science — measured **392 (11.9% of `|other|`)**

`Player.initPlayerEffects` applies the player's difficulty effect unconditionally:

```csharp
// Player.cs:15772-15790
changeEffectPlayerCount(infos().Globals.BASE_EFFECTPLAYER, 1);
// … game options …
changeEffectPlayerCount(difficulty().meEffectPlayer, 1);   // :15784 — no gate

if (!isHumanOriginal())                                     // :15786 — the AI-only one
{
    changeEffectPlayerCount(game().advantage().meEffectPlayer, 1);
}
```

The absence of a human/AI gate on `:15784` is conspicuous precisely because `advantage()` two lines later has one. `difficulty.xml` maps each tier to an `EFFECTPLAYER_DIFFICULTY_*`, which pays a flat `aiYieldRate YIELD_SCIENCE`:

| Difficulty | raw | display |
|---|---|---|
| ABLE / GOOD | 120 | +12 |
| STRONG | 110 | +11 |
| NOBLE | 100 | +10 |
| GLORIOUS | 90 | +9 |
| MAGNIFICENT | 80 | +8 |
| GREAT | 70 | +7 |

Note `difficulty.xml` itself carries **no** science field — only `aiYieldStockpile` (a one-off) and rebel/raid probabilities. That is why searching for a difficulty *modifier* comes up empty: it is a flat rate, reached indirectly through `EffectPlayer`.

Already on the blob: `players[].difficulty` (`parsers/players.ts:81`, from `<Difficulty><PlayerDifficulty>` at `:144-147`).

### 2. `EFFECTCITY_BASE` — measured **277 (8.4% of `|other|`)**

Every city gets it at founding:

```csharp
// City.cs:7774, in City.init
[...]

Issue #148, opened 2026-07-19
Itemize city project science (Archives, Philosophy Forums) in the Techs breakdown


becked:
> **Re-reviewed 2026-07-20.** Independently reproduced: Archives **237**, Forums **41**, total **278** — exact. Non-stacking is confirmed and is in fact *stronger* than stated below. But the scope is too narrow: **12 other projects pay science**, and a second conditional grant hangs off the very same Archive hook this issue wires up. See [Sources this issue does not yet cover](#sources-this-issue-does-not-yet-cover).

## Background

The Techs tab's science breakdown (`src/lib/game-detail/science-techs.ts`) itemizes specialists, buildings/resources, laws, percent modifiers, and the leader's court science. **City projects are itemized nowhere**, so all Archive and Philosophy-Forum science falls into the signed `Other` remainder — even though the save carries the data and a parser for it already exists.

## What the save has

Every `<City>` carries a `<ProjectCount>` block:

```xml
<ProjectCount>
  <PROJECT_FORUM_1>1</PROJECT_FORUM_1>
  <PROJECT_AUSTERITY>1</PROJECT_AUSTERITY>
  <PROJECT_PROTECTION_EDICTS>1</PROJECT_PROTECTION_EDICTS>
  ...
</ProjectCount>
```

`parseCityProjectCounts` already exists in `src/lib/parser/parsers/cities.ts` — it is **never called**: not imported by `parsers/index.ts`, and there is no field for it on `FullGameData`. (It's one of several orphaned parsers; `parseCityProjectsCompleted` and `parseCityProductionQueue` are in the same state.)

## The science

**Archives** pay flat science through `EFFECTCITY_PROJECT_ARCHIVE_1..4` `aiYieldRate`:

| Project | raw | display |
|---|---|---|
| `PROJECT_ARCHIVE_1` | 10 | +1 |
| `PROJECT_ARCHIVE_2` | 20 | +2 |
| `PROJECT_ARCHIVE_3` | 40 | +4 |
| `PROJECT_ARCHIVE_4` | 80 | +8 |

**The tiers do not stack.** Each has `iMaxCount 1` and `abInvalidBy` listing every higher tier (ARCHIVE_1 lists 2/3/4; ARCHIVE_4 has none). The itemization must take the **highest tier present per city**, not the sum — summing would over-count by 7 per fully-upgraded city and drive `Other` negative.

[...]

Issue #147, opened 2026-07-19
Itemize governor science in the Techs breakdown (largest measured slice of Other)


becked:
> **Re-reviewed 2026-07-20.** An independent rebuild of the harness reproduces the headline measurement (**528.4** vs 528.6 claimed — within rounding). Three corrections applied below: the "In-game verification" card is the **Charisma→Civics** column, not science; the `miRateModifier` consumer count was wrong; and a scenario-mod override plus a join gotcha are now recorded. The core claim and the measurement both stand.

> **Corrected 2026-07-19.** The original text claimed "governors are opinion-free". That is too broad — `City.calculateGovernorOpinionYield` exists and governor opinion does pay a yield. The claim that actually holds, and that this issue depends on, is narrower: **the governor rating→yield percentage modifier is opinion-free.** That has since been confirmed against live game output (see [In-game verification](#in-game-verification)). The measurement below is unaffected.

## Background

The Techs tab's science breakdown (`src/lib/game-detail/science-techs.ts`) itemizes specialists, buildings/resources, laws, percent modifiers, and the leader's court science, and dumps everything else into a signed `Other` remainder. **Governors are the largest measured component of that remainder, and the governor percentage modifier is exactly computable from the save without porting the opinion system.**

## Why the governor modifier is not blocked on opinion

`InfoHelpers.getRatingYieldModifierGovernor` (`InfoHelpers.cs:1185`) takes **no `OpinionCharacterType` parameter**:

```csharp
public virtual int getRatingYieldModifierGovernor(RatingType eRating, YieldType eYield, int iValue, Game pGame)
{
    return boostRating(mInfos.rating(eRating).maiYieldGovernorModifier[eYield], iValue, pGame);
}
```

Every sibling in the same family does take one — `getRatingYieldRateCourt` (:1238), `getRatingYieldRateCouncil` (:1285), `getRatingYieldAgentPercent` (:1189). Opinion's `miRateModifier` has **nine** non-display consumers in GameCore — eight in `InfoHelpers.cs` (`:1198`, `:1256`, `:1295`, `:1315`, `:1326`, `:1337`, `:1356`, `:1368`) plus `PlayerAI.cs:6463`, where the AI hoists the modifier out of the info record and applies it to `getRatingYieldRateAgent` itself. The governor modifier is on none of them.

[...]

PR #176, opened 2026-07-28
feat: Techs tab — full science-source itemization + leader/knowledge rail markers

alcaras:
## What

Extends the Techs tab's **Science Sources** breakdown and the science graph's annotation rail — every number anchored on the game's own formulas (Reference/XML + game source, cited at each site).

**New breakdown sections/rows:**

- **Nation, family & religion** — the flat per-city EffectCity sources the save now lets us price:
  - **Sages cities**: +1 science per placed specialist in the family's cities (`EFFECTCITY_FAMILYCLASS_SAGES aiYieldRateSpecialist` — any specialist, unlike Constitution's urban-only rate).
  - **Babylonia**: +1/city from the nation's player effect.
  - **Dualism**: in each city holding a religion that established the theology, its rate × ALL religions present, once per holding religion (`City.cs resetReligionCityEffect` × `getReligionCount()`).
  - **Science city projects**: Archive I–IV (+1/+2/+4/+8, `bSingle` so they pay once) and Convoys (repeatable).
- **Governors** (in Modifiers): percent city science off the governor's Wisdom — `boostRating` over `aiYieldGovernorModifier` (+2%/boosted point, plain `triangleBoost`, no offset), with the Competitive Mode linearization, estimated per city against the same reconstructed base as the Library rows. Gated on a known Competitive flag, same rationale as the court section.
- **Discontent** (in Modifiers, negative): −5% science per negative happiness level per city (`yield.xml iNegativeHappinessModifier` × |level|, `getHappinessLevelYieldModifier`). This was the big unpriced negative — on the test game below it moves the signed Other remainder from **−49.9 to +12.3**, making it a genuine floor again.

Since these are all EffectCity **base** yields, they also feed the per-city bases the percent-modifier estimates multiply, matching `City.cs calculateBaseYield`.

**New rail rows (two-player games):**

- **Leader changes** (Wisdom icon): who died/abdicated and both rulers' Wisdom — exact for the outgoing ruler, since ratings freeze at death.
- **Knowledge shifts** (text chips, "N→C"): the game's Primitive/Naive/Competent/Learned/Erudite classification from `Player.calculateKnowledgeOf` — cumulative science as an integer percent of the opponent's, bucketed by knowledge.xml's thresholds (67/83/120/150, Erudite the catch-all per `InfoPercentBase`'s int.MaxValue default). Debounced: a tier must hold 3 consecutive turns (threshold-hovering flip-flops the raw classification every turn); the final run always stands.

## Screenshots

**Science Sources** with the new sections — Sages, Babylonia, Archives, Governors, Discontent:

sections

**Dualism** on a two-religion theology game (each city holding both counts twice):

dualism

**The rail** — leader changes (Wisdom icons) and knowledge shift chips (`C→N` … `N→C`):

rail

## Parser 2.14.0

Four additive fields: `city_statistics.cities[]` gains `religions` (reusing the presence list `parseCityReligions` already reads for the map), `project_counts`, `governor_xml_id`, and `happiness_level` (signed; current saves write `TeamHappinessLevel`, older ones `TeamDiscontentLevel` which the game loads negated — `City.cs:1748` — normalized in the parser); `game_religions[]` gains `theologies` (the `<ReligionTheology>` dotted presence tags). Older blobs lack the fields and the new rows are simply omitted.

**Version note:** this PR numbers the bump 2.14.0 and deliberately skips 2.13.0, which is reserved by the open projects PR (#172). If this lands first, #172 should renumber to 2.15.0 on rebase (a lower-than-current version would break its re-imports); if #172 lands first, this rebases cleanly on top.

**Deploy order:** Worker before frontend (as always for a parser bump). Existing games pick the new rows up on re-upload — reindex can't backfill parser-derived fields.

## Testing

- Verified against re-imported 2.14.0 games: section totals reconcile exactly to the final science rate; Dualism's establishment turns match +15–20 jumps in the rate series; the Halab-style multi-religion city prices out exactly (holders × religions present); Discontent turns the previously-negative Other positive on both test games.
- `npm run lint` / `svelte-check` clean; `cloud/` unit suite green except the 4 pre-existing `canonical-map-options` failures on `main`.
- The bake's new tables are generic (whole familyClass/nation/theology/project sweeps), so a future mod or patch that adds another science-carrying entry appears without code changes.

### DIFF
--- a/cloud/src/schemas/game.ts
+++ b/cloud/src/schemas/game.ts
@@ -108,12 +108,13 @@ export const MAX_DISABLED_IMPROVEMENTS = 1_000;
 //         player who owned the wonder's tile on the turn it completed, read
 //         from the ownership history, rather than whoever holds the tile at
 //         the end. Blobs below 2.12.0 credit a captured wonder to its captor.
-// 2.14.0: adds per-city religion presence, project counts, and governor
-//         xml_id on city_statistics.cities, plus theologies on
-//         game_religions — the Techs tab's science-source breakdown reads
-//         all four. Purely additive; older blobs lack the fields and the
-//         new breakdown rows are omitted. (2.13.0 is reserved by the open
-//         projects_produced PR; this PR deliberately skips over it.)
+// 2.14.0: adds per-city religion presence, project counts, governor
+//         xml_id, and happiness_level on city_statistics.cities, plus
+//         theologies on game_religions — the Techs tab's science-source
+//         breakdown reads all five. Purely additive; older blobs lack the
+//         fields and the new breakdown rows are omitted. (2.13.0 is
+//         reserved by the open projects_produced PR; this PR deliberately
+//         skips over it.)
 export const KNOWN_PARSER_VERSIONS = new Set([
 	"2.0.0",
 	"2.1.0",
--- a/cloud/src/schemas/game.ts
+++ b/cloud/src/schemas/game.ts
@@ -108,6 +108,12 @@ export const MAX_DISABLED_IMPROVEMENTS = 1_000;
 //         player who owned the wonder's tile on the turn it completed, read
 //         from the ownership history, rather than whoever holds the tile at
 //         the end. Blobs below 2.12.0 credit a captured wonder to its captor.
+// 2.14.0: adds per-city religion presence, project counts, and governor
+//         xml_id on city_statistics.cities, plus theologies on
+//         game_religions — the Techs tab's science-source breakdown reads
+//         all four. Purely additive; older blobs lack the fields and the
+//         new breakdown rows are omitted. (2.13.0 is reserved by the open
+//         projects_produced PR; this PR deliberately skips over it.)
 export const KNOWN_PARSER_VERSIONS = new Set([
 	"2.0.0",
 	"2.1.0",
@@ -127,13 +133,14 @@ export const KNOWN_PARSER_VERSIONS = new Set([
 	"2.10.0",
 	"2.11.0",
 	"2.12.0",
+	"2.14.0",
 ]);
 
 // The latest accepted version. Echoed back on stats responses and
 // embedded in stats cache keys so a parser bump (after the matching
 // extraction code lands) naturally orphans every old entry. Bump in
 // lockstep with the `KNOWN_PARSER_VERSIONS` addition above.
-export const CURRENT_PARSER_VERSION = "2.12.0";
+export const CURRENT_PARSER_VERSION = "2.14.0";
 
 // ----- Reusable atoms -----
 
--- a/scripts/bake-science-yields.ts
+++ b/scripts/bake-science-yields.ts
@@ -79,6 +79,7 @@ interface Entry {
 	iValue?: string;
 	iPercent?: string;
 	iTriangleOffset?: string;
+	iNegativeHappinessModifier?: string;
 	aiYieldOutput?: { Pair?: YieldPair | YieldPair[] };
 	aiYieldRate?: { Pair?: YieldPair | YieldPair[] };
 	aiYieldModifier?: { Pair?: YieldPair | YieldPair[] };
@@ -194,6 +195,18 @@ async function main(): Promise<void> {
 		findEntry(yields, "YIELD_SCIENCE", "yield.xml").iTriangleOffset,
 		"YIELD_SCIENCE iTriangleOffset",
 	);
+	// Percent science per city discontent level (yield.xml
+	// <iNegativeHappinessModifier>, applied ×|happiness level| when the level
+	// is negative — InfoHelpers.getHappinessLevelYieldModifier). Negative.
+	const scienceDiscontentModifier = requireInt(
+		findEntry(yields, "YIELD_SCIENCE", "yield.xml").iNegativeHappinessModifier,
+		"YIELD_SCIENCE iNegativeHappinessModifier",
+	);
+	if (scienceDiscontentModifier >= 0) {
+		throw new Error(
+			"bake-science-yields: YIELD_SCIENCE iNegativeHappinessModifier is not negative",
+		);
+	}
 	const competitiveEquivalentRating = requireInt(
 		findEntry(
 			globalInts,
@@ -648,6 +661,17 @@ async function main(): Promise<void> {
 	lines.push("// around COMPETITIVE_EQUIVALENT_RATING — to get the city's %.");
 	lines.push(
 		`export const WISDOM_GOVERNOR_SCIENCE_MODIFIER = ${wisdomGovernorScienceModifier};`,
+	);
+	lines.push("");
+	lines.push(
+		"// Percent city science PER DISCONTENT LEVEL (yield.xml",
+	);
+	lines.push(
+		"// iNegativeHappinessModifier ×|level| — getHappinessLevelYieldModifier).",
+	);
+	lines.push("// Negative: an unhappy city genuinely earns less.");
+	lines.push(
+		`export const SCIENCE_DISCONTENT_MODIFIER = ${scienceDiscontentModifier};`,
 	);
 	lines.push("");
 	lines.push(
--- a/scripts/bake-science-yields.ts
+++ b/scripts/bake-science-yields.ts
@@ -70,17 +70,22 @@ interface Entry {
 	AssetVariation?: string;
 	EffectCity?: string;
 	EffectCityExtra?: string;
+	EffectPlayer?: string;
 	TechPrereq?: string;
 	LawClass?: string;
 	Specialist?: string;
 	zIconName?: string;
 	iCost?: string;
 	iValue?: string;
+	iPercent?: string;
 	iTriangleOffset?: string;
 	aiYieldOutput?: { Pair?: YieldPair | YieldPair[] };
 	aiYieldRate?: { Pair?: YieldPair | YieldPair[] };
 	aiYieldModifier?: { Pair?: YieldPair | YieldPair[] };
 	aiYieldCourtRate?: { Pair?: YieldPair | YieldPair[] };
+	aiYieldGovernorModifier?: { Pair?: YieldPair | YieldPair[] };
+	aiYieldRateSpecialist?: { Pair?: YieldPair | YieldPair[] };
+	aiYieldRateReligion?: { Pair?: YieldPair | YieldPair[] };
 	aiImprovementClassModifier?: { Pair?: YieldPair | YieldPair[] };
 	aaiResourceYieldOutput?: { Pair?: ResourceYieldPair | ResourceYieldPair[] };
 }
@@ -126,6 +131,11 @@ async function main(): Promise<void> {
 		laws,
 		lawClasses,
 		techs,
+		familyClasses,
+		nations,
+		theologies,
+		projects,
+		knowledges,
 	] = await Promise.all([
 		loadEntries(resolve(infosDir, "improvement.xml")),
 		loadEntries(resolve(infosDir, "improvementClass.xml")),
@@ -138,6 +148,11 @@ async function main(): Promise<void> {
 		loadEntries(resolve(infosDir, "law.xml")),
 		loadEntries(resolve(infosDir, "lawClass.xml")),
 		loadEntries(resolve(infosDir, "tech.xml")),
+		loadEntries(resolve(infosDir, "familyClass.xml")),
+		loadEntries(resolve(infosDir, "nation.xml")),
+		loadEntries(resolve(infosDir, "theology.xml")),
+		loadEntries(resolve(infosDir, "project.xml")),
+		loadEntries(resolve(infosDir, "knowledge.xml")),
 	]);
 
 	const effectByType = new Map(effects.map((e) => [e.zType, e]));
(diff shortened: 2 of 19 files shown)


alcaras:
Self-review (correctness + fit against CLAUDE.md's contributing checks). One finding was caught post-submit and fixed in the last commit; the rest came up clean.

**Found and fixed**

- **Leader tooltip showed raw \`NAME_*\` enums** and a bare verb. It now renders full ruler lines — archetype glyph, display name with regnal numeral, "died, aged N" with the Wisdom glyph, cause of death, reign length, and the successor line — using the same formatting rules as the Leaders tab. The name/death-reason formatting was living as LeaderCard-local helpers, so it moved to `$lib/utils/formatting` (`rulerDisplayName`, `deathReasonLabel`) and LeaderCard now delegates to the shared copies rather than keeping a second implementation. (Multiplayer saves genuinely name dynasty characters "Hittite Male20" — that matches the Leaders tab; single-player games show real names.)

**Checked, no finding**

- *Reuse*: city religion presence reuses the existing `parseCityReligions` parse (the one the map tiles consume) instead of re-reading the tag; the discontent/governor estimates ride the existing per-city modifier machinery; archetype icons/name/death formatting share the Leaders tab's helpers; the roman-numeral tier label matches the game's own "Archive II" naming.
- *Null handling*: `??`/`!= null` throughout the new scoring code; every `||` in the diff is a boolean guard, not a data fallback.
- *Colors/literals*: no new hex literals — tooltip shades come from the shared `TOOLTIP_*` exports, chips use the player color, icons resolve through `getSpritePath`.
- *Parallel surfaces*: the `RailMarker.label` extension is optional, so the Military rail is untouched; `EventRail` renders label > sprite > dot with the existing fallback chain.
- *Domain vocabulary*: `happiness_level` (signed) matches the game's current concept — the save's legacy `TeamDiscontentLevel` tag is normalized exactly the way `City.cs:1748` loads it; knowledge tiers, theology names, and project labels all come from Reference/XML.
- *Generated files*: all new constants are baked (`bake:science-yields` extended), no hand edits; the tables are whole-category sweeps so mods/patches flow through without code changes.
- *Bounds/versioning*: parser fields are blob-additive; migration-free; the 2.13.0/2.14.0 interplay with #172 is called out in the PR body.

- The rail (leader changes, knowledge chips) renders only for two-player matchups, same as the existing science rail rows; FFA games still get the full breakdown table.
- The remaining `Other` residue is the opinion-scaled court and family-opinion percent modifiers — both need data the save doesn't record per character/family, documented at the `other` field.
- The Cities tab could surface the new per-city fields (governor, discontent, religions) in its table — deliberate non-goal here to keep the PR scoped.

becked:
Read this against the five open science-source issues (#147, #148, #152, #131, #143) and #212, and checked the claims against `Reference/XML`, `Reference/Source`, and the 12 saves in `test-data/saves`. The arithmetic holds up — everything below is either "verified, no action" or a gap that's adjacent to what's already here.

## Verified against the game, no action needed

Every new formula reconciles:

- **Sages** — `EFFECTCITY_FAMILYCLASS_SAGES aiYieldRateSpecialist` 10 → +1, and `City.cs:11913-11926` scales it by `getSpecialistCount()`, not `getUrbanSpecialistCount()`. The "any specialist, unlike Constitution's urban-only rate" claim is exactly right.
- **Dualism** — `resetReligionCityEffect` (`City.cs:3687-3699`) bumps the theology effect once per *present religion holding it*, and `City.cs:12007` pays `iYield × getReligionCount()` per effect count (no `bSingle` on `EFFECTCITY_THEOLOGY_DUALISM`, so the count isn't clamped). `holders × rate × religions.length` is the game's own arithmetic.
- **Discontent** — `InfoHelpers.cs:2664-2667` is `-(level) × miNegativeHappinessModifier`, and `YIELD_SCIENCE iNegativeHappinessModifier` is −5. The "positive levels pay nothing for science" branch is right too (`YIELD_SCIENCE` has no `iPositiveHappinessModifier`).
- **The `TeamHappinessLevel` team resolution** is the subtle one, and it's correct. The tag is an array per team; non-owner teams sit at the `-1` constructor sentinel (`City.cs:338`), and only `getTeamHappinessLevel(getTeam())` is the city's real level (`City.cs:6344-6355`). Resolving it through the owner team the way the culture join does is what keeps this from manufacturing −5% on every city in the game.
- **Summing Archive tiers is safe.** `City.finishProject` (`City.cs:8589-8597`) zeroes the count of every invalidated project, so the save only ever carries the top tier. Confirmed empirically: **0 of 282 cities** across the corpus hold two tiers of either the Archive or Forum line. (The PR body credits `bSingle` for this — `bSingle` is what makes each *tier* pay once; `abInvalidBy` + `finishProject` is what stops tiers stacking. Same result, different mechanism, worth getting into the comment since the two are easy to conflate.)
- **Dead governors aren't a problem.** #147's scope asks for a dead-governor skip mirroring the leader handling. Checked all 12 saves: **0 cities** point `<GovernorID>` at a character with a `DeathTurn` — the game clears the post on death. No filter needed.

## Gaps

**1. Philosophy → Forums is missing** — the other half of #148. `EFFECTCITY_LAW_PHILOSOPHY_EXTRA` pays `EFFECTCITY_PROJECT_FORUM` +1 science per city, while `LAW_PHILOSOPHY` is active. Measured over the corpus: **41 forum-holding cities across 8 Philosophy players = +41 science/turn**, reproducing #148's figure exactly. The `other` doc comment already discloses this stays in the remainder, so it's a scope call rather than a defect — but it's the same conditional-grant shape as Dualism, which *is* implemented.

**2. The Scholar-archetype Archive grant is missing, and it's the cheapest thing here.** `EFFECTCITY_TRAIT_SCHOLAR_ARCHETYPE_ALL` pays **+2** per city holding any Archive — twice Philosophy's per-city value. Measured: **8 archive-cities across 3 scholar-leader players = +16/turn**. It hangs off `EFFECTCITY_PROJECT_ARCHIVE`, the empty marker this PR **already reads and discards as 0 science** (it's each Archive tier's `EffectCityExtra`).

Both of these are the same XML shape — `aaiEffectCityYieldRate`, "effect X pays science when effect Y is present in the city" — and the extended bake never reads that field. That's a whole class of source the sweep can't see, which is worth noting alongside the "generic sweeps, so mods flow through" claim.

**3. The project sweep reads only `project.xml`.** `project-event.xml` and its four DLC siblings are never loaded, so `PROJECT_LOCAL_ASCETIC` (+2) and `PROJECT_NEIGHBORS_FEAST_PERSIA` (+2) can't ever appear in `PROJECT_SCIENCE`. Not hypothetical — 3 and 1 completions in the corpus respectively, which is #148's measured 8.0 for "other flat-science projects". (`familyClass.xml` / `nation.xml` / `theology.xml` / `knowledge.xml` have no DLC variants, so those three sweeps genuinely are complete.) Separately, the 8 percent-science projects (`PROJECT_MIDWIFERY` turns up ×4 in the corpus) are a different unit and belong in the percent pass, not `PROJECT_SCIENCE`.

**4. Discontent isn't the only unpriced negative — two more live in the same function.** `City.calculateTotalYieldModifier` (`City.cs:4667-4671`) is governor + happiness + **damage** + **assimilate**. `YIELD_SCIENCE` carries neither `mbNoDamage` nor `mbNoAssimilate`, so both apply, and both are already in the save as `<Damage>` and `<AssimilateTurns>` on the City node — the same additive-parser-field shape this PR is already using. Assimilate is `max(-turns, -50)`:

```
Aksum p0:  6 of 15 cities assimilating — −28 / −26 / −22 / −12 / −8 % science
Rome  p0:  4 of 16 — −25 / −13 / −10 / −5 %
Kush  p0:  Tushpa at −26 %
16 of 282 cities carry damage (−50 % × damage/HPMax)
```

Not a blocker for this PR, but "Discontent was the big unpriced negative" is only two thirds true, and the remaining third is cheap.

**5. This probably unblocks #152.** #152 is stalled on exactly one thing: subtracting every measured source drives 8 of 39 players' remainders negative, so *some* unmodelled negative has to exist, and it has to be found before difficulty science (11.9% of `|other|`) or `EFFECTCITY_BASE` (8.4%) can ship. Discontent is a strong candidate — **180 of 282 cities** in the corpus sit at negative happiness. In the save #152 singles out as its counterexample (`match_426504735`, Assyria, `Other` = 2.3 against a +8 difficulty grant), all three of Assyria's cities are negative, totalling 11 levels = **−55 percentage-points** of city science.

Would you be up for re-running your harness with governors + discontent priced and reporting what happens to those 8 residuals? If they resolve, #152 clears, and both of its sources need **no parser change at all** — `players[].difficulty` and `cityCount` are already on the blob. `EFFECTCITY_BASE` is also worth more than its face value here: as a `cityFlat` entry it lifts the reconstructed base that every percent estimate in this PR multiplies, so it tightens the governor, Library and Discontent rows rather than just adding +1/city.

## #212

This adds two rows to the rail #212 argues is asserting a source the save never records, without touching `scienceSpikes`. Two connections worth recording:

- `PROJECT_INQUIRY` appears **8 times** across the corpus — the exact source #212 argues those spikes are being misattributed away from.
- #212's recommended option B is a baked `EVENTSTORY_* → can this grant science?` table, and `bake-science-yields.ts` — which this PR is already extending, and which already reads `bonus.xml` — is where it goes. Whichever of these lands second inherits a smaller diff if the bake shape is agreed now.

Also note the new per-city `project_counts` is a better source for #212's option-C caption line than the player-level `projects_produced` it currently proposes.

## Mechanical

- **`PARSER_VERSION` 2.14.0 is taken.** #194 landed it on `main` in `43b2e75` (2026-08-16). This needs 2.15.0, plus the matching `KNOWN_PARSER_VERSIONS` / `CURRENT_PARSER_VERSION` update — that's one of the three conflicts.
- **The PR is currently `CONFLICTING`** — `cloud/src/schemas/game.ts`, `src/lib/game-detail/helpers.ts`, `src/lib/game-detail/LeaderCard.svelte`. Every file it touches has moved on `main` since the branch point.
- **`rulerDisplayName` now duplicates `rulerName`** (`src/lib/game-detail/helpers.ts:1056`), which landed on `main` after this branched and does the same `formatEnum` + regnal-numeral job. Worth collapsing onto main's on rebase. `deathReasonLabel` is a genuine extract — `deathLabel` is still LeaderCard-local on `main`.
- **`const ROMAN = ["", "I", "II", …]`** in `science-techs.ts` duplicates `toRomanNumeral` (`$lib/utils/formatting:81`). "Archive II" *is* the game's own name (`text-project.xml:254` — `Archive II~an Archive II~Archives II`), so the label is right; the repo's idiom for turning a tiered zType into its real game name is a baked `*_NAMES` table (`IMPROVEMENT_NAMES`, with `formatEnum` as fallback), and this PR already extends the bake and already reads `project.xml`. A `PROJECT_NAMES` bake off `text-project.xml` would also give `PROJECT_CONVOY` its proper name and cover modded projects.

---

alcaras:
Went through this against the same three sources — `Reference/XML`, the game source, and current `main`. Everything here holds; I couldn't find a claim that didn't. Two additions, one scope question, and what the rebase will carry.

## Re-verified

- **Damage + assimilate.** `calculateTotalYieldModifierForGovernor` is governor + happiness + damage + assimilate, and `YIELD_SCIENCE` carries neither `mbNoDamage` nor `mbNoAssimilate`, so both apply — `Math.Max(-AssimilateTurns, -50)` and `Damage × -50 / getHPMax()`, with `CITY_ASSIMILATE_YIELD_MODIFIER` and `CITY_DAMAGE_YIELD_MODIFIER` both `-50` in `globalsInt.xml`. "Discontent was the big unpriced negative" being two-thirds true is fair.
- **The two `aaiEffectCityYieldRate` grants.** `EFFECTCITY_TRAIT_SCHOLAR_ARCHETYPE_ALL` pays `iValue 20` keyed on `EFFECTCITY_PROJECT_ARCHIVE`; `EFFECTCITY_LAW_PHILOSOPHY_EXTRA` pays `10` keyed on `EFFECTCITY_PROJECT_FORUM`. The field is one the bake never reads at all, so I agree the finding is the *class*, not the two entries — and it's the same conditional-grant shape as Dualism, which is already implemented, so the objection to implementing them is weak.
- **The DLC projects.** Both pay through their `EffectCity` entry (`aiYieldRate` `20` each), which is why the science isn't visible in `project-event.xml` itself. One thing worth adding here: **`scripts/bake-project-icons.ts` already sweeps `project-event.xml` plus the four DLC siblings**, so this is reusing an idiom the repo already has rather than inventing one. That makes it cheap enough that I'd rather close it here than file it. Agreed that the 8 percent-science projects are a different unit and belong in the percent pass.
- **Mechanical.** `CURRENT_PARSER_VERSION` is `2.14.0` on `main` as of `43b2e75`, so this goes to `2.15.0` with the matching `KNOWN_PARSER_VERSIONS` entry. Confirmed the three conflicting files.
- Also took the correction on `bSingle` vs `abInvalidBy` + `finishProject` — the comment credits the wrong mechanism for the right result, and the empirical 0-of-282 is the better justification. That goes into the comment.

## Two additions

**1. `rulerDisplayName` isn't just a duplicate after rebase — it's a regression.** `main`'s `rulerName` resolves through `characterName()`, i.e. the baked `NAME_TEXT` lookup that landed in `3326a74` as the #216 fix. Mine does `formatEnum(firstName, "NAME_")`, which is exactly the pre-#216 behavior — so keeping it puts "Hittite Male03" back into the science rail's leader tooltip *and* into `LeaderCard`, on the branch that reintroduces it. It gets deleted rather than collapsed, and both call sites move to `rulerName`. (Small correction: it lives at `src/lib/utils/formatting.ts:501`; `helpers.ts:1056` is `main`'s `rulerName`.) `deathReasonLabel` stands as a genuine extract.

**2. `PROJECT_NAMES` belongs in `scripts/build-manifests.ts`**, not `bake-science-yields.ts` — that's where `IMPROVEMENT_NAMES` and `NAME_TEXT` are generated and where the sidecar-override machinery already is. Right idea, one file over. It also gets `PROJECT_CONVOY` its real name for free, as you note.

## What the rebase carries

1. `2.15.0` + `KNOWN_PARSER_VERSIONS`, and the three-file conflict resolution.
2. `rulerDisplayName` deleted → `rulerName` at both call sites.
3. `PROJECT_NAMES` bake in `build-manifests.ts`; the local `ROMAN` array goes.
4. `project-event.xml` + the four DLC siblings in the project sweep, mirroring `bake-project-icons.ts`.
5. Scholar-archetype (+2/archive-city) and Philosophy (+1/forum-city) via `aaiEffectCityYieldRate`. Bake-only, no parser change, and it closes the second half of #148 plus the archetype slice.

## One scope question: damage + assimilate

I lean toward taking these here too, and the reason is version economics rather than scope. They're two additive fields off the same City node `happiness_level` already comes from, and this PR is already paying the parser bump and the reparse. Deferred, they cost a bump of their own and a second full reparse for what is otherwise a ~10-line parser change and one more row in the modifier pass. If you'd rather keep the diff bounded I'll file them instead — your call, and I'll go with whichever you say.

## #152 and #212

Yes to the harness re-run — I'll price governors + discontent (+ damage/assimilate if they land here) and report on #152 with what actually happens to the 8 negative residuals. If they resolve, agreed that difficulty and `EFFECTCITY_BASE` need no parser change, and the point about `EFFECTCITY_BASE` lifting the reconstructed base that every percent estimate multiplies is the part I'd underweighted — it tightens governors, Library and discontent rather than just adding a flat row.

On #212: agreed on settling the bake shape now, since `bake-science-yields.ts` is open in this PR either way. The `PROJECT_INQUIRY` × 8 finding is the useful one — a project completion is exactly the kind of one-off the spikes are currently attributing to story events. And you're right that the new per-city `project_counts` beats player-level `projects_produced` for option C's caption; I'll say so on the issue.

becked:
Take damage + assimilate here — the version-economics argument is the right one. This bump is already buying a full reparse, and asking every game to be re-uploaded a second time for a ~10-line parser change and one more modifier row is a worse trade than a slightly wider diff. Go ahead with the rebase list exactly as you've written it, plus those two, all under `2.15.0`.

Your three corrections all land:

- **`rulerDisplayName`.** Right on both counts — it's in `formatting.ts`, not `helpers.ts`, and it's a regression rather than a duplicate. `rulerName` (`src/lib/game-detail/helpers.ts:1065`) resolves through `characterName()` (`src/lib/utils/formatting.ts:65`), the baked `NAME_TEXT` lookup #216 landed in `3326a74`; the new function is `formatEnum(firstName ?? "", "NAME_") || "Unknown"`, i.e. the pre-#216 behavior, and keeping it would carry that into `LeaderCard` as well. Delete it, both call sites to `rulerName`. `deathReasonLabel` stands.
- **`PROJECT_NAMES` → `build-manifests.ts`.** Agreed — that's where `IMPROVEMENT_NAMES` (`:517`) and `NAME_TEXT` (`:584`) are written and where the sidecar-override machinery already lives. Right idea, one file over, as you say.
- **The DLC sweep precedent.** `scripts/bake-project-icons.ts:71` is the idiom (`/^project-event(-.*)?\.xml$/`, base file first then the DLC siblings sorted). Reusing it beats introducing a second convention for the same sweep.

**One thing to watch on the damage row.** The denominator isn't a constant: `getHPMax()` is `CITY_HP + getHPMaxExtra()` (`City.cs:2330-2332`), and `HPMaxExtra` accumulates from `effectCity.miCityHP` (`City.cs:5550`). Seven EffectCity entries carry that field and three of them are Walls, Moat and Towers — so any walled damaged city has a max well above `CITY_HP`'s 20, and a hardcoded denominator overstates its penalty. Worth confirming the save writes the extra directly before falling back to reconstructing it; the new per-city `project_counts` covers Walls / Moat / Towers / Improvised Defences, but not Rising Star, Paranoid Capital, the Monumental Buddhas, the Hill Fort, or the Silk/Spice resource effects.

Everything else is agreed as written — the harness re-run on #152, settling the `bake-science-yields.ts` shape now while it's open in this PR rather than after #212 lands, and `project_counts` over player-level `projects_produced` for #212's option-C caption.

---

alcaras:
Rebased and pushed — six new commits on top, PR body updated. `MERGEABLE`, and `lint` / `svelte-check` / `prettier --check` are clean. The `cloud/` suite is green except the four `canonical-map-options` failures, which fail on `main` too and are nowhere near this branch.

## The rebase list, as agreed

- **2.15.0** with `KNOWN_PARSER_VERSIONS` and `CURRENT_PARSER_VERSION` in step; the three conflicts resolved.
- **`rulerDisplayName` deleted**, both call sites on `rulerName`. The tooltip picked up `rulerCognomen` while I was there — it had its own inline `formatEnum(c.cognomen, "COGNOMEN_")`, the same duplication one layer down, and the block is explicitly in the Military rail's style, which uses the helper.
- **`PROJECT_NAMES`** in `build-manifests.ts`, `ROMAN` gone. One thing this turned up: **the Economy tab had a second, divergent project label** (`EconomyTab.svelte:604`) appending an *arabic* numeral, so the same project read "Archive 2" there and "Archive II" on the Techs tab. Both now go through one `projectDisplayName` in `helpers.ts`, next to `improvementDisplayName`. 163 overrides baked, and `PROJECT_CONVOY` gets its real name as predicted.
- **The event-project sweep**, via `bake-project-icons.ts`'s predicate. Both DLC projects land at +2, matching your measurement.
- **Both `aaiEffectCityYieldRate` grants.** Baked as `LAW_PROJECT_SCIENCE` and `ARCHETYPE_PROJECT_SCIENCE`, resolved in the bake down to source → project → science so the consumer needs no effect graph — it already has the player's laws, their leader, and each city's project counts. The leader rule matches the court section's: a ruler off the throne pays nothing.
- **Damage + assimilation**, priced.

## On the damage denominator — you were right to flag it

Checked before building: **the save never writes the HP total.** `miHPMaxExtra` is `SimplifyIO` network state only — `City.writeXML` has no element for it or for `HPMax` — so the fallback you anticipated is the only option. Confirmed empirically too: a save with `PROJECT_WALLS` and `PROJECT_MOAT` on its cities carries `<Damage>` and no HP tag anywhere.

So the denominator is rebuilt from `project_counts`: `CITY_HP` (20) plus the baked `PROJECT_CITY_HP`. Enumerating `<iCityHP>` across `effectCity*.xml` gives six entries, and **four are project-backed** — Walls, Moat, Towers, Improvised Defences, all `bSingle`, so each pays its HP once however often it was completed. The two that aren't are `EFFECTCITY_PARANOID_CAPITAL` and `EFFECTCITY_IMPROVEMENT_HILL_FORT`; a city with either reads slightly harsher than the game charged it, and the code comment says exactly that.

The bake also now **asserts `YIELD_SCIENCE` still sets neither `<bNoDamage>` nor `<bNoAssimilate>`** — if a patch ever opts science out, the bake fails instead of leaving these rows quietly charging a penalty the game stopped applying.

## The #152 harness run

Ran the real parser + `scienceBreakdown` over **200 finished multiplayer saves (71 players)** via `npx tsx --tsconfig ./tsconfig.json`, with governors, discontent, the two grants, damage and assimilation all priced:

- The new rows fire for **24 of 71 players**. Philosophy pays **+1 to +21** science/turn (the +21 is an FFA Rome with a lot of Forums), a Scholar leader **+4 to +6**; **16 players** carry a damaged or assimilating city.
- **No player is left with a negative `Other` from over-counted city sources.** The only three negatives are eliminated realms sitting at rate 0, where the Competitive stipend is the entire remainder — a different bug shape, not an over-count, and worth its own look.
- The residual that remains is **positive and large — mean `|Other|` 46.7**. That is the shape #152 predicts: with the negative side now covered, what's left over is *unmodelled positive* sources.

So from this corpus, **#152 unblocks**: the negative-residual objection doesn't survive discontent + damage + assimilation being priced, and both of its sources need no parser change. Your point about `EFFECTCITY_BASE` being worth more than face value is the part I'd underweighted — as a `cityFlat` entry it lifts the base every percent estimate multiplies, so it tightens governors, Library and all three negative rows rather than just adding a flat row. Happy to take that next unless you want it in someone else's hands.

Different corpus from your 12 saves, so the numbers won't line up with your 8-of-39 — but the direction is unambiguous.

## #212

Untouched here, as agreed, and the bake shape is settled by what landed: `LAW_PROJECT_SCIENCE` / `ARCHETYPE_PROJECT_SCIENCE` establish the "source → thing the blob records → science" table shape in `bake-science-yields.ts`, so option B's `EVENTSTORY_* → can this grant science?` table drops in beside them without a second convention. I'll note the `project_counts` point on the issue.

PR #177, opened 2026-07-29
feat: Military tab — power-standing shift chips (depends on #176)

alcaras:
> **Depends on #176** — this branch stacks on it (the chip rendering and the shared standings machinery live there), so the diff below includes #176's commits until it merges. Only the last two commits are this PR: `feat(bake): military power tiers from power.xml` and `feat(military): power-standing shift chips on the rail`.

## What

The Military rail gains a power-standing row, the counterpart of the Techs rail's knowledge shifts: tier-shift chips for the game's **Much Weaker / Weaker / Similar / Stronger / Much Stronger** classification of one side's military power against the other.

power standing rail

## Anchoring

This is the game's own diplomacy-screen classification, end to end:

- `Player.calculatePowerOf` buckets **theirPower × 100 / ourPower** (integer division) through `InfoHelpers.getBestPercentValue` over power.xml's thresholds — ≤40 Much Weaker, ≤75 Weaker, ≤133 Similar, ≤250 Stronger, and Much Stronger the catch-all (no `iPercent` → `InfoPercentBase`'s int.MaxValue default).
- Both sides of that comparison are `Player.calculateTotalStrength()` — which is **exactly the value the save records per turn as `MilitaryPowerHistory`** (`Player.cs:16952`), i.e. the blob's `military_power` series, i.e. the same curve the Military Power chart above the rail already plots. No proxy, no approximation.

The thresholds are baked (`npm run bake:power-tiers`, joins `bake:all`; fails hard if power.xml loses its catch-all or renames tags).

## Mechanics

- Chips read `W→MW`, `≈→S`, `S→MS` — Similar gets the **≈** glyph because Similar and Stronger share an initial. Hover gives the full names and the exact percent: *"Weaker → Much Weaker · T12 — Military power at 36% of Rome's"*. The two players' rows mirror each other by construction.
- Same 3-turn hold as the knowledge chips, so threshold-hovering doesn't spam the rail; the final run always stands.
- The classify/debounce/flip machinery is extracted from science-techs into a shared `standings.ts` (a `getBestPercentValue` port + the run-length filter), and the knowledge markers become a thin adapter over it — one implementation for both rails. That refactor is why this stacks on #176.
- Rail-only change: renders for two-player matchups (the Military rail's existing gate); FFA games are untouched. No parser bump, no migration, no worker change.

## Testing

- `npm run lint` / `svelte-check` clean; no cloud-side changes.
- Verified live on re-imported duels: chips mirror between the players (Hittite `≈→W` where Rome shows `≈→S`), percents match hand-computed ratios from the plotted series, and the Hittite–Rome game's chips narrate the war visible in the chart (slides to Much Weaker by T12, recovers to Similar mid-game, Weaker↔Similar oscillation as the lines converge late).


alcaras:
Self-review (correctness + fit against CLAUDE.md's contributing checks), scoped to this PR's two commits (the rest of the diff is #176).

**Found and fixed** (amended into the branch)

- `standingTier` was exported from `standings.ts` with no consumer outside the module — now module-private.

**Checked, no finding**

- *The sparse-save trap*: the save writes `MilitaryPowerHistory` only on change (`setTurnMilitaryPower` skips equal values), so naive turn-matching of the two players' series would misclassify stretches where only one side's power moved. Not an issue here: `derivePlayerHistory` forward-fills every turn 1..totalTurns before the blob is written, so both series are dense and the per-turn ratio is exact.
- *Direction and tie behavior*: percent = my power × 100 / opponent's, which is `getStrength(mine, theirs)` — "my standing as the opponent classifies me", matching the tooltip wording; the `getBestPercentValue` port keeps C#'s later-entry-wins-on-equal-threshold behavior. Unlike the knowledge chips (whose cumulatives are ÷10-rounded in the blob), military power is stored raw, so this classification is bit-exact against the game's.
- *Reuse*: history resolution via the shared `findByPlayer` idiom; chips ride #176's `RailMarker.label` chain unchanged; the knowledge markers were refactored onto the same `standings.ts` machinery rather than leaving two copies of the classify/debounce/flip logic; `EventRail`'s only two consumers (Military, Techs) both verified.
- *Labels*: the initial-collision (Similar/Stronger → "S") is handled with the ≈ glyph for Similar and a per-set shorthand rule; knowledge keeps plain initials (P/N/C/L/E are unique).
- *Bake*: `bake:power-tiers` registered and added to `bake:all`; fails hard if power.xml loses its catch-all entry or renames tags; no hand edits to generated files.
- *Nulls/colors/literals*: `??`/`!= null` in the data path, chips in the player color, no new hex.

- Two-player matchups only — the Military rail's existing gate; FFA unchanged.
- The 3-turn hold before a shift counts is a display choice (threshold-hovering flip-flops the raw classification every turn), documented in `standings.ts`; the final run always stands so the last chip agrees with the end state.
- Chips read the recorded end-of-turn power, same as the chart above — intra-turn swings during a battle round aren't visible in the save.

becked:
**Notes to ourselves — parking these to revisit, not a change request on the branch.** Nothing below blocks; it's the list I want in front of us when we come back to this.

Rebased onto current `main` and pushed (`241d2fe`, `74c94f2`). #176 merged with rewritten SHAs — its merged form is parser 2.15.0, this branch was stacked on the pre-merge 2.14.0 — so a plain rebase tried to replay all of #176; replayed just this PR's two commits with `git rebase --onto`. Conflicts were mechanical: `bake:all` took main's longer list with `bake:power-tiers` inserted before `bake:wonders`, and `science-techs.ts` kept main's imports plus the new `./standings` one. Also ran prettier on `bake-power-tiers.ts` and `TechsTab.svelte` (they were the repo's only `format:check` failures) and squashed each fix into its originating commit. `lint`, `svelte-check` (1317 files, 0 errors), and `format:check` are clean.

### The anchoring checks out

Verified against `Reference/Source`, not taken on faith:

- `Player.cs:19479` — `calculatePowerOf(ePlayer)` is `getStrength(player(ePlayer).calculateTotalStrength(), calculateTotalStrength())`.
- `InfoHelpers.cs:2393` — `iPercent = iTheirValue * 100 / iOurValue`, smallest threshold >= iPercent wins. `standingTier()` reproduces this exactly.
- `Player.cs:17022` — `setTurnMilitaryPower(turn, calculateTotalStrength())`. `MilitaryPowerHistory` really is the compared quantity; no proxy.
- `power.xml`'s 40/75/133/250 + capless `MUCH_STRONGER` match the baked table, and the blank template `<Entry>` is correctly dropped by `.filter(e => e.zType)`.
- The sparse-history worry is already handled upstream: `MilitaryPowerHistory` only records turns where power *changed*, but `derive/player-history.ts` forward-fills to a dense 1..N series, so the per-turn join is real.
- The bake script matches the house pattern exactly (the `resolveConfig` + prettier + "no changes" short-circuit family: science-yields, orders-sources, family-colors, owtt).

### To revisit

1. **"Mirrors by construction" is over-claimed.** At reciprocal boundaries the rows disagree: a player at exactly 40% reads Much Weaker, but the opponent's 250% reads *Stronger*, not Much Stronger — both buckets are inclusive->=. Same at 75 <-> 133. The code is faithful to the game's own asymmetry; it's the description that's too strong. The debounce also runs independently per row, so the two rows can shift on different turns.

2. **Eliminated players.** `setTurnMilitaryPower(turn, 0)` fires on player death (`Player.cs:6927`). `standingShiftMarkers` skips turns where the opponent is 0, so the survivor's row gets no chip at the kill while the dead player's row gets a `->MW`. The game would classify that as Much Stronger (`iOurValue <= 0` -> `int.MaxValue`). Probably immaterial since the game ends there, but it's a real duel scenario.

3. **`POWER_ABBREV` is a hand-maintained map keyed on baked zTypes.** The Techs side derives its shorthand (`knowledgeName(t)[0]`); power can't, since Similar/Stronger collide — hence the `≈`. If power.xml gains or renames a tier, the bake still passes (>=3 tiers, one catch-all) and the UI silently falls back to ambiguous initials via `powerAbbrev`'s `?? powerName(t)[0]`. That fallback is also the kind of unreachable branch our "no dead or speculative code" rule discourages.

4. **Is `standingShiftLabel` earning its keep?** Its whole body is `${abbrevOf(from)}→${abbrevOf(to)}`; the caller supplies the interesting half. Two call sites, three lines of indirection. Inlining the template at both sites and sharing only `standingShiftMarkers` is the alternative.

5. **Structural divergence from the Techs rail.** `TechsTab` declares `SCIENCE_RAIL_KINDS` and builds a `Record<kind, RailMarker[]>`, then filters; `MilitaryTab` appends the `power` row imperatively after the `RAIL_KINDS.filter(...)` map, because power markers don't come from `eventRail`. Row *order* is consistent (standing last in both) and `orderedPlayers[1 - i]` copies the Techs opponent idiom exactly — the question is whether the Military rail should adopt the record-and-filter shape rather than growing a special case.

6. **Two lookup idioms now live in `MilitaryTab.svelte`.** This PR uses `findByPlayer(playerHistory, ...)` while `endingPower` (line ~146) does a plain `.find(h => h.player_id === p.playerId)`. Since `PlayerHistory.player_id` is a non-nullable `number`, `findByPlayer`'s nation fallback is unreachable here, so they're equivalent and the PR's choice is the repo helper — it's just uneven within one file.

7. **Comment drift in the shared `standings.ts`.** Its header says "percent = ours × 100 / theirs" while the generated `power-tiers.ts` header says "theirPower × 100 / ourPower". Same math from different vantage points, but side by side it reads like a contradiction. And the "blob's ÷10 rounding" rationale it inherited from the knowledge case is a property of the science series, not of military power.

No tests here, which is consistent — `src/` has no test files repo-wide.

PR #179, opened 2026-07-29
feat: Economy rail — city markers wear their family crest (depends on #169)

alcaras:
> **Depends on #169** (the Economy tab) — this branch stacks on it, so the diff includes the economy-stack commits until #169 merges. Only the last two commits are this PR.

## What

The Economy rail's cities-founded markers drop the generic CITY_FOUNDED icon for **the city's family crest**, and the tooltip gains the founding ordinals:

city crest markers

- The family is the one during the **founder's** tenure (`player_families`), so a captured city keeps its founder's crest rather than the conqueror's — same rule the map's turn slider uses.
- Tooltip: crest inline in the title, then `Founded on turn 20` / `Landowners city #2` / `Overall city #4` — both ordinals counted per player in founding order.
- Crest resolution prefers the per-family crest when we ship the art, else the always-available `FAMILYCLASS_X → ARCHETYPE_X` fallback.

## Comes with a real bug fix

Extracting that resolution rule surfaced that the repo had **three** copies of it, and the map's copy was broken: blob family values already carry the `FAMILY_` prefix (`FAMILY_ANTIGONUS`) and the art ships as `CREST_FAMILY_ANTIGONUS`, but SpriteMap's existence check looked up `CREST_FAMILY_${family}` → `CREST_FAMILY_FAMILY_…`, which never exists — so per-family crests **never rendered on the map tooltip** and silently fell back to the archetype crest. The first commit extracts one shared `familyCrestKey` into `game-detail/helpers` (fixing the lookup) and points SpriteMap, the Cities table's Family column, and the new rail markers at it.

## Mechanics

- `EventRail`'s crest sizing splits in two: the nation crest labelling each band keeps its 24px (now its own constant); crest *markers* render at the 16px marker default so family crests sit at marker weight.
- UI-only — no parser, worker, or schema changes; pre-2.10.0 blobs (no `player_families`) fall back to the city's current family, and a city with no resolvable crest renders the existing colored dot.

## Testing

- `npm run lint` / `svelte-check` clean on the branch.
- Verified live: distinct crests per marker on both players' rows, ordinals hand-checked against the founding order, capture case exercised via `player_families`, and the map tooltip unaffected visually for archetype-crest families (the per-family crest fix only changes families whose art we actually ship — the Greek-DLC set).

### DIFF
--- a/src/lib/SpriteMap.svelte
+++ b/src/lib/SpriteMap.svelte
@@ -11,6 +11,7 @@
 		NATION_ALIASES_URL,
 	} from "$lib/generated/atlas-manifest";
 	import { SPRITE_MANIFEST } from "$lib/generated/sprite-manifest";
+	import { familyCrestKey } from "$lib/game-detail/helpers";
 	import MapTooltip from "$lib/MapTooltip.svelte";
 	import Checkbox from "$lib/ui/Checkbox.svelte";
 
@@ -302,10 +303,8 @@
 	});
 
 	// city_name → resolved crest sprite key (or null when no crest is
-	// renderable). Resolution prefers a per-family crest when we ship the
-	// art; otherwise falls back to the family-class archetype crest, which
-	// always exists for non-null family_class values. Existence is checked
-	// against the SPRITE_MANIFEST baked from static/sprites/crests/CREST_FAMILY_*.png.
+	// renderable), via the shared familyCrestKey rule (per-family crest when
+	// we ship the art, else the family-class archetype crest).
 	//
 	// The family is resolved for whoever owns the city at the represented turn:
 	// a city's family only changes when another nation conquers it, so
@@ -321,16 +320,10 @@
 				ownerAtTurn != null
 					? c.player_families?.find((f) => f.player_xml_id === ownerAtTurn)
 					: undefined;
-			const family = pf?.family ?? c.family;
-			const familyClass = pf?.family_class ?? c.family_class;
-			let key: string | null = null;
-			if (family && SPRITE_MANIFEST[`crests/CREST_FAMILY_${family}`] != null) {
-				key = family;
-			} else if (familyClass) {
-				// FAMILYCLASS_CHAMPIONS → ARCHETYPE_CHAMPIONS
-				key = familyClass.replace(/^FAMILYCLASS_/, "ARCHETYPE_");
-			}
-			map.set(c.city_name, key);
+			map.set(
+				c.city_name,
+				familyCrestKey(pf?.family ?? c.family, pf?.family_class ?? c.family_class),
+			);
 		}
 		return map;
 	});
--- a/src/lib/SpriteMap.svelte
+++ b/src/lib/SpriteMap.svelte
@@ -11,7 +11,7 @@
 		NATION_ALIASES_URL,
 	} from "$lib/generated/atlas-manifest";
 	import { SPRITE_MANIFEST } from "$lib/generated/sprite-manifest";
-	import { familyCrestKey } from "$lib/game-detail/helpers";
+	import { familyCrestKey, familyForOwner } from "$lib/game-detail/helpers";
 	import MapTooltip from "$lib/MapTooltip.svelte";
 	import Checkbox from "$lib/ui/Checkbox.svelte";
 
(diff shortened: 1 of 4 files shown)


alcaras:
Self-review (correctness + fit against CLAUDE.md's contributing checks), scoped to this PR's two commits (the rest of the diff is the #168/#169 stack).

**Verified, no finding**

- *The fix's premise*: swept every family value in real blobs (city `family`, `player_families`, the families list) — all carry the `FAMILY_` prefix, none bare — so correcting the lookup from `CREST_FAMILY_${family}` to `CREST_${family}` can't regress any legacy shape, and the manifest keys (`crests/CREST_FAMILY_ANTIGONUS`) confirm the old check could never have matched.
- *Deduplication is complete*: no `FAMILYCLASS_ → ARCHETYPE_` fallback remains outside `familyCrestKey`; SpriteMap, the Cities table's Family column, and the new rail markers all delegate. The Cities column already had the correct lookup, so its behavior is unchanged — only the map tooltip gains the per-family crests it was always supposed to show.
- *Frozen `web/` viewer*: SpriteMap is symlinked into the legacy share viewer, and its new `$lib/game-detail/helpers` import resolves there — `game-detail/` is in the same symlink set, and helpers was already in the viewer's dependency closure via the tab components.
- *Ordinals*: cities sorted by `founded_turn` with JS's stable sort, counted per player; the family ordinal keys on `family_class`, which is one-family-per-class per player in Old World, and the founder's family comes from `player_families` (the same capture-safe rule the map's turn slider uses) with the current family as the pre-2.10.0 fallback.
- *Rail behavior*: a null crest degrades to the existing colored dot; near-simultaneous foundings collapse under the rail's existing proximity merge (first crest kept, tooltips stacked) — same as every other marker kind. The `EventRail` size split touches a different hunk than #176's label extension, so the two merge cleanly in either order.
- *Nulls/colors/literals*: `??` in the data path, `||`-free, the inline crest `` follows the established tooltip-icon idiom, no new hex.

- The tooltip's inline crest and the marker show the same art at slightly different sizes — deliberate, so the tooltip is self-contained when markers merge.
- "Overall city #N" counts cities the player *founded* (matching the marker set); captured cities aren't in anyone's ordinals.

becked:
Rebased on current `main` (`59a3203`) — clean, no conflicts — and pushed a review-fix commit (`d2565a5`).

Reviewed for fit against the surrounding code. What I changed:

- **`familyCrestKey` re-implemented the `CREST_` prefix rule.** It checked the manifest by hand-building `crests/CREST_<family>`, while `getSpritePath` already owns that prefix for the `crests` category (`helpers.ts:509-511`) — and the code this replaced in `CITY_COLUMNS` went through `getSpritePath`. Same drift class as the double-prefix bug `de5b492` fixes. Now routed through `getSpritePath`.
- **The per-owner family resolution was copied at both call sites.** `player_families.find(...)` then `?? c.family` / `?? c.family_class` appeared in `SpriteMap.svelte` and `EconomyTab.svelte` — the only two `player_families` consumers in the repo. Extracted `familyForOwner(city, playerXmlId)` next to `familyCrestKey`; the map passes the owner-at-turn, the rail passes the founder.
- **Tooltip wording.** `Founded on turn N` → `Founded turn N`, matching the wonder marker on the same rail (`Completed turn N`).
- **Folded the marker's parallel `familyClass ?` ternaries** into one branch, which drops the `nthOfFamily!` assertion.

Checked and left alone:

- Null `iconValue` is handled — `EventRail` falls back to a colored dot, so a city with no resolvable crest renders fine.
- Dropping `crests: 24` from `RAIL_ICON_SIZE` in favour of `RAIL_GUTTER_CREST_SIZE` is safe: these city markers are the only crest *markers* in any rail, and they land on 16 — the same size the old `icons`/`CITY_FOUNDED` markers rendered at.
- `CITY_COLUMNS` behaviour is unchanged by the extraction (the anchored regex is stricter, not different).
- The archetype fallback claim holds — all 10 real `FAMILYCLASS_*` zTypes have a matching `CREST_ARCHETYPE_*` in the sprite manifest.
- The `??` fallbacks are safe: the parser types `playerFamilies` as `{ playerXmlId: number; familyName: string }[]`, so a found entry never has a null family.

One thing noted and deliberately left as-is: the tooltip labels a city by its family **class** (`Champions city #2`) while the Cities and Families tabs name the family (`Sargonid`). Defensible either way — the archetype crest is what actually renders for all but a handful of families — so keeping it.

`npm run lint` and `svelte-check` clean (1309 files, 0 errors). No `cloud/` files touched.

PR #180, opened 2026-07-29
feat: Families tab — the game's family colours on lines and names (depends on #170)

alcaras:
> **Depends on #170** (the Families tab) — this branch stacks on it, so the diff includes the #168–#170 stack until those merge. Only the last two commits are this PR.

## What

Each family's opinion line — and its name in the turns-by-band chart below — now uses **the game's own colour for that family**, instead of the generic chart-series cycle for lines and the nation colour for names. The name below reads as a key for the line above: Claudius purple, Fabius salmon, Julius gold, exactly as in-game.

family colours

## Anchoring

The game gives every family × nation combo a unique colour, resolved through a four-file chain: `family.xml` (`TeamColor` + `iColorIndex`) → `teamColor.xml` (`aePlayerColors[i]`, zero-based per `City.cs:3264`) → `playerColor.xml` (`AssetColor`) → `color.xml` (hex). Families belong to exactly one nation, so the family zType alone keys the colour. Baked as `FAMILY_COLORS` (53 families) by a new `bake:family-colors` (joins `bake:all`; fails hard on an unresolvable palette assignment rather than silently dropping a family).

## Mechanics

- Nation headings in the band chart stay in the axis white — the coloured names under them are what tie a family's row to its line.
- Fallbacks follow the repo's `?? getChartColor(i)` idiom: a family the bake doesn't know (mods) gets the indexed chart colour on its line and the player colour on its name.
- UI-only — no parser, worker, or schema changes.

## Testing

- `npm run lint` / `svelte-check` clean on the branch.
- Verified live on a Hittite–Rome duel: all six families' lines and names match, and the colours agree with the in-game family palette.


alcaras:
Self-review (correctness + fit against CLAUDE.md's contributing checks), scoped to this PR's two commits. A clean pass — no findings to fix; the verification trail:

**Verified, no finding**

- *Palette indexing*: zero-based per `City.cs:3264` (`maePlayerColors[family().miColorIndex]`), and empirically confirmed against the in-game palette — Fabius (index 0) resolves to `#ffae8b` and Julius (index 3) to `#e9bf4e`, matching what the game shows for those families.
- *Coverage*: every family in real blobs is in the baked table (all six of the test duel's, plus the DLC nations — 53 total); the bake fails hard on an unresolvable palette assignment rather than silently dropping a family, and refuses a suspiciously small table.
- *Parallel surfaces*: swept the branch for other family-coloured renders — the Families tab's two charts are the only family-keyed series/labels. The worker-turns-by-family chart deliberately stays nation-coloured (its bars compare the two nations within each family row, so nation colour is the correct key there), and the band chart's nation headings stay in the axis white by design.
- *Fallbacks*: the repo's documented idiom (`FAMILY_COLORS[f] ?? getChartColor(i)` on lines, player colour on names) for families the bake doesn't know — mods.
- *Fit*: bake follows the established self-contained pattern (registered, joins `bake:all`, no hand edits to generated files); no dead exports (`FAMILY_COLORS` has its consumer); no new hex literals in components — the colours all come from the generated game data.

- The tooltip heading in the band chart keeps the player colour — it names the nation context, not a family.
- Colours are the game's `AssetColor`; the XML also ships `_TEXT` variants but they're identical for family colours, so only one table is baked.

alcaras:
Second, adversarial review pass (treating the earlier self-review as the author's claims and trying to refute them). One real gap found — **fixed and folded into the bake commit** (branch force-pushed):

**Found and fixed**

- **The bake missed the campaign/scenario families.** The Mods scenarios add ~30 families the same way the base game does (`Mods/*/Infos/family-add.xml` — Thebes, Hyksos, Kush, the Amarna dynasties…), with their own `teamColor-add`/`color-add` chains, and uploaded campaign saves carry them. The repo's own precedent (`bake-improvements`) sweeps `Mods/*/Infos/*-{add,change}.xml`; the first cut of this bake read only the base files, so campaign families would have silently taken the fallback colours. The bake now merges base + Mods last-wins with per-field `-change` semantics — **82 families** (53 base + 29 scenario), every chain resolving. Base-family resolution failures still fail the bake; unaudited mod content warns and falls back instead, matching `bake-improvements`' stance. Base colours are byte-identical to the first bake (spot-checked Julius `#e9bf4e`), so the rendered output for normal games is unchanged.

**Re-verified from the first pass, still holding**

- Zero-based palette indexing (empirically matches the in-game palette), full blob coverage, the `?? getChartColor(i)` fallback idiom, no other family-coloured surfaces on the branch (the worker-turns chart correctly stays nation-keyed).

**New disclosure**

- The darkest family colour in the table is Xidun's `#AD002C` (a deep crimson) — legible as a 2px chart line, borderline as 11px label text on the dark panel. It's the game's own colour, so it's kept faithful rather than lightened; every other family sits comfortably above it.

becked:
Reviewed, and pushed two follow-up commits to this branch.

**What I verified.** Re-derived the four-file join from a local `Reference/XML` and the committed `family-colors.ts` matches exactly — 82/82 keys, zero diffs, no family silently skipped. The `City.cs:3264` citation is verbatim (`infos().teamColor(family().meTeamColor).maePlayerColors[family().miColorIndex]`) and the zero-based read reproduces the screenshot: Fabius `#ffae8b`, Claudius `#c9abff`, Julius `#e9bf4e`. `bake:finalize`'s reconcile only walks the atlas/sprite asset dirs, so the new self-contained module is safe from orphan deletion. Nothing else in the app renders a named family in colour — the Cities table's Family column is crest + plain text, `stats/charts/families.ts` works in family *classes*, and `MapTooltip` draws a crest only.

**e8d4e1f — resolve family colour through one config helper.** The lookup-plus-fallback was inlined at both call sites with fallbacks that disagreed: the line took `getChartColor(i)`, the name took `t.player.color`. So for a family the bake doesn't know, the name stopped keying the line above it — the one case the feature exists to handle. `getFamilyChartColor(family, fallbackIndex)` now lives in `$lib/config` mirroring `getNationChartColor`, and both panels go through it.

Making them agree needed the index as well as the helper: `opinionBandTallies` sorts family rows by founding turn while `familyOpinionSeries` sorts alphabetically, so a band row's own `i` was never the number its line used. `familySeriesIndex` maps a family to its position in its own nation's series, and both panels index by that.

Same commit updates the opinion-chart section comment, which still said families "get distinct colours from the shared series palette" — the behaviour this PR replaced.

**5899c1d — bake ordering.** The `bake:*` script list follows `bake:all`'s run order everywhere else, so `bake:family-colors` moved ahead of `bake:wonders` to match. Also dropped the baker count from the `bake:all` comment in the bake skill, since it drifts every time one lands.

**One thing left for you:** the description says the bake produces 53 families. That's the count in base `Infos/family.xml` alone — the committed table has 82, the rest coming from the campaign Mods the script merges. The behaviour is right; the number in the writeup isn't.

`npm run lint` clean, `svelte-check` 1309 files / 0 errors / 0 warnings, `prettier --check` clean.

PR #182, opened 2026-08-01
fix(video): date live content by when it aired, not when its VOD went up

becked:
Reported in Discord by Konstant: the "X hours ago" pills on the home creator feed are wrong. Confirmed on a freshly-opened tab, and confirmed to affect archived streams but not a currently-live one.

## The bug

Nearly every video in the creator feed is a tournament cast, and casts are streamed live. For live content, the channel Atom feed's `<published>` is when the **VOD** was published — which lands hours after the broadcast ended, routinely in the next calendar day. `videoFromEntry` reads that tag straight into `Video.published_at`, so the feed dated every cast by its VOD publish instant. Because that instant is *later* than the air time, casts rendered as more recent than they are.

Six casts from one channel, sampled 2026-08-01. `startTimestamp` is from the watch page's `liveBroadcastDetails`; `<published>` is the Atom feed's value, identical to the watch page's `publishDate`/`uploadDate`:

| video | aired (`startTimestamp`) | `<published>` (what we used) | skew | YouTube's own label |
|---|---|---|---|---|
| `e5eFJgzYz3Q` fiddlers25 v nobody, Part 5 | 2026-07-31T01:04:43Z | 2026-07-31T15:27:03Z | +14.4h | Streamed live on **Jul 30** |
| `CLauyN3KQys` fiddlers25 v nobody, Part 4 | 2026-07-29T01:04:34Z | 2026-07-29T16:09:39Z | +15.1h | Streamed live on **Jul 28** |
| `LE9A41q8GCk` Sabertooth v Nicknight, Part 3 | 2026-07-27T02:01:09Z | 2026-07-27T15:52:18Z | +13.9h | Streamed live on **Jul 26** |
| `8SJOXJod6rA` alcaras v HazardBringsAxe, Part 2 | 2026-07-26T18:00:09Z | 2026-07-27T08:00:24Z | +14.0h | Streamed live on **Jul 26** |
| `ilZDbW_0p1o` alcaras v HazardBringsAxe | 2026-07-25T18:01:04Z | 2026-07-25T22:13:47Z | +4.2h | Streamed live on **Jul 25** |
| `nqMGQ-svQGk` fiddlers25 v nobody | 2026-07-17T00:18:21Z | 2026-07-17T15:29:35Z | +15.2h | Streamed live on **Jul 16** |

All six report `isLiveContent: true`. Skew +4h to +15h, five of six crossing a day boundary.

This reproduces the report exactly: at ~2026-08-01T11:27Z the feed showed Part 5 as "20 hours ago" — precisely `now − 2026-07-31T15:27:03Z` — while the stream aired ~38 hours earlier, which YouTube presents as Jul 30. It also explains why a currently-live stream looked right: a stream that hasn't ended has no VOD publish instant to skew away from its start.

**Not caching**, which was the first hypothesis. The KV SWR layer and the `s-maxage=60` on `/v1/creator-videos` affect *which* videos appear, not their timestamps; the rendered label matched `<published>` to the minute. **Not the last release** either — `git log c0ef8b4..HEAD` over the home page, `VideoCard`, `formatting.ts`, `channels.ts`, and `video/` is empty.

## Why videos.list

The Atom feed carries nothing about the broadcast. `<published>` and `<updated>` are its only two timestamps, and `<updated>` tracks view-count churn — it moved ~35 minutes before I fetched it because `media:statistics` ticked — so it sits *further* from the air time, not closer. oEmbed returns no timestamps at all. The watch page does carry `liveBroadcastDetails`, but that's an undocumented blob inside ~1.2 MB of HTML, one fetch per video. `videos.list` is the only supported source.

## The changes

**`parseVideosListPage` / `applyBroadcastStarts`** — pure, unit-tested. Maps video id → `liveStreamingDetails.actualStartTime`, then re-dates. An ordinary upload carries no `liveStreamingDetails` at all, so its absence from the map is the discriminator rather than a sentinel value; a broadcast scheduled but never aired has the object without `actualStartTime` and is likewise absent.

**`fetchBroadcastStarts`** — batched at 50 ids (`videos.list` costs 1 unit per call regardless of id count). Never throws, unlike the feed fetches around it: enrichment improves a date we already have rather than being a precondition for having one, so a quota-exhausted or failing call degrades that batch to its feed `<published>` instead of discarding a feed already fetched successfully. Batches are independent, so one bad page doesn't cost the others. It reports whether any batch failed, which is what the cache decision below turns on.

**A degraded result is served but never cached.** Falling back to the feed's `<published>` is the pre-change behavior, but *caching* that fallback isn't: `getVideosCached` would write the VOD dates to KV and serve them for up to the 1h soft / 24h hard TTL, which is the bug this branch exists to fix, re-frozen for an hour. It also runs against the module's own convention — every other fetch path throws on a transient upstream failure precisely so a prior good entry survives. So both keyed paths now finish their pipeline (sort, dedupe, cap) and then throw `UncacheableVideos` carrying the finished list; `refresh()` serves that payload and skips the write. A warm entry keeps its prior good dates; a cold miss serves the feed dates once and retries on the next request.

`fetchRecent`'s signature is unchanged, so the throw is the whole contract. The one caller outside the cache layer — `./per-ankh admin add-channel`, whose post-write sanity check calls `fetchRecent` directly — unwraps it and uses the payload, since a one-off "does this channel id point at a live feed?" doesn't care about the dates. It matches on `name` rather than `instanceof` because `cloud/` is CJS and its runtime named exports don't cross into that ESM script (same reason the registry is imported through default interop there).

**`withBroadcastStarts`** — no key ⇒ returns the input untouched, and that is the documented keyless behavior rather than a degraded result, so it doesn't suppress caching. Dedupes ids so a video a curator listed twice doesn't burn a slot in the 50-id batch.

**Wired into both keyed paths** — `fetchYouTubeRecent` (home feed + profile Videos tab) and `fetchYouTubePlaylistVideosViaApi` (tournament Videos tab). The keyless RSS playlist path is deliberately unchanged: it only runs when there is no key to call `videos.list` with, so there is nothing to wire.

**Enrichment runs before the sort in both paths.** Ordering was affected too, not just labels — re-dating a cast can move it past a video published after it, and two casts from one evening would otherwise sort by when their VODs went public the next day.

**`byPublishedDesc` extracted to `video/types.ts`.** The comparator existed in four inline copies — two in `youtube.ts`, two in `channels.ts` (`handleUserVideos` and `mergeCreatorFeed`, which is where the "newest first" contract is actually promised to callers) — and this branch adds a fifth call site. One definition, five users.

**`CACHE_VERSION` 4 → 5** so warmed entries holding VOD dates orphan, rather than serving wrong dates for up to the 24h hard TTL.

**Docs.** Four places described the recent-videos path as needing no key — `cloud/wrangler.toml`, two endpoints in `docs/api-reference.md`, the admin CLI, and the `VideoEnv` doc comment. `docs/api-reference.md` also now states what `published_at` means for live content on all three video endpoints, and that a failed correction is served but not cached.

**Logging.** A non-OK `videos.list` logs at error, like the file's other two upstream failures, rather than putting the same event name on two levels.

## Cost

One extra quota unit per channel refresh (`MAX_VIDEOS` is 12, so one call), amortized behind the existing 1h SWR TTL — 24/day per linked channel, against a 10k/day default. The playlist path adds `ceil(n / 50)`, so up to 10 units on top of its existing 10.

The structural cost: the channel hot path was advertised as key-free, and this makes a keyed build spend quota on it. It degrades to the feed date without a key rather than becoming a hard dependency, and the module header and `wrangler.toml` are rewritten to say so.

## Known limitation

On the channel path the cap to `MAX_VIDEOS` runs on the feed's own order, so the feed's dates still decide *which* entries we keep — enrichment only ever moves a broadcast earlier, so one of the ~3 entries past the cap can end up newer than a kept broadcast and go unsurfaced anyway. Labels and ordering within the kept set are correct. Noted in the code rather than fixed; moving the slice after the sort would cost nothing in quota if we decide the membership edge case matters.

## Verification

Rebased on current `main` (the D1/R2 instrumentation work; no overlap with `cloud/src/video/`).

- `cloud/` `tsc --noEmit` — clean
- `cloud/` `npm test` — 578 passing, 10 of them new (6 for the date correction, 4 for the cache-skip)
- `npm run check` (svelte-check) — 1290 files, 0 errors
- `npm run lint` and `prettier --check` — clean
- `scripts/` isn't covered by any configured typecheck, so `scripts/admin/commands/channels.ts` was checked with `tsc` directly — clean

The 4 `canonical-map-options` failures are pre-existing and unrelated (#115/#141) — confirmed by running that file on a stashed clean tree, where it fails identically.

**Not verified:** the `videos.list` call itself is only exercised against real YouTube with a key. The tests cover the pure parse/apply functions and the cache's serve-don't-write branch, not the network path — worth a staging check before prod.

## Not in scope

`formatRelativeToNow` reads `Date.now()` directly rather than the repo's existing reactive clock (`nowMs()`, `src/lib/stores/now.svelte.ts`), so none of its four callers re-render as time passes and a page left open shows a frozen badge. That was the first hypothesis here and Konstant ruled it out by reproducing on a fresh tab. Now filed as #184.

https://claude.ai/code/session_018MoWizF9ZomtCiKd51nabx

### DIFF
--- a/cloud/src/channels.ts
+++ b/cloud/src/channels.ts
@@ -25,6 +25,7 @@ import {
 	supportedPlatforms,
 } from "./video/registry";
 import {
+	byPublishedDesc,
 	ChannelResolutionError,
 	type Video,
 	type VideoEnv,
@@ -207,8 +208,8 @@ export async function handleUserVideos(
 
 	const videos = perChannel
 		.flat()
-		// Newest first across all platforms; ISO timestamps sort lexically.
-		.sort((a, b) => (a.published_at < b.published_at ? 1 : -1))
+		// Newest first across all platforms.
+		.sort(byPublishedDesc)
 		.slice(0, MAX_MERGED_VIDEOS);
 
 	return jsonResponse({ videos }, 200, cors);
@@ -243,15 +244,12 @@ export interface CreatorVideo extends Video {
 
 // Merge per-channel lists (each already attributed to its creator) into the
 // home feed: newest-first across all creators, capped. Pure — the DB query and
-// per-channel fetch live in buildCreatorFeed. ISO timestamps sort lexically.
+// per-channel fetch live in buildCreatorFeed.
 export function mergeCreatorFeed(
 	perChannel: CreatorVideo[][],
 	cap = MAX_CREATOR_FEED_VIDEOS,
 ): CreatorVideo[] {
-	return perChannel
-		.flat()
-		.sort((a, b) => (a.published_at < b.published_at ? 1 : -1))
-		.slice(0, cap);
+	return perChannel.flat().sort(byPublishedDesc).slice(0, cap);
 }
 
 // Assemble the feed: every linked channel joined to its owner, each channel's
--- a/cloud/src/video/cache.test.ts
+++ b/cloud/src/video/cache.test.ts
@@ -1,0 +1,113 @@
+import { describe, expect, it } from "vitest";
+import { getVideosCached } from "./cache";
+import { UncacheableVideos, type Video, type VideoEnv } from "./types";
+
+const video = (id: string, published_at: string): Video => ({
+	id,
+	title: id,
+	url: `https://www.youtube.com/watch?v=${id}`,
+	thumbnail_url: null,
+	published_at,
+	platform: "youtube",
+});
+
+const AIRED = [video("BROADCAST01", "2026-07-31T01:04:43Z")];
+const VOD_DATED = [video("BROADCAST01", "2026-07-31T15:27:03Z")];
+
+// The cache only ever get()s and put()s one key, so a Map stands in for KV.
+// `puts` records writes, which is what these tests are actually about.
+function fakeEnv(): {
+	env: VideoEnv;
+	store: Map<string, string>;
+	puts: string[];
+} {
+	const store = new Map<string, string>();
+	const puts: string[] = [];
+	const env = {
+		SESSIONS_KV: {
+			get: (k: string) => Promise.resolve(store.get(k) ?? null),
+			put: (k: string, v: string) => {
+				puts.push(k);
+				store.set(k, v);
+				return Promise.resolve();
+			},
+		},
+	} as unknown as VideoEnv;
+	return { env, store, puts };
+}
+
+// Collects the background refresh the stale path hands to waitUntil, so a test
+// can await it instead of racing it.
+function fakeCtx(): { ctx: ExecutionContext; settled: () => Promise<unknown> } {
+	const pending: Promise<unknown>[] = [];
+	const ctx = {
+		waitUntil: (p: Promise<unknown>) => {
+			pending.push(p);
+		},
+	} as unknown as ExecutionContext;
+	return { ctx, settled: () => Promise.all(pending) };
+}
+
+// Age the single cached entry past the 1h soft TTL. Reads the key back out of
+// the store rather than rebuilding it, so this doesn't pin CACHE_VERSION.
+function makeStale(store: Map<string, string>): void {
+	const [key] = [...store.keys()];
+	const entry = JSON.parse(store.get(key) as string) as { fetched_at: number };
+	entry.fetched_at = Date.now() - 2 * 60 * 60 * 1000;
+	store.set(key, JSON.stringify(entry));
+}
+
+describe("getVideosCached", () => {
+	it("caches a clean fetch", async () => {
+		const { env, puts } = fakeEnv();
+		const videos = await getVideosCached(env, "youtube", "UC1", () =>
+			Promise.resolve(AIRED),
+		);
+		expect(videos).toEqual(AIRED);
+		expect(puts).toHaveLength(1);
+	});
+
+	it("serves a degraded fetch without caching it", async () => {
+		const { env, puts } = fakeEnv();
+		const videos = await getVideosCached(env, "youtube", "UC1", () =>
+			Promise.reject(new UncacheableVideos(VOD_DATED)),
+		);
+		// Cold miss: better to show the VOD dates than an empty feed — but they
+		// must not be persisted, or they'd be served for the whole TTL.
+		expect(videos).toEqual(VOD_DATED);
+		expect(puts).toHaveLength(0);
+	});
+
+	it("keeps a stale entry's good dates when the refresh degrades", async () => {
+		const { env, store, puts } = fakeEnv();
+		await getVideosCached(env, "youtube", "UC1", () => Promise.resolve(AIRED));
+		makeStale(store);
+		puts.length = 0;
+
+		const { ctx, settled } = fakeCtx();
+		const served = await getVideosCached(
+			env,
+			"youtube",
+			"UC1",
+			() => Promise.reject(new UncacheableVideos(VOD_DATED)),
+			ctx,
+		);
+		await settled();
+
+		expect(served).toEqual(AIRED);
+		expect(puts).toHaveLength(0);
+		const [key] = [...store.keys()];
+		expect(
+			(JSON.parse(store.get(key) as string) as { videos: Video[] }).videos,
+		).toEqual(AIRED);
+	});
+
+	it("still swallows a hard fetch failure on a cold miss", async () => {
+		const { env, puts } = fakeEnv();
+		const videos = await getVideosCached(env, "youtube", "UC1", () =>
+			Promise.reject(new Error("youtube feed responded 503")),
+		);
+		expect(videos).toEqual([]);
+		expect(puts).toHaveLength(0);
+	});
+});
--- a/cloud/src/video/cache.ts
+++ b/cloud/src/video/cache.ts
@@ -22,7 +22,10 @@ import type { Video, VideoEnv, VideoProvider } from "./types";
 // v4: playlist fetches now dedupe repeated video ids (a curator re-adding a
 //     video); orphan the v3 entries so a warmed playlist doesn't keep serving a
 //     duplicate-containing list that crashes the Videos tab's keyed {#each}.
-const CACHE_VERSION = 4;
+// v5: live content is dated by when the broadcast aired rather than when its
+//     VOD was published; orphan the v4 entries so a warmed feed doesn't keep
+//     serving dates that are hours-to-a-day too recent.
+const CACHE_VERSION = 5;
 // Serve a cached entry without refetching under this age.
 const SOFT_TTL_MS = 60 * 60 * 1000; // 1h
 // KV hard expiry — a safety net far past the soft TTL.
--- a/cloud/src/video/cache.ts
+++ b/cloud/src/video/cache.ts
@@ -9,10 +9,16 @@
 // repopulates it, so a profile view never blocks on YouTube once warm; a cold
 // miss fetches synchronously. Transient upstream errors never overwrite a good
 // entry (fetchRecent throws rather than returning []) — worst case we serve
-// slightly-stale videos.
+// slightly-stale videos. A fetch that succeeded but degraded says so with
+// UncacheableVideos, and is served without being written (see refresh).
 
 import { logError } from "../log";
-import type { Video, VideoEnv, VideoProvider } from "./types";
+import {
+	UncacheableVideos,
+	type Video,
+	type VideoEnv,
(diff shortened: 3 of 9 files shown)


becked:
## Review

Correct fix for a correctly-diagnosed bug, with the cache invalidation and failure-mode handling a date-semantics change needs. Four findings below; the first is the only one I'd want fixed before merge.

**Verified locally on `d8aff96`:** `cloud/` `tsc --noEmit` clean, `youtube.test.ts` 28 passing. One note on the verification list in the description: `npm run lint` is `eslint src`, and `cloud/` has no eslint config or lint script — so it doesn't exercise any file in this PR. `format:check` (prettier over `.`) is the style gate that does.

### What's right

- **Absence-as-discriminator** in `parseVideosListPage` rather than a sentinel, with the never-aired-broadcast case covered by a test.
- **The degradation contract.** `fetchBroadcastStarts` never throwing, with per-batch independence, is the right call, and the comment explains it in terms of what enrichment *is* — an improvement to a date we already have — not just what it does.
- **`CACHE_VERSION` 4 → 5.** Without it, warmed entries serve wrong dates for up to the 24h hard TTL. Changelog entry is consistent with v3/v4.
- **Id dedup before batching**, tied back to the existing `dedupeById` rationale.
- Enrichment before the sort in both paths, with the ordering consequence spelled out.

---

### 1. `fetchYouTubeRecent` caps before enriching, and the stated reason doesn't hold

`cloud/src/video/youtube.ts:437-444`

```ts
// Cap on the feed's own order before enriching, so this costs one videos.list
// call rather than one per feed page; then re-sort, because re-dating a
// broadcast to when it aired can move it past a video published after it.
const recent = await withBroadcastStarts(
    parseYouTubeFeed(xml).slice(0, MAX_VIDEOS),
    env.YOUTUBE_API_KEY,
);
return recent.sort(byPublishedDesc);
```

There are no "feed pages" on this path — `fetchYouTubeRecent` does exactly one fetch of `/feeds/videos.xml?channel_id=…`, which returns ~15 entries and has no paging. 15 ids fit in one 50-id `VIDEOS_LIST_BATCH` just as 12 do, so the cap saves zero quota.

What it costs is correctness. The cap selects the top 12 **by VOD date**, discarding positions 13–15; enrichment then moves the survivors *earlier*. So a position-13 ordinary upload can end up newer than a re-dated cast that was kept:

| feed pos | video | `<published>` (feed order) | after enrichment |
|---|---|---|---|
| 5 | a cast | Jul 25 (VOD) | **Jul 24** (aired) |
| 13 | an upload | Jul 24 12:00 | Jul 24 12:00 — dropped by the cap |

If the upload's real date falls between the cast's air time and its VOD date, it belongs in the top 12 and isn't there. Skew is up to +15h in the sampled data, so that's the width of the window.

This is the exact failure `fetchYouTubePlaylistVideos`'s own comment warns against — "sort newest-first before capping … so the cap keeps the newest entries, not whichever happen to sit first" (`youtube.ts:483-488`) — and it leaves the two paths ordering their operations differently for no benefit. `fetchYouTubePlaylistVideosViaApi` already does enrich → sort → cap.

**Fix** — same quota, one fewer divergence:

```ts
const recent = await withBroadcastStarts(
    parseYouTubeFeed(xml),
    env.YOUTUBE_API_KEY,
);
return recent.sort(byPublishedDesc).slice(0, MAX_VIDEOS);
```

and drop the "one per feed page" clause from the comment — the remaining half ("re-sort, because re-dating a broadcast can move it past a video published after it") is the part that's true here.

---

### 2. `docs/api-reference.md` wasn't updated

CLAUDE.md lists it as authoritative for the Worker's HTTP API, and the module header in `youtube.ts` got exactly this correction. Two lines are now wrong on a keyed deployment:

**Line 318** (`GET /v1/users/:user_id/videos` → Notes):

> YouTube videos come from the unauthenticated channel RSS feed.

Suggested: *YouTube videos come from the unauthenticated channel RSS feed; where `YOUTUBE_API_KEY` is configured, one `videos.list` call per refresh re-dates live content to when the broadcast aired (the feed's `<published>` is when the VOD went up, hours later). Without the key the feed's own dates stand.*

**Line 334** (`GET /v1/creator-videos` → Notes):

> The cold build's per-channel fetches run in parallel over mostly-warm caches (~one RSS fetch at worst).

Suggested: *… (~one RSS fetch plus one `videos.list` call at worst).*

Unrelated and pre-existing, so not this PR's to fix — line 334 also says the feed is "Cached as one pre-assembled KV entry," which no longer matches `buildCreatorFeed` (D1 membership read per request, only per-channel entries cached). Worth a separate docs pass.

---

### 3. `byPublishedDesc` consolidates 2 of 4 copies

The extraction is right, but the same comparator is still inlined at `cloud/src/channels.ts:211` (`handleUserVideos`) and `cloud/src/channels.ts:253` (`mergeCreatorFeed`), and it can't be reused there because the new helper is module-private in `youtube.ts`. It's also platform-agnostic logic — it sorts `Video`, nothing YouTube-specific — living in the provider that `video/types.ts` describes as the platform-specific layer.

**Suggested:** move it to `cloud/src/video/types.ts` (which `channels.ts` already imports from) and export it, generic over `Video` so `CreatorVideo[]` sorts too:

```ts
// Newest first. ISO timestamps sort lexically; an entry missing a date ("")
// falls to the end. One definition for every path that promises "newest first".
export function byPublishedDesc(a: Video, b: Video): number {
    return a.published_at === b.published_at ? 0 : a.published_at < b.published_at ? 1 : -1;
}
```

Note the added `=== 0` case: the current form returns `-1` for both `compare(a, b)` and `compare(b, a)` when timestamps are equal, which is an inconsistent comparator. Harmless in practice (V8 won't crash; order among ties is just arbitrary) and pre-existing — but the extraction is the moment to fix it.

---

### 4. The structural cost, concretely

You asked for a second opinion on it. The degradation story is sound — no key ⇒ feed dates, same as before — but the quota enrichment spends isn't isolated to enrichment:

- **No single-flight on the SWR refresh.** `getVideosCached` schedules a `refresh` via `ctx.waitUntil` on *every* stale read, with no lock (`cloud/src/video/cache.ts:97-103`). That was free before; it's now 1 unit each. `/v1/users/:id/videos` returns via `jsonResponse` with no `s-maxage` (unlike the 60s edge cache on `/v1/creator-videos`), so N concurrent profile viewers inside one refresh window is N units.
- **Two hard-dependency paths draw on the same budget.** `resolveYouTube` returns 502 `RESOLVE_FAILED` on a quota error — a user can't add a channel. `fetchYouTubePlaylistVideosViaApi` throws on a 403, so a cold-miss tournament Videos tab caches `[]`. Only enrichment is designed to degrade.

Steady state is fine — a few hundred units/day for a couple dozen channels against 10k — so I wouldn't block on it. But "best-effort enrichment" and "resolution that hard-fails" now share one budget. Cheapest mitigation is observability: the `logWarn("youtube_videos_list_failed")` already fires on a 403, so it's alertable as-is; consider just noting in the module header that quota exhaustion surfaces first as `RESOLVE_FAILED`, not as a wrong date. A recency cutoff (skip the call when nothing in the feed is recent enough for a 15h skew to change its label) would zero out dormant channels, but it's a special case for quota — I'd hold it until quota is actually a problem.

---

### Smaller notes

- **Batches are sequential** — `youtube.ts:383`. The comment says each batch is independent, but the loop awaits each in turn. On a 500-video playlist cold miss that's 10 sequential round trips after the 10 paging fetches, on a request a user is waiting on. `Promise.all` over the slices matches the stated independence and halves cold-miss latency; subrequest limits aren't a concern at 20.
- **Log naming** — `youtube.ts:392,402`. `youtube_videos_list_failed` is emitted at both `warn` (with `yt_status`/`yt_detail`) and `error` (with `error_class`/`error_message`), so a consumer filtering on that event name gets two shapes. The two sibling upstream-failure sites in this file (`youtube_channel_resolve_failed`, `youtube_playlist_api_failed`) both use `logError(event, null, { yt_status, yt_detail })`. Warn is defensible here since this one is non-fatal — but then consider warning on the exception branch too, or splitting the event names.
- **Test coverage** matches the repo — there's no fetch-stubbing idiom anywhere in `cloud/`, so pure-function tests are the consistent choice. The untested surface is the batch loop and the never-throws contract, which is exactly what the "worth a staging check" caveat names.

Issue #183, opened 2026-08-01
Lint cloud/ — the Worker source is not covered by npm run lint


becked:
## What

`npm run lint` is `eslint src` (root `package.json`), so it covers the SvelteKit app only. Nothing lints `cloud/` — neither `cloud/src/` (the Worker) nor `cloud/test/`. `cloud/package.json` has `test` and `typecheck` scripts but no `lint`, and no CI job runs lint or tests (`.github/workflows/` has only `claude.yml`).

## Why it matters

`tsc` is the only automated check on the Worker, and `cloud/tsconfig.json` sets neither `noUnusedLocals` nor `noUnusedParameters` — so unused imports, unused params and the rest of the lint surface pass silently.

Concrete instance: `dee2cef` removed the last two `setLogField` call sites from `cloud/src/games.ts`, leaving the import unused. `npx tsc --noEmit` was clean and `npm run lint` was clean. It was caught by hand.

## What a fix involves

Pointing the existing config at `cloud/` doesn't work as-is — all 130 files fail to parse:

```
Parsing error: "parserOptions.project" has been provided for @typescript-eslint/parser.
The file was not found in any of the provided project(s): cloud/src/admin.ts
```

`eslint.config.js` sets `parserOptions.project: "./tsconfig.json"` (the SvelteKit tsconfig, which doesn't include `cloud/`) and `globals: { ...globals.browser }`. So the Worker needs its own flat-config block:

- `files: ["cloud/**/*.ts"]` with `parserOptions.project` covering `cloud/tsconfig.json` and `cloud/tsconfig.test.json`.
- Worker globals rather than browser ones. `@cloudflare/workers-types` are ambient TS types, not ESLint globals, so `ExecutionContext` / `ExportedHandler` / `ScheduledController` / `Tracing` read as `no-undef` — the usual answer is to turn `no-undef` off for TS files, since the compiler already catches undefined identifiers.
- A `lint` script in `cloud/package.json`, and/or widening the root one.

Unknown until the config parses: how many findings the existing Worker code has. Worth landing the config first and triaging separately from any auto-fix.

Adjacent, same gap: `cloud/tsconfig.json` could set `noUnusedLocals`/`noUnusedParameters` and catch part of this without ESLint at all.

PR #185, opened 2026-08-01
perf(cloud): per-request D1/R2 timing on the access log, plus Workers Logs, OTLP export and tracing

becked:
Instruments the Worker so issue #150 can be answered with numbers instead of by reading code, and wires the two sinks those numbers go to. Nothing here changes app behaviour — it is measurement plus config.

The premise: latency on the anonymous read path is dominated by cross-region round trips. D1 and R2 both live in ENAM, the Worker runs at the colo nearest the viewer, so a reader in Sydney pays a transpacific hop *per query*, not per request. Every candidate fix in #150 trades accuracy or complexity for one of those hops, and the ranking so far comes from reading code rather than measuring.

## What lands on the access log

`d1_ms`, `d1_queries`, `d1_wall_ms`, `d1_events_ms`, `d1_events_queries`, `r2_ms`, `colo`, `blob_cache`.

`d1_ms` is the sum of per-query in-flight time and MAY exceed `duration_ms` — two concurrent 60ms queries sum to 120ms inside a ~60ms request. `d1_wall_ms` is the union of the busy intervals. The gap between them is how much overlap a request actually achieved, which is exactly what a parallelisation fix would move and what a sum-only metric would score as a no-op.

The `events` subset is broken out because the per-IP rate-limit `COUNT(*)` is the most repeated blocking hop in the app, and a baseline that can't isolate it can't score removing it.

## How the coverage is structural

Both D1 handles are wrapped in `routeEnv` — the one place both are derived — so every query a handler issues is timed with no call site changing. `instrumentD1` is generic and type-preserving on purpose: typing it `(db: QueryableD1) => D1Database` would compile and would quietly delete the barrier that stops `countEventsSince` accepting a replica session. Proxies rather than object literals, because D1's methods live on a host prototype and a literal would silently drop `getBookmark()`/`withSession()`/`exec()`.

`readBlob` takes the `blob_cache` tag as a required parameter and sets the field itself. The field is only meaningful if its *absence* means "this route read no blob at all", and that held by convention until a third caller was added without it — so it's a compile error now rather than something review has to catch.

Only I/O is measurable this way. A Worker's clock advances on I/O, not CPU, so CPU time cannot be derived from these timestamps at all.

## Sinks

`[observability]` in `cloud/wrangler.toml`: Workers Logs for a 7-day queryable window, OTLP export to Honeycomb for the durable archive, and automatic tracing. Both `persist = true` explicitly, because Cloudflare's own OTLP example sets it `false` — copying that would silently delete the hot window.

`dispatch()` wraps each handler in `ctx.tracing.enterSpan(r.route, …)` and sets `route` on the span. Root spans carry `url.path`, whose cardinality is one cell per game id, and there is no `http.route` attribute — so without this, "p95 by route × colo" isn't a query. The guard next to it exists because `ExecutionContext.tracing` is declared non-optional but is absent on the context `createExecutionContext()` returns; unguarded, an absent `tracing` would 500 every route. Tracing is a measurement and must not be able to take the app down.

**Two things need doing outside this repo before any of it flows**, both silent on failure: the `per-ankh-logs` and `per-ankh-traces` destinations have to exist in the Cloudflare dashboard, and the config only takes effect on the next Worker deploy. Runbook in `docs/cloud-deploy-plan.md` §6.1.2.

## Read the doc as researched, not observed

`docs/cloud-deploy-plan.md` §6.1 is new and carries a status banner saying so: none of the `[observability]` config has ever run, and claims about what Cloudflare indexes and what a record looks like on arrival come from vendor documentation — which has already proved wrong here once. §6.1.5 is a ten-minute first-deploy checklist that converts it to observed, and it wants running *before* the multi-week measurement window, not after. Two claims are settled from workerd source rather than docs: `db.query.text` excludes bound values (which matters, because spans bypass `scrubPii` and go to a third party), and `ctx.tracing.enterSpan` works at our compatibility date.

§6.1.1 records why Logpush → R2 was abandoned — job creation 403s with a zero quota against an empty job list on this account — so nobody retries it from scratch.

## Verification

- `cloud/` `tsc --noEmit`: clean
- `npm run lint`, prettier: clean
- `cloud/` `npm test`: 566 passing (525 baseline + 41 new)

The 4 `canonical-map-options` failures are the pre-existing DOTA schema drift on `main` (#115/#141), untouched here.

**Not verified:** whether the platform's D1/R2 spans actually nest under the route span. Nothing in the test harness exports spans, so the tests pin our half — span named with the normalized route, attribute matching the access line, no span on unmatched paths or preflights, and every D1 round trip issued inside the callback, which is the precondition nesting rests on. The rest is §6.1.5 item 5.

Also unverified until it runs once: that the access line arrives as indexed fields rather than one opaque `message` string. `emit()` calls `console.log(JSON.stringify(...))`, and Cloudflare's structured-logging examples pass an object. If it lands as a string, the fix is small but the data collected until then is lost — hence checklist item 1.

## Adjacent

`r2_ms` covers reads that finish inside the Worker. The save-download routes hand `obj.body` straight to the client, so the bytes cross after `emitAccessLog` has fired and there is no moment at which the Worker knows the transfer finished — those routes log `r2_ms: 0` while moving the largest objects in the app. Their `bucket.get()` shows up as an auto-instrumented R2 span instead. Scoped in the field's comment rather than papered over.

`cloud/` is not covered by `npm run lint` — filed as #183. An unused import in `cloud/src/games.ts` survived both `tsc` and lint in this branch and was caught by hand.

https://claude.ai/code/session_0134gU4vL8682JAgdH5g5fDb


becked:
## Review

Read the full diff, traced the surrounding code, ran the checks.

**Verified locally on `perf/d1-instrumentation`:**

- `cloud/` `tsc --noEmit` — clean.
- `cloud/` `npm test` — 566 passed, 4 failed; all four in `canonical-map-options` (known DOTA drift, #115/#141). Matches the PR body.
- The `@ts-expect-error` pin in `d1.test.ts` is real: `instrumentD1(QueryableD1)` does not assign to `D1Database`.

One scope caveat on those two gates: `cloud/tsconfig.json` has `include: ["src/**/*.ts"]`, so `cloud/test/**` is **not** typechecked, and per #183 root `lint` is `eslint src`, so `cloud/` isn't linted at all. The two new files under `cloud/test/integration/` have neither gate — same hole that let the unused `games.ts` import through.

**What's right**, briefly, so it doesn't get lost below: proxy-over-literal with `RAW_STATEMENT` unwrapping before `batch()` and `forward()` binding to the real receiver is correct on all three counts (`Reflect.get(target, prop, target)` handles prototype-resident getters properly). `mergeBusyMs` is correct — sort by start, extend on overlap *and* adjacency, flush on gap, non-mutating, order-independent, and all six cases are pinned. Count-at-issue / time-at-settle, and refusing to invent an end for an in-flight query, is the right call. `timing` as its own `LogContext` slot rather than a `fields` entry (which is spread verbatim) has a test pinning it. Test placement matches the vitest projects exactly.

Five things below. (1) and (2) are worth settling before deploy; (3)–(5) are "the field says less than its name promises" — cheap now, expensive to discover three weeks into a measurement window.

---

### 1. `enterSpan` vs `startActiveSpan` — the wrapper's whole value rests on which one this is

`cloud/src/index.ts:959`

`node_modules/@cloudflare/workers-types/index.d.ts:4089` declares **both** `enterSpan` and `startActiveSpan` with identical signatures, and `Span` has an explicit `end()`. That pairing exists precisely because the two differ in when the span closes. The callback here returns a *pending promise* immediately:

```ts
return tracing.enterSpan(r.route, (span) => {
	span.setAttribute("route", r.route);
	return r.handler(request, routeEnv(env, r), m, ctx);
});
```

If `enterSpan` closes the span when the callback *returns* rather than when its promise settles, the wrapper span has ~0 duration and nothing nests under it. That takes out **both** halves of the stated goal — "p95 by `route` × `colo` is a query over this span alone" needs the duration as much as the nesting — not just the nesting half §6.1.5 item 5 defers.

Resolution: use the form that is correct under either semantics, rather than deferring to first-deploy.

```ts
return tracing.startActiveSpan(r.route, (span) => {
	span.setAttribute("route", r.route);
	return r.handler(request, routeEnv(env, r), m, ctx).finally(() => span.end());
});
```

`r.handler` returns `Promise<Response>` on every route in the table, so the `.finally()` is total. The stub in `cloud/test/integration/tracing.test.ts` already models the settle-then-record shape, so it needs only the method rename plus an `end()` recording assertion; test 6 (`d1AtEnter` / `d1AtSettle`) keeps its meaning unchanged.

If you'd rather keep `enterSpan`, then §6.1.5 item 5 needs to check the wrapper span's **duration** as well as nesting — a 0ms span with a correct `route` attribute is the silent failure here.

### 2. §6.1.4's privacy claim about spans is factually wrong

`docs/cloud-deploy-plan.md:615`

> What does still reach the provider unfiltered is the request URL — `url.path`/`url.full` on the fetch handler span, carrying game and user ids. That is the same data the access line's `path` field already exports, so tracing widens the path, not the field set.

`cloud/src/log.ts:153` sets `path: url.pathname`. The access line has **never** carried the query string, so `url.full` widens the field set, and `scrubPii` sees none of it. Concretely new to Honeycomb:

- `q=` on `GET /v1/users/search` (`cloud/src/users.ts:69`) and `GET /v1/games` (`cloud/src/games.ts:1996`) — user-typed, and on the users search it is frequently a username. `username` is in `PII_KEYS` (`cloud/src/log.ts:36-46`).
- `discord_id` / `username` on `GET /v1/auth/dev/login` (`cloud/src/auth.ts:643-645`). Dark in prod (`!env.DEV_LOGIN || isSecureRequest(request)`), but a crafted request is still spanned before the 404.

Given the root guardrail is "PII never leaves its lane and is never logged," this wants a corrected sentence in §6.1.4 and an explicit decision on the search terms, not a claim that nothing changes. If the answer is "acceptable," say so there — the section's framing is "read as researched," and this is the one place it asserts something reassuring that the code contradicts.

### 3. `r2_ms` excludes R2 writes, which *are* measurable

`cloud/src/log.ts:95-111`, `cloud/src/games.ts:1718-1737`

`handleGameUpload` awaits two parallel puts inside the handler — the parsed blob and the raw ZIP:

```ts
await Promise.all([
	env.SHARE_BUCKET.put(blobKey, compressedData, { … }),
	env.SHARE_BUCKET.put(zipKey, rawZip, { … }),
]);
```

Those complete inside the Worker and are exactly as measurable as `readBlob`. So `POST /v1/games` — the heaviest R2 path in the app — logs `r2_ms: 0`. The field comment scopes itself to "reads that complete inside the Worker"; the *name* doesn't, and anyone aggregating `p95 r2_ms by route` will read upload as doing no R2 work.

Either resolution works, but pick one:

- **Include them** — wrap the `Promise.all` in `beginR2Read()` (it already returns a closer; `const end = beginR2Read(); try { … } finally { end() }`), and widen the `StorageTiming.r2Ms` comment from "reads" to "reads and writes that complete inside the Worker."
- **Rename** to `r2_read_ms` / `readMs`, so the name carries the scope the comment does.

### 4. `blob_cache` absence and `r2_ms: 0` fail in the same direction on the same routes

`cloud/src/blob-cache.ts:75-80`

`blob_cache IS NULL AND r2_ms = 0` reads as "this request did no R2 work" — and is also exactly what the three streamed `saves/*` routes (`cloud/src/games.ts:2864`, `cloud/src/games.ts:3082`, `cloud/src/share-legacy.ts:409`) and the upload route emit. Nothing on the access line separates "touched no object" from "moved the largest object in the app."

The required-tag parameter is a good move; the comment above it just overstates what it buys:

> blob_cache is only meaningful if its *absence* means "this route read no blob at all"

Absence means *no `games/*` blob read through the cache tier* — the three direct `bucket.get()` sites above are blob reads that emit no tag. Two changes resolve it:

- Tighten that comment to say `games/*` blob, so the invariant it states is the one that holds.
- Give the streamed routes a discriminator so the query is expressible — e.g. `setLogField("r2_op", "streamed")` at those three `bucket.get()` sites, or fold them into a `beginR2Read()`-tagged helper that logs the `get` portion and marks the body as unmeasurable.

### 5. `d1.ts` overstates the coverage it establishes

`cloud/src/d1.ts:87`

> Both handles are wrapped in `routeEnv`, so every query the Worker issues is timed and counted onto the access log

`SECURITY_DB` is never wrapped, and `emitSecurityEvent` runs *after* `emitAccessLog` via `ctx.waitUntil` (`cloud/src/index.ts:1023-1028`), so its inserts can't land on the line even in principle. The cron sweep is called out as out of scope a few lines down; the security tee isn't.

Resolution is one word of precision — "every query **a handler issues through `SHARE_DB` or `EVENTS_DB`** is timed and counted" — plus a clause noting `SECURITY_DB` is deliberately outside it because the tee fires after the access line. Worth fixing because the entire value of the block is that the numbers are trusted to be complete.

---

### Minor

- `cloud/test/integration/games/storage-timing.test.ts:73` asserts `d1_queries === 3` exactly. Good tripwire, but it is precisely the test the #150 fix will break by design — a line saying "changed the query plan? update this number, don't assume a regression" turns a confusing failure into a decision.
- §6.1.2's destination table has body cells wider than its header/separator rows. Renders fine in GFM, and prettier is off for `*.md`, so nothing will quietly fix it.
- `warnTracingUnavailable`'s isolate-global flag couples the no-tracing test to file ordering. The test comment covers it and the assertion fails loudly rather than passing quietly — fine as-is, noting it only so it isn't rediscovered.

becked:
## Review findings addressed — 4391501

All five, plus the two actionable minors.

### 1. `enterSpan` → `startActiveSpan`

`cloud/src/index.ts`. The wrapper now uses the form that is correct under either semantics:

```ts
return tracing.startActiveSpan(r.route, (span) => {
	span.setAttribute("route", r.route);
	return r.handler(request, routeEnv(env, r), m, ctx).finally(() => span.end());
});
```

Two riders came with it.

**The guard checks the method, not just the object.** `startActiveSpan` is newer than our `compatibility_date` (`2024-12-01`), and the old guard only asked whether `ctx.tracing` existed — a missing *method* would have thrown and turned every route into a 500, which is exactly the failure the guard was written to prevent. It is now `!tracing || typeof tracing.startActiveSpan !== "function"`, so the worst case is untraced plus a `tracing_unavailable` warn. What I could establish about availability: workerd `1.20260722.1` carries both methods on the same `Tracing` interface, forwards to both from its own `cloudflare-internal:tracing-helpers`, and defines no `enable_*observability*` flag to gate either — and §6.1.4's existing finding that `enterSpan` works at our date means the object is exposed to us ungated. So they are expected to arrive together, and the guard is what makes "expected" safe.

**The test suite can't confirm it, and the doc no longer implies it can.** `tracing.test.ts` drives a hand-written stand-in (`createExecutionContext()` supplies no `tracing`), so a rename there proves only that our code calls what the stub records. §6.1's "two claims are observed" note now says plainly that `enterSpan` was exercised at that date and `startActiveSpan` was not, and §6.1.5 item 5 became three ordered checks: are there wrapper spans at all (with the warn as the tell), is `route` groupable, then duration **and** nesting — since a 0ms span with a correct attribute is the silent failure.

The stub records `ends` and `d1AtEnd`, and a new test pins `ends === 1` with `d1AtEnd === d1AtSettle`. An `end()` that fired on callback return would record `d1AtEnd` equal to `d1AtEnter`, so the regression this whole change is about is now a test failure rather than a first-deploy discovery. The untraced test drives both guard limbs — `tracing` absent, then a `tracing` carrying only `enterSpan` (which throws if dispatch ever reaches for it) — still asserting exactly one warn across the pair, so the isolate-global flag stays honest.

### 2. §6.1.4's privacy claim

Corrected, and the decision recorded rather than the claim softened. `url.full` carries the query string; `path: url.pathname` never has. Named concretely: `q=` on `GET /v1/users/search` and `GET /v1/games` (user-typed, frequently a username on the former — a value `PII_KEYS` redacts in a log field), and `discord_id`/`username` on `GET /v1/auth/dev/login` (dark in prod behind `!env.DEV_LOGIN || isSecureRequest`, but spanned before the 404).

**This is accepted, not mitigated** — the attribute is set by the platform on a span this repo doesn't create, so the only alternative is not tracing at all. Both `CLAUDE.md` files drop the "PII is never logged" absolute, since with tracing on it is false; the root one keeps the lane rule, and `cloud/src/CLAUDE.md` keeps the enforceable version ("never put PII in a log field — the deny-list is the last line of defense, not the first"). The §6.1 third-party bullet points at §6.1.4 for the detail.

### 3. R2 writes

Included, and the scope taken past the two puts you named: every R2 operation in the Worker is awaited in-handler, so limiting it to the upload would have left `r2_ms` meaning "reads and *some* writes" — the same defect one layer down. Wrapped: the parallel upload puts, the orphan-cleanup deletes, and the delete-game deletes.

Each `Promise.all` opens **one** window rather than one per operation, so concurrent work can't sum past `duration_ms` (the D1-side overlap problem doesn't get recreated here). `beginR2Read` → `beginR2Op`, since it no longer times only reads. `StorageTiming.r2Ms` documents both exclusions as decisions: the streamed downloads (unmeasurable by construction) and the legacy share routes (frozen, pending decom — left alone rather than wired up and then deleted).

### 4. The discriminator

`setLogField("r2_op", "streamed")` on `GET /v1/games/:id/download` and its admin twin, with an integration test asserting `r2_op === "streamed"`, `blob_cache` undefined, `r2_ms === 0` — sitting next to the existing "route that reads no blob" test, which now also asserts `r2_op` is absent. The two lines are otherwise identical on every R2 field and mean opposite things; that's the pair the tag exists to separate.

**One correction to the finding:** only two of the three sites are streamed. `share-legacy.ts:409` does `await object.arrayBuffer()` and decompresses in the Worker, streaming the *decompressed* bytes — its R2 read completes in-handler and is measurable. It's excluded here because legacy share is slated for decom, not because it can't be measured.

The `blob_cache` comment is scoped to `games/*` blobs rather than "no blob at all", which is the invariant that actually holds — the legacy read is a bucket-root `{shareId}.json.gz`, so the scoped version stays true with legacy untouched.

### 5. The coverage claim

Scoped in `cloud/src/d1.ts`, and in `cloud/src/index.ts:884`, which carried the same overstatement — fixing one would have left the other contradicting it. Both now say "every query **a handler issues through `SHARE_DB` or `EVENTS_DB`**", with `SECURITY_DB` called out as deliberately outside: the tee fires after `emitAccessLog` via `ctx.waitUntil`, so its inserts couldn't land on the line even if the handle were wrapped.

### Minors

The `d1_queries === 3` assertion carries a note that the #150 fix will change it by design — a failure there means update the count, not hunt a regression. §6.1.2's table cells are aligned. `warnTracingUnavailable`'s isolate-global flag left as-is per your note.

### Verification

- `cloud/` `tsc --noEmit`: clean
- `cloud/` `npm test`: 568 passed (566 baseline + 2 new), 4 failed — all four the known `canonical-map-options` DOTA drift (#115/#141)
- prettier: clean; root `npm run lint`: clean (still doesn't reach `cloud/`, #183)

**Two gaps worth stating rather than leaving to be found:**

- **The upload's write timing has no integration assertion.** Miniflare's R2 is local, so durations round to 0 — the same reason the existing blob-cache test pins the definition rather than a measured difference. An `r2_ms > 0` assertion on upload would be flaky. The accumulator is covered by the unit test on `beginR2Op`; the wiring is covered by review.
- **On #183:** `cloud/tsconfig.test.json` already exists and includes `test/**` — the gate is written, just not wired to any npm script (`typecheck` runs the default config, which is `src`-only). Running it by hand surfaces 3 pre-existing type errors in `test/integration/tournament/{match-numbers,public-reads}.test.ts`. Both files touched here are clean under it. Left alone as out of scope.

PR #188, opened 2026-08-03
feat(tournament): late pairing — add a match to an open Swiss round

alcaras:
Adds the one operation the substitution workflow was missing: pairing a late entrant into the **still-open Swiss round**, through a real endpoint with real guards — instead of a hand-written D1 INSERT.

## Why

The platform already supports mid-tournament substitution end-to-end: withdraw (`POST …/withdraw`), occupant substitution (`PATCH …/slots/:slot_id`, "edit slot (substitution)"), reinstate (`DELETE …/withdraw`), with occupant snapshots keeping history honest. But a slot reinstated *after* its division's current round was paired sits idle until the next round generates — a full round behind. The schema comment rejecting match-level slot swaps names the gap precisely: they'd "leave the displaced slot without a make-up game." This PR is the make-up game.

Concrete case: the 2026 Community Tournament wants to swap two substitutes into New World mid-Round-2 and have them play each other as their Round 2, rather than entering Round 3 at 0-1 with only three rounds left.

## Endpoint

`POST /v1/tournaments/:id/rounds/:round_id/matches` `{ slot_a_id, slot_b_id }` → `201 { match }` (tournament admin; audited as `match_added`).

Everything about the match is derived, nothing hand-entered:

- **Map** comes from `assignMapsToPairings` — the same engine round generation uses, with the fresh-map rule over the full tournament history and a deterministic seed.
- `match_index` is the round's next; `match_number` is the tournament's next (shared `NEXT_MATCH_NUMBER_SQL`); pick order defaults to `slot_b`, matching generation.
- Reporting flows through the normal advance path: the round simply won't auto-close until the added match reports, and its result feeds the next round's pairing like any other.

Guards, each with its own code: Swiss phase only (`TOURNAMENT_LOCKED`) — the championship bracket is never hand-edited; `ROUND_CLOSED`; `WRONG_DIVISION`; `SLOT_WITHDRAWN`; `SLOT_INACTIVE` (already advanced/eliminated, the pairing engine's own "active" definition via `computeRecord`); `ALREADY_PAIRED` (byes included); `SAME_SLOT`.

## UI

Admin view of the Swiss standings: rows that are active but **unpaired in the division's open round** — only a mid-round reinstate produces them, so the affordance is invisible everywhere else — get a **Pair** button beside Swap. Same popover+combobox picker as Swap (SwapPicker generalized into `SlotPickerPopover`, one implementation with labels/titles as props rather than a second copy), then a confirm dialog that spells out the round and that there is no un-add.

Screenshots (local fixture):

| Pair picker on the unpaired rows | Confirmation | Added |
|---|---|---|
| picker | confirm | done |

## Tests

`cloud/test/integration/tournament/add-round-match.test.ts` pins the full arc — withdraw both → round 2 generates without them → reinstate → add the catch-up match — and asserts: the added match blocks round close until reported and then feeds round 3's pairing (both substitutes appear in R3); map script is fresh for both players; `match_index`/`match_number`/pick-order conventions; the `match_added` audit row; and every rejection code (`ROUND_CLOSED`, `SAME_SLOT`, `ALREADY_PAIRED`, `WRONG_DIVISION`, `SLOT_WITHDRAWN`, `ROUND_NOT_FOUND`, `NOT_TOURNAMENT_ADMIN`, `UNAUTHORIZED`).

Docs updated: `docs/tournament-rules.md` (Late pairing under Withdrawals; the "can admins change pairings?" answer) and `docs/api-reference.md`.

## Checks

`npm run lint`, `svelte-check` clean; `cloud npm test` 580 passing (the 4 `canonical-map-options` failures are the pre-existing local Reference-XML drift, untouched).


alcaras:
## Self-review

Full-diff review for correctness and fit, negatives verified against `main`. One real defect found and fixed (c4b04cb), plus two smaller items; the rest verified clean.

**Fixed in c4b04cb**

- **TOCTOU race on the INSERT (the one real bug).** The handler validated round status and slot eligibility, then ran a plain INSERT. In the window between, a concurrent *report* can close the round — and round-close synchronously generates the next round, which would pair these very slots (they're active and unpaired), leaving a zombie pending match in a complete round that blocks `transition-championship` and double-books both players. A concurrent *add* could likewise double-pair a slot. This is exactly the race family the repo already guards elsewhere: the claim-slot INSERT in `player.ts` collapses concurrent writers via `meta.changes === 0`. The INSERT is now `INSERT … SELECT … WHERE EXISTS (round still in_progress) AND NOT EXISTS (either slot already in a match this round)`, with `changes === 0` mapped back to a precise `ROUND_CLOSED` / `ALREADY_PAIRED`. One subtlety worth flagging for review: `NEXT_MATCH_NUMBER_SQL` embeds a bare `?`, and SQLite gives a bare `?` max-used-index + 1 — mixing it with `?N` params is only safe with every other parameter explicitly numbered (`?1`–`?6`, so the constant's `?` lands at 7). The first draft of the guard had a genuine `?`/`?1` index collision; the committed version numbers everything and the happy-path test exercises the full statement.
- **Phase gate now tested**: `TOURNAMENT_LOCKED` (no builder phase reaches championship, so the test flips `tournaments.status` directly — the gate only reads that column).
- **Pair picker's `titleIneligible` was `""`** — the render branch guarantees eligibility so it never showed, but an empty required prop is a booby trap if the branch ever loosens. Real string now.

**Verified clean**

- **Bind order / derived fields**: `match_index` = round's `MAX+1` (generation uses loop position `i+1`, so no second copy of an existing fragment exists to extract — checked); `match_number` via the shared `NEXT_MATCH_NUMBER_SQL`; pick order to `slot_b` matching generation; occupant snapshots left NULL while pending, exactly like `buildSwissRoundStatements`' pending INSERT.
- **Map assignment**: same engine, full-history `matchRefs`, deterministic seed including slot ids (a second added match in the same round draws independently). Forfeited round-1 maps count as "played" for the substitutes — conservative and consistent with how generation would treat them.
- **`maybeAdvanceAfterMatchReport` downstream**: reporting the added match closes the round through the normal path; the `!downstream` guard means a late report into an already-advanced round can't spawn a duplicate next round.
- **Return shape** matches the sibling (`handlePatchMatchMap` also returns the raw `loadMatch` row); client method returns void + `invalidateAll`, per the repo's caller-refreshes idiom.
- **Audit/rate-limit**: rides `tournament_admin` (existing retention bucket — no `retention.ts` change needed); action logged after the committed write, so no audit ghosts on failed adds.
- **UI fit**: no second popover implementation — SwapPicker generalized into `SlotPickerPopover` with labels/titles as props, both actions on one implementation; `SwissStandings` has a single admin call site (the `[slug]` page — the matches-page `onSubstitute` surface doesn't render standings rows), so no unwired sibling; eligibility client-side mirrors the server gates but the server stays authoritative; runes-only, no conditional-read `$effect`s; confirm dialog because there is deliberately no un-add endpoint.
- **Docs**: soft-wrapped one-paragraph-per-line; rules doc's "can pairings be overridden?" answer updated to stay truthful; `pairing.ts`'s "no re-pair endpoint" comment amended rather than left to rot.

**Known limits (deliberate)**

- `SLOT_INACTIVE` has no dedicated test — reaching a 3-win/3-loss slot needs several reported rounds; the gate is one `computeRecord` call (the pairing engine's own active definition). The guarded-write's `changes === 0` path is likewise untestable without true interleaving; the sequential pre-checks shadow it.
- No un-add: intentional (no delete-match surface exists at all); the confirm dialog carries the warning.

Checks: `npm run lint`, `svelte-check` clean; `cloud npm test` 581 passing (the 4 `canonical-map-options` failures are the pre-existing local Reference-XML drift on `main`).

becked (commented):
## Overview

Adds `POST /v1/tournaments/:id/rounds/:round_id/matches` — an admin-only "late pairing" that inserts a catch-up match between two active, unpaired, same-division slots into an already-generated Swiss round. Backend handler + valibot schema + route, an `SlotPickerPopover` generalization of `SwapPicker`, a `Pair` affordance on eligible Swiss standings rows, an integration test, and doc updates.

**The design is sound and the implementation matches the engine's own conventions.** I found no correctness bug in the handler. Map assignment, `match_index`/`match_number`/pick-order derivation, the audit row, the rate-limit path, and the tournament-scoped slot load all line up with what round generation and the sibling admin handlers do. Security is clean: admin gate + rate limit via `authedTournament`, `loadSlotInTournament` prevents a cross-tournament slot splice from the body, nanoid regex on both the route and the body, no PII in the audit metadata, `staleTolerant` correctly omitted on a write route.

Verified specifics worth stating since they're easy to get wrong and this PR got them right:

- `COALESCE(MAX(match_index), 0) + 1` matches `buildSwissRoundStatements`' 1-based `i + 1` (`admin.ts:2926`), and there's no `UNIQUE(round_id, match_index)` to collide on.
- The `?1`–`?6` + bare-`?` numbering claim is correct — `NEXT_MATCH_NUMBER_SQL` (`admin.ts:2492`) carries a bare `?`, which SQLite assigns max-assigned + 1 = `?7`.
- The empty-map-pool `assignMap` throw isn't reachable here: the pool is append-only past `setup` (`admin.ts:750-764`) and start rejects empty.
- Byes are present in `data.matches`, so a bye recipient correctly lands in the client's `paired` set and gets no false `Pair` button.

## Issues

**1. Orphaned section banner — `cloud/src/tournament/admin.ts:1941`**

The new handler was inserted between the `PATCH /v1/tournaments/:id/matches/:match_id/map` banner comment and the function it documents. `handlePatchMatchMap` now sits at line 2145, ~200 lines below its own header. Move the new block below `handlePatchMatchMap`, or above that banner.

**2. Prettier regressions in four spots**

`npm run lint` is `eslint src` only — there's no prettier gate, so "lint clean" doesn't cover this. I confirmed `main`'s versions of all four files pass `prettier --check` and this branch's don't (tailwind class-order noise excluded):

- `+page.svelte:155` — `new Set(...)` over-wrapped
- `+page.svelte:630` — `async function addMatch(division, slotId, otherSlotId)` exceeds 80 cols
- `+page.svelte:1204` — `unpairedInOpenRound={...}` split mid-expression
- `add-round-match.test.ts:121` — `expect(...)` wrap

`npm run format` fixes all four.

**3. `SlotPickerCandidate` is exported with no consumer — `SlotPickerPopover.svelte:2`**

The rename promoted a plain local `interface SwapCandidate` into a new `<script module>` block and exported it, but nothing imports it — the only reference is `candidates: SlotPickerCandidate[]` inside the same component. `src/lib/tournament/CLAUDE.md` calls out "no exported helper or prop without a consumer." Keep it a local interface unless a call site actually types against it.

**4. `titleIneligible` is unreachable at the Pair call site — `SwissStandings.svelte`**

`{#if onAddMatch && isPairEligible(s)}` combined with `eligible={true}` means "Can't pair — already has a match in the open round" can never render. It's also an asymmetry worth deciding deliberately: Swap renders on every non-withdrawn row and explains itself when disabled; Pair vanishes. Either render Pair on the same rows Swap uses and let `eligible={isPairEligible(s)}` drive the title, or drop the dead prop for this site.

**5. Open-round detection fails closed when the round has no pending match — `+page.svelte:150-153`**

The comment says "A closed round means the division's Swiss is finished (a follow-up round would otherwise exist)." That isn't quite true — a round can be `in_progress` server-side with zero pending matches: a bye-only round, or a round generated with zero pairings after the division's field withdrew. In those cases the endpoint would accept the pair but no `Pair` button appears — which is the mass-withdrawal path this feature exists for. It fails safe (hides rather than errors), and the client genuinely has no rounds list (`+layout.ts` loads tournament/standings/bracket/matches only), so the heuristic is reasonable. At minimum soften the comment to say what it actually detects.

**6. The concurrency comment overstates what the `WHERE` buys — `admin.ts`**

The guard re-asserts round status atomically *at insert time*, but `maybeAdvanceAfterMatchReport` (`admin.ts:3051-3055`) reads its pending set and then batches the round close + next-round generation as separate statements. An INSERT landing in that window leaves a pending match in a `complete` round with round N+1 already generated without the pair. Sub-second interleaving of an admin add with the round's final report; the result still records, it just doesn't feed N+1 — recoverable, not corrupting. I'd soften "re-asserts both invariants atomically" rather than add machinery.

**7. Test coverage is two codes short of the claim**

The PR body says "every rejection code," but `SLOT_INACTIVE` (advanced/eliminated) and `SLOT_NOT_FOUND` aren't exercised. `SLOT_INACTIVE` is the one gate whose logic isn't self-evident — it routes through `computeRecord`'s active definition — so it's the one most worth pinning. Also untested: a *second* added match in the same round, which is what the slot-ids-in-the-seed decision exists for (`match_index` 3, independent draw).

## Minor

- **Map round-spread is silently dropped.** `assignMapsToPairings` gets a single pairing, so `usedScriptsThisRound` starts empty and the added match can reuse a script already used elsewhere in the round. The PRIMARY per-player fresh-map rule — the one that matters — still holds. Worth a caveat in the comment rather than a fix.
- **Route grouping.** The new route sits at the tail of the `// Slots` block; there's a match-mutation group at `index.ts:506-558`. The comment explains the choice, so this is taste.
- **Import order.** `handleAddRoundMatch` was inserted after `handlePatchTournament` in an otherwise-alphabetical block (`index.ts:113`).
- **`AddRoundMatchSchema` is byte-identical to `SwapSlotsSchema`.** Defensible as intent-documenting, just noting it.
- **`docs/tournament-implementation-notes.md`** (the as-built record) wasn't touched; rules + API reference were. Optional.

## Verification note

I did not run `cloud npm test`, `svelte-check`, or the test suite — my checkout is on `main` and I didn't fetch the PR branch. The prettier findings come from checking this branch's file contents fetched via `gh api`, against the repo's `.prettierrc`.

alcaras:
Thanks for the thorough pass — all seven applied in e4d260e, plus your minor notes where they had a concrete fix.

1. **Banner**: the handler block moved below `handlePatchMatchMap`; the banner and its function are adjacent again.
2. **Prettier**: all four spots (and every other touched file) now pass `prettier --check` — noted that `npm run lint` being eslint-only means my "lint clean" never covered this.
3. **`SlotPickerCandidate`**: back to a local interface; the `<script module>` block is gone.
4. **Pair visibility**: keeping hidden-until-relevant deliberately (a permanently-disabled Pair on every row of every tournament felt like noise for an affordance only a mid-round reinstate can use — alcaras concurs). `titleIneligible` is now optional with the contract documented ("may be omitted at a call site whose render branch already guarantees eligibility"), and the Pair site omits it rather than passing dead text.
5. **Open-round heuristic**: comment now says what it actually detects — no-pending-in-highest-round usually means Swiss is done, but can also be a rare still-open all-decided round (bye-only, mass-forfeit), and the client can't tell the two apart without a rounds list; Pair hides in both, failing safe. If the mass-withdrawal case ever bites a real TO, threading round ids/status into the layout load is the fix — deferred until it does.
6. **Concurrency comment**: softened to what the guard actually buys — it closes the check-to-insert window; the narrower window inside `maybeAdvanceAfterMatchReport`'s read-then-batch remains and is described as recoverable, exactly as you characterized it.
7. **Tests**: `SLOT_NOT_FOUND` added; `SLOT_INACTIVE` pinned by tightening `swiss_losses_to_eliminate` to 1 so the substitutes' forfeit loss makes `computeRecord` call them eliminated (no builder knob for Swiss config, so the test flips the column directly, same as the phase-gate test); and a second-added-match test (indexes 3 and 4, distinct match numbers). Writing that last one reproduced the reinstate-timing semantics the hard way: reinstating each slot before the last withdrawal's forfeit closed round 1 got them paired into round 2 normally — the test now withdraws all four before reinstating any, with a comment saying why.

Minor notes: map round-spread caveat added to the handler comment; import alphabetized; route stays in the slots/lifecycle block (the comment carries the rationale); `AddRoundMatchSchema` kept separate as intent-documentation per your read. `docs/tournament-implementation-notes.md` deliberately left untouched — it reads as a point-in-time session record, and the durable documentation lives in the rules doc + API reference this PR already updates; happy to add an entry if you'd rather it stay current.

Checks: lint / svelte-check / prettier clean; tournament integration suite 239 passing (all 30 files).

PR #193, opened 2026-08-05
feat(tournament): sesh export lists a match that outlived its sitting as owing its next part

alcaras:
Mini PR for a TO workflow gap in the sesh export.

## The gap

A pending match whose only scheduled sitting has passed **vanishes from the "Copy upcoming (sesh)" post entirely**: it isn't upcoming (nothing ahead), and it isn't "To be scheduled" (no part missing a time). In practice TOs were adding a blank part by hand after every unfinished session, just to keep the match visible in the schedule post.

But a pending match is by definition unfinished — when a session actually ends the game, the save gets uploaded and the match reported promptly. So a match whose every sitting is timed and aged out (started more than `LIVE_WINDOW_MS = 4h` ago) is simply *between sessions*, and it owes the schedule its next sitting.

## The change

`owesNextSitting` (new helper in `parts.ts`, reactive via the shared clock like its siblings) returns the part number such a match owes, or null when it doesn't: no sittings yet (the existing never-scheduled case), an explicitly open part (the TO already added one), or a sitting still ahead / plausibly live. The sesh export's To-be-scheduled block now lists the owed part as a synthetic split line.

Verified on a local fixture, clipboard contents via the real Copy button:

- sitting **8h ago**, match still pending →
  ```
  To be scheduled

  Match 012 (Part 2) - @A v @B
  ```
- same sitting moved to **1h ago** (inside the live window) → match stays out of both blocks, exactly as before.

No behavior change for never-scheduled matches, hand-added open parts, or matches with a future sitting. `lint` / `svelte-check` / prettier clean.

### DIFF
--- a/src/lib/tournament/parts.ts
+++ b/src/lib/tournament/parts.ts
@@ -184,6 +184,28 @@ export function partPlayed(part: TournamentMatchPart, now = nowMs()): boolean {
 // Reads the shared reactive clock (nowMs), like its sibling helpers here, so a
 // $derived that lists these (the Cast view, the popover's caster post) drops a
 // sitting as its time passes instead of waiting for the next data refetch.
+// A pending match whose every sitting is timed and has aged out (started
+// more than LIVE_WINDOW_MS ago) did not finish in its last session — a
+// finished match gets its save uploaded and reported promptly, so a still-
+// open one is between sittings and owes the schedule its next part. Returns
+// that next part's number, or null when the match doesn't owe one: no
+// sittings yet (a different "to be scheduled" case), an explicitly open
+// part (the TO already added the blank sitting by hand), or a sitting
+// still ahead or plausibly live. Reads the shared reactive clock like its
+// sibling helpers, so a match starts owing as its last sitting ages out.
+export function owesNextSitting(m: TournamentMatch): number | null {
+	if (m.status !== "pending" || m.slot_b_id == null) return null;
+	const parts = matchParts(m);
+	if (parts.length === 0) return null;
+	const cutoff = nowMs() - LIVE_WINDOW_MS;
+	for (const p of parts) {
+		if (p.scheduled_at == null) return null;
+		const t = Date.parse(p.scheduled_at);
+		if (Number.isNaN(t) || t > cutoff) return null;
+	}
+	return parts.length + 1;
+}
+
 export function upcomingScheduledParts(
 	matches: TournamentMatch[],
 	graceMs = 0,
--- a/src/lib/tournament/parts.ts
+++ b/src/lib/tournament/parts.ts
@@ -213,9 +213,8 @@ export function owesNextSitting(m: TournamentMatch): number | null {
 	if (parts.length === 0) return null;
 	const cutoff = nowMs() - LIVE_WINDOW_MS;
 	for (const p of parts) {
-		if (p.scheduled_at == null) return null;
-		const t = Date.parse(p.scheduled_at);
-		if (Number.isNaN(t) || t > cutoff) return null;
+		const t = partInstant(p);
+		if (t == null || t > cutoff) return null;
 	}
 	return parts.length + 1;
 }
--- a/src/lib/tournament/parts.ts
+++ b/src/lib/tournament/parts.ts
@@ -198,23 +198,21 @@ export function upcomingScheduledParts(
 	});
 }
 
-// A pending match whose every sitting is timed and has aged out (started
-// more than LIVE_WINDOW_MS ago) did not finish in its last session — a
-// finished match gets its save uploaded and reported promptly, so a still-
-// open one is between sittings and owes the schedule its next part. Returns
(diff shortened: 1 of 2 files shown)


alcaras:
## Self-review

Small diff, so the review is mostly negative results — checked against the surrounding schedule/parts machinery:

**Fixed (b962f6d):** the new helper had been inserted *between* `upcomingScheduledParts`' doc comment and the function itself, stranding the older comment above the wrong function. Placement only; no behavior change.

**Verified clean:**
- **Boundary consistency with the app's own definitions.** `owesNextSitting` uses the same `LIVE_WINDOW_MS` cutoff and the same strict-inequality side as `liveAndUpcoming` — a sitting at exactly `now − 4h` is "plausibly live" in both places, so a match leaves the Live list and starts owing its next sitting at the same instant, never both or neither.
- **Every early-out is a real case**: bye (`slot_b_id == null`), non-pending, no sittings (the existing never-scheduled branch), an explicitly open part (the TO's hand-added blank keeps working exactly as before), unparseable dates (treated as unknown → don't owe).
- **Reactivity**: reads `nowMs()` like its sibling helpers, so the sesh text recomputes as sittings age out — no stale copy from a tab left open.
- **No behavior change for any existing case** — the restructure only adds a fallback after the open-parts branch returns nothing; matches with a future or live sitting, never-scheduled matches, and hand-added blanks all take their old paths.
- **Championship matches** benefit too (nothing swiss-specific in the helper) — a pending bracket match that outlived its sitting also resurfaces, which is right.
- Verified live against a fixture through the real Copy button + clipboard read: 8h-old sitting → listed as "(Part 2)"; the same sitting at 1h → stays out.

**Known limit:** the Live/Calendar *views* still drop such matches entirely (their existing "aged out" rule) — this PR deliberately only fixes the sesh copy, where the TO workflow pain was. If the views should also show "between sessions" matches, that's a follow-up with its own design questions.

becked (commented):
Reviewed, and pushed four follow-ups to the branch. The change itself is right — the reasoning about a pending match being unfinished holds up (the upload path sets `status = 'complete'` in the same UPDATE that sets `game_id`, so there's no pending-with-a-save-uploaded state for the export to misread). These are all fit against what's already in `parts.ts`.

**`0508892` — read the aged-out boundary through `partPlayed`**

The loop was a second definition of the test `partPlayed` already owns, and the only reference to `LIVE_WINDOW_MS` outside `partPlayed`'s own body. `0326a14` had just collapsed the same duplication out of `liveAndUpcoming`, so this puts the new helper on the same footing. It also threads one `now` through the pass the way `liveAndUpcoming` does, so every part of a match is classified against the same instant instead of each re-reading the clock. Behaviour-preserving: `partPlayed` returns false for a part with no usable time, which is exactly what the explicit `t == null` check did.

**`a2056de` — bind the match once in the to-be-scheduled block**

The block now has three `seshMatchLine` call sites and each repeated `matchNumber`/`versus` verbatim while only the part tag differed. A local `line(partNumber, split)` closes over the match, so the branches read as `line(1, false)` / `line(partNumber, split)` / `line(next, true)`. Output unchanged.

**`2daa8f2` — an unparseable part time reads as an open part**

The one behaviour change. The two branches are complementary — a match is listed because it has an open part, or because the helper says it owes one — but they were asking different questions: the filter tested `scheduled_at` directly, the helper goes through `partInstant`. A part whose stored time is non-null and unparseable satisfied neither (not open, `scheduled_at` is set; not aged out, `partInstant` is null so `partPlayed` is false), so the match dropped out of the post entirely — the exact symptom this PR exists to fix. Both branches now read `partInstant`, which documents that unscheduled and unparseable collapse because no caller distinguishes them.

**`d93de7d` — `owesNextSitting` → `owedPartNumber`**

`parts.ts` keeps predicates and value-getters visibly apart (`hasStartedPart` / `isMatchScheduled` / `partPlayed` vs `partInstant` / `nextScheduledAt` / `scheduledParts`), and `isMatchScheduled` / `nextScheduledAt` are the same fact under both shapes. A predicate-shaped name returning `number | null` is why the call site immediately renamed it to `next`. To be clear, "sitting" was not the problem — `docs/tournament-rules.md` calls these "parts (sittings)" and `matches-table.ts` already exports `rowIsPendingSitting`; it was the shape.

`eslint`, `svelte-check` (0 errors / 0 warnings across 1303 files) and `prettier --check .` are clean.

---

One thing I left alone that's worth your read. `LIVE_WINDOW_MS` is 4h, and its own comment justifies that number by stream-link liveness: after the window the ephemeral `…/live` URL has died, so the row leaves the Live panel — a false positive there costs a disappeared row. Here the same constant drives an assertion in a TO's post. A sitting that runs past 4h gets listed as owing a Part N+1 while it's still being played, and an evening game reported the next morning claims an owed sitting all night. Still clearly better than the match vanishing, so I'm not holding anything on it — but the threshold is inherited rather than chosen for this job, and it may want its own constant later.

Issue #166, opened 2026-07-26
Typecheck cloud/test/ — 3 errors, one a silent coverage hole in tournament match numbering


becked:
## Problem

`cloud/tsconfig.json` has `"include": ["src/**/*.ts"]`, so `npm run typecheck` (`tsc --noEmit`) covers worker source and the unit tests beside it, but not `cloud/test/` — 51 files, ~11.7k lines, the whole Miniflare integration suite. Root `npm run lint` is `eslint src`, so `cloud/` isn't linted either. Nothing reads those files except Vitest at runtime.

That gap let a real defect through, fixed in 76b167a on `perf/d1-read-replication`: `test/integration/tournament/concurrent-advance.test.ts` called `maybeAdvanceAfterMatchReport(env, ...)` with the `cloudflare:test` env, which is just the `wrangler.toml` bindings. The handler's signature had gained a required `EVENTS_DB`, which `index.ts` derives per-request in `routeEnv` — so the test was passing an env shape production never produces. It didn't fail, because both call sites early-return before the audit write. A type error would have caught it immediately.

## Measured cost: 3 errors

Adding `"test/**/*.ts"` to the existing `include` is the entire config change needed. All 51 test files enter the program (verified with `--listFiles`), with no new `types` entry, no `@types/node`, and no path mapping. `tsc` goes 1.12s → 1.34s. Two things already in place make it that cheap: `test/env.d.ts` imports `@cloudflare/vitest-pool-workers/types`, which brings in the `cloudflare:test` module declaration; and nothing under `test/` imports `node:*` (only `vitest.config.mts` does, and it's outside `include`). So the deliberate "no `@types/node` in `cloud/`" constraint is unaffected.

**One real defect** — `test/integration/tournament/match-numbers.test.ts:69`:

```
error TS2322: Type '"mid-championship"' is not assignable to type 'TournamentPhase | undefined'.
```

`TournamentPhase` (`test/helpers/builders.ts:133`) is `"setup" | "swiss-round-1-generated" | "swiss-round-1-complete"`. `advanceTo` is only ever compared against the first two (`builders.ts:315`, `:326`), so `"mid-championship"` falls through both early returns onto the swiss-round-1-complete path. The test's own comment says *"swiss rounds + championship round(s) all generated by the real engine"* — it generates no championship rounds, so championship match numbering is unverified. The file passes green today (3/3).

[...]

PR #194, opened 2026-08-05
feat(game-detail): Orders tab — the action economy, charted and itemized

alcaras:
Adds an **Orders** tab to the game detail view — the action economy, charted and itemized. Sits after Techs; works for duels and FFAs; **no parser change** (every input already ships in today's blobs, so the whole existing corpus renders it).

## Why orders and legitimacy share a tab

They're mechanically one system: `Player.calculateNonCityYield` includes `getLegitimacy() × yield.xml ORDERS <iPerLegitimacy>` for **every** player — a tenth of an order per legitimacy point, per turn. In most games it's the single biggest orders line. So the tab pairs two charts (orders/turn and legitimacy, all players, nation colours, succession sawtooths visible) with end-of-game itemizations of both.

## The itemization, anchored on the game's calculation

**Orders/turn** (exact rows, from `bake-orders-sources.ts` walking source → `EffectPlayer` → `aiYieldRate` over `YIELDS_MULTIPLIER`):
- **Legitimacy** — final legitimacy × 0.1;
- **Difficulty** — the per-player MP handicap from `game_details.players` (The Able +10 … The Great +5);
- **Laws** (Serfdom +5) and **ruler traits** (Master Strategist +2, Trusted +1), active at end.

**Legitimacy** (`Player.getLegitimacy` / `Character.getLegitimacy`):
- **Dynasty cognomens** — each ruler's cognomen worth divided by reign recency (integer division; current ruler full, predecessor half): "Amata the Fountainhead +60", "Ifra the Noble 70÷3";
- **Ambitions** — flat `FINISHED_AMBITION_BONUS` (+10) per completion, named from the goal data;
- **Story events, by timing** — a jump not explained by that turn's ambitions, on a turn a story event fired for that player, is credited to the event by name: "Mysterious Sword + Character Warmongers +30", "Calamities What Of The Gods +12". Succession turns are skipped (the dynasty term reshuffles), sub-2 leftovers stay unattributed, and event names are normalised across their `P.N.`/family/religion prefixed copies.

Everything the blob can't reconstruct — council/court ratings (a good chancellor is real orders), cathedrals/shrines, agents, trade — lands in a **signed Other** row against the save's true totals, with a footnote naming the suspects. On the demo duel it's small on the legitimacy side (+2 for one player) and honest about the court on the orders side.

One correctness note from the game source, documented in the footnote: the save's yield-rate history records `calculateYield(eYield, false)` — **gross** production. Orders consumed by working/fortifying units feed only the in-game display, never the recorded series, so they are structurally absent from both the chart and the totals here.

## Layout

Duels render each metric as **one face-off table** — rows are the union of both players' sources ordered by the larger value, players in facing columns, "—" where only one side has the source. FFAs keep one card per player.

## Screenshots

duel page

Duel face-off tables:

orders table
legitimacy table

FFA (5 players, per-player cards):

ffa

## Checks

`npm run lint`, `svelte-check`, prettier clean. New bake step `bake:orders-sources` registered in `bake:all`; the generated module is frontend-only (no worker mirror needed). Old blobs degrade gracefully (a 2.6.1 FFA lacks cognomen data — those rows simply stay in Other).


alcaras:
## Self-review

Full-diff pass for correctness and fit; three findings fixed (2f300ae), one structural limitation documented rather than papered over, and the negatives verified against the game source and the blob.

**Fixed**

- **Duel rows failed to merge when ambition counts differed.** The ambition row's label carried its per-player count ("Ambitions finished (4)"), so the face-off table only merged both sides when the counts happened to be equal — the demo game masked this. Label is now constant; the count and the ambition names live in the detail line.
- **`type Pick` shadowed TypeScript's built-in `Pick<T, K>`** — legal, but a trap for the next editor. Renamed `PickFn`.
- **Two footnote omissions**: MP *advantage* compensation grants flat orders (+1…+10 by tier, `advantage.xml → EffectPlayer`) but the blob doesn't record it per player, so it lives in Other and the footnote now says so; and the cognomen-overlap note below.

**The structural limitation, stated plainly**

The legitimacy decomposition mixes two honest but overlapping views: the dynasty rows are **end-state levels** (each ruler's cognomen worth ÷ reign recency — exact arithmetic per `Character.getLegitimacy`), while the event rows are **timed flows** (series jumps credited to same-turn story events). A cognomen *earned mid-reign* is therefore counted twice — once in its ruler's row, once inside whatever jump it produced (possibly labelled with a coincident story event). Because Other is computed as `total − Σ rows` and is signed, the overlap nets out arithmetically — the total always reconciles — but an event row adjacent to a cognomen award can over-claim. Succession turns are already excluded (that's where *past* rulers' divisor-shifts land, so those don't double-count). Distinguishing "cognomen award" from "event reward" inside one jump needs the award turn, which the blob doesn't carry; if the parser ever picks up cognomen timestamps, this cleans up fully. Footnote on the Other row now explains it.

**Verified clean**

- **The coupling constant**: `ORDERS_PER_LEGITIMACY = 0.1` is `yield.xml` ORDERS `iPerLegitimacy(1)` over `Constants.YIELDS_MULTIPLIER(10)`, applied unconditionally in `Player.calculateNonCityYield` — not Schemer-gated (that flag governs the one-shot convert action, a different mechanism).
- **Gross-vs-net**: the save's rate history is written as `calculateYield(eYield, false)` — the unit toll feeds only the live display. The footnote states it; the tab's totals reconcile against what the save actually records.
- **The flat-orders source list is genuinely short** — swept every `effectPlayer` carrying ORDERS: 7 difficulties, 4 advantages, Serfdom, two traits, nothing from techs or nations. The bake's ≥5-sources gate would catch a parse drift.
- **Per-player difficulty** comes from `game_details.players` (present since old blobs; null-safe when absent).
- Bake registered in `bake:all`; generated module is frontend-only (no worker mirror, so no drift test needed — the Worker never computes this). Old blobs degrade to Other rather than erroring (verified on a 2.6.1 FFA).
- Fit: `formatEnum` everywhere a zType surfaces, `getNationChartColor` for series, `GOAL_NAMES` for ambitions, the LeaderCard `NAME_*` rule for characters; runes-only; both eslint-disables carry justifications; no new prop left unwired (storyEvents flows to both layouts).

**Known limits (deliberate):** council/court rating yields (the chancellor) are the largest un-itemizable orders term and the footnote leads with them; no tests (frontend tab code, matching its siblings — the bake carries the validation gates).

becked (commented):
Reviewed against the branch rebased onto current `main` (the only rebase conflict was `package.json`: `main` added `bake:family-colors` to the `bake:all` chain while this branch added `bake:orders-sources` — both belong). On that rebase, `npm run lint` and `npm run check` are clean (0 errors, 0 warnings, 1312 files). There's no frontend test harness (`vitest.config.mts` exists only in `cloud/`), so the absence of tests for `orders.ts` isn't a gap against this repo.

The domain modelling is the strong part — the arithmetic is anchored on real XML (`iValue 100 / YIELDS_MULTIPLIER 10 = 10` for `EFFECTPLAYER_DIFFICULTY_ABLE` checks out), `aiYieldRate` is correctly read as an array-of-int map rather than an AI-only field, and the signed remainder is the right call. The findings are in the wiring.

## Correctness

**1. `OrdersTab.svelte:223`, `:294` — negative rows render as `+−80`.** Both cells hardcode a leading `+`, but `fmt()` (`:177`) already emits its own `−` for negatives. Cognomen legitimacy is negative for `COGNOMEN_BAD: -80`, `BLOODY: -100`, `WEAK: -60`, `UNREADY: -40`, `UNFORTUNATE: -20` (`orders-sources.ts:36,38,85,86,90`), and `legitimacyEndBreakdown` only skips exact zero (`orders.ts:159`). The `Other` row two lines down already does it correctly (`:234`, `:313`).

**2. `OrdersTab.svelte:213` — duplicate keys throw `each_key_duplicate`.** The FFA table keys on `row.label`, and `legitimacyEndBreakdown` can emit the same label twice: an event name credited on two different turns (`orders.ts:223`), or two rulers producing the same `characterLabel`. Svelte throws in dev *and* prod (`svelte/src/internal/client/errors.js:136-146`). The duel branch escapes only because `unionRows` collapses by label into a Map — which means it silently merges two distinct events into one row instead.

**3. `orders.ts:63` — `characterLabel` drops the regnal numeral.** The comment claims "Same rendering rule as LeaderCard", but `LeaderCard.svelte:42-47` is `c.suffix > 1 ? \`${name} ${toRomanNumeral(c.suffix)}\` : name`, with a comment saying it exists to disambiguate rulers sharing a first name. Dropping `suffix` is both the divergence and the direct cause of the duplicate keys above. `toRomanNumeral` is already in `$lib/utils/formatting`.

**4. `OrdersTab.svelte:59, 74, 127, 130, 137, 141, 156, 166` — `p.player_id` instead of the resolved `p.playerId`.** `PlayerInfo.player_id` is optional by design ("Added in PARSER_VERSION 2.6.0; optional so the frontend can still read older blobs", `types/PlayerInfo.ts:5-12`); `resolveDetailPlayers` exists to supply a non-optional `playerId` (`helpers.ts:900-911`). On a pre-2.6.0 blob every `p.player_id` is `undefined`, so the yield/history lookups match nothing (empty charts, "no orders data"), the law and goal filters return empty — and `gameDetails.players.find(gp => gp.player_id === p.player_id)` at `:127` matches `undefined === undefined`, handing **every** player the first player's difficulty, legitimacy and ruler. Every sibling tab uses `p.playerId` (`LeadersTab:37`, `LawsTab:43`, `EconomyTab:395`, `WondersTab:60`, `SettingsTab:104`).

**5. `OrdersTab.svelte:50-51` — `colorOf` discards the resolved colour, breaking mirror matches.** `DetailPlayer` already carries `.color` (`helpers.ts:874`), and `resolvePlayers` deliberately gives two same-nation players *distinct* palette colours (`collides ? getChartColor(ordinal) : getNationChartColor(...)`, `helpers.ts:864-866`). Recomputing `getNationChartColor(p.nation, i)` gives both Assyrias the same series colour and the same card heading. `src/lib/game-detail/CLAUDE.md` calls out mirror matches specifically; `LeadersTab:164` handles it.

**6. `OrdersTab.svelte:126-128` — the `info` lookup is redundant.** `players` is `resolveDetailPlayers(gameDetails.players, effectiveRoster)` (`GameDetailView.svelte:194-196`) and `DetailPlayer = PlayerInfo & …`, so `p.legitimacy`, `p.difficulty` and `p.leader_character_xml_id` are already on `p`.

**7. `GameDetailView.svelte:470, 594` — the tab is ungated.** The Leaders tab gates both its trigger (`:462`) and content (`:548`) on `hasLeaders`, because its data only exists in newer blobs. Orders depends on the same `characters` / `player_goals` / `leader_character_xml_id` (2.8.0 / 2.11.0) and renders two empty charts plus cards whose entire total is "Other" on an old blob.

## Fit

**8. Both charts already exist.** Orders/turn renders from `YIELD_CHART_CONFIG`'s `YIELD_ORDERS` entry (`helpers.ts:173-178`) through `createYieldChartOption` (`helpers.ts:1111`, consumed by `YieldsTab:56` and `TechsTab:230`); legitimacy renders in `LeadersTab:122-174`. `lineChart()` (`OrdersTab:85-113`) is a third option shape for the same two series.

**9. `filledLineStyle` omitted.** `helpers.ts:1004-1008`: "Every line chart on the game detail page spreads this so they read as one family." `OrdersTab:104-111` hand-rolls `showSymbol: false, lineStyle: { width: 2 }` and drops the area fill.

**10. `OrdersTab.svelte:98` — hardcoded `#c4b998`.** Only occurrence in `src/`; the token is `--color-tan` (`app.css:34`). It also makes the legend visible, whereas `CHART_THEME` sets `legend.show: false` on purpose (`charts.ts:55-57`) with series visibility driven by the persisted `chartFilters` — and `PLAYER_CHART_KEYS` already contains both `"orders"` and `"legitimacy"` (`helpers.ts:119,126`). These charts opt out, so a player toggled off in Yields or Leaders reappears here.

**11. `findByPlayer` unused.** `helpers.ts:948-961` is the single-row id-match-with-nation-fallback helper for exactly the lookups at `OrdersTab:57-60, 73-74, 129-130, 136-138`.

**12. `orders.ts:197-209` reimplements the `EVENTSTORY_` normalisation.** `expeditionEvents` (`science-techs.ts:248-272`) already documents and implements it — and dedupes on `(event, turn)`, which orders.ts doesn't.

**13. Footnote strings copy-pasted.** The orders note appears verbatim at `OrdersTab:264` and `:353`; the legitimacy note at `:264` and `:367`.

**14. `LEGACY_AMBITION_LEGITIMACY` has no consumer** (`orders-sources.ts:26`, emitted at `bake-orders-sources.ts:211-213`). Relatedly, `takeFrom(techs, …)` / `takeFrom(nations, …)` (`bake:141-142`) currently produce no keys, and `ordersEndBreakdown` never looks up a tech or nation — if a patch adds one, it lands silently in Other. The generated header comment (`bake:200-203`) says "laws, techs, nations, and ruler traits", but the data is `DIFFICULTY_*`, `LAW_*`, `TRAIT_*`.

**15. `OrdersTab.svelte:284, 290` — `unionRows`/`valueFor` run in the template**, rebuilding the Map and scanning per cell on every render; sibling tabs put this in `$derived`.

becked:
Pushed six of the seven correctness findings as four commits on this branch (`daf64d7..1f4611f`) — fixed here rather than handed back as a list, so the diff is the review response. `npm run lint` and `npm run check` are clean on the result (0 errors, 0 warnings, 1312 files).

**`13d6ad0` — finding 3, the regnal numeral.** `characterLabel` appends `toRomanNumeral(c.suffix)` when `suffix > 1`, after `formatEnum` and before the cognomen, matching `LeaderCard.svelte:43-48`.

**`7be79bd` — finding 2, the duplicate `{#each}` key.** Fixed at the source rather than in the key, because keying alone would have left the duel table's silent merge in place: an event credited on several turns now accumulates into a single row, the jumps summed, every turn named in the detail line. This changes the arithmetic as well as the keys — a repeat that showed as rows of 6 and 4 is now one row of 10. With finding 3 in, ruler labels are unique too, so both producers of the collision are closed.

**`38932f5` — findings 4, 6 and 5, together.** All eight lookups move from `PlayerInfo.player_id` to the resolved `DetailPlayer.playerId`; the `gameDetails.players.find(...)` lookup is deleted outright, since `DetailPlayer` already carries `legitimacy`, `difficulty` and `leader_character_xml_id`; and the per-player colour is `p.color` instead of a recomputed `getNationChartColor(p.nation, i)`, so a mirror match draws its two sides apart again. Removing that lookup left the `gameDetails` prop with no consumer, so it's gone from the component and from the call site in `GameDetailView.svelte` — flagging that as the one edit beyond the findings as written.

**`1f4611f` — finding 1, the sign.** A `signed()` helper alongside `fmt()`, used by the source rows and the remainder in both tables; totals keep the unprefixed `fmt`. `+−100` for a ruler cognomened Bloody is gone, and the sign logic is in one place instead of three.

**Finding 7 — my reasoning there was wrong.** I attributed the empty charts to missing 2.8.0 data, but the charts read `allYields` and `playerHistory`, which older blobs have; what I saw was finding 4's symptom, and it clears with that fix. On a blob without `characters` the tab now degrades rather than empties: both charts render, orders keeps its Legitimacy/Difficulty/Laws rows and loses only the ruler-trait row, and legitimacy keeps ambitions and events while the dynasty cognomen rows fold into `Other`. Since gating the whole tab on `hasLeaders` would hide two working charts, my recommendation is to leave it ungated as it is — if the cognomen rows quietly folding into `Other` is the part that bothers us, the narrower move is to gate the itemization cards, not the tab.

The fit findings (8–15) are untouched — the charts duplicating `YIELD_CHART_CONFIG`/`LeadersTab`, `filledLineStyle`, the `#c4b998` literal and the legend opt-out, `findByPlayer`, the `EVENTSTORY_` normalisation, the duplicated footnotes, `LEGACY_AMBITION_LEGITIMACY`, and the template-time `unionRows`.

becked:
The eight fit findings are in as two commits (`ccfc7fb..89d6ea9`), same as before — fixed on the branch rather than handed back. `npm run lint` and `npm run check` are clean (0 errors, 0 warnings, 1312 files); `prettier --check` is clean on everything touched (`scripts/bake-family-colors.ts` warns, but it's untouched here and warns on `main` too).

**`89d6ea9` — findings 8–13 and 15.** Both charts now come from the builders that already drew them. Orders/turn is `YIELD_CHART_CONFIG`'s `YIELD_ORDERS` entry through `createYieldChartOption`; legitimacy was inline in `LeadersTab`, so it moved to `createLegitimacyChartOption` in `helpers.ts` and both tabs call it. `lineChart()` and the `#c4b998` legend are gone, which puts these two plots back under `CHART_THEME` and under the page's per-player selection — `GameDetailView` binds `chartFilters.orders` and `.legitimacy` in — and gets them `filledLineStyle` plus `ChartContainer`'s fullscreen, which the tab was also missing. In the same commit: the duel table's union walk and per-cell lookups move from the template into a `$derived` (retiring `PickFn` and its eslint-disable), the two footnotes become one const each, the remaining per-player reads go through `findByPlayer`/`ownedByPlayer`, and the `EVENTSTORY_` prefix strip becomes `storyEventType` in `helpers.ts`, shared with `expeditionEvents`.

**`ccfc7fb` — finding 14.** The `takeFrom(techs, …)` / `takeFrom(nations, …)` sweeps are dropped: they produced no keys, and since `ordersEndBreakdown` has no per-tech or per-nation lookup, a patch that added one would have landed silently in Other. `LEGACY_AMBITION_LEGITIMACY` is deleted — I checked whether it could gain a consumer instead, and it can't: `goal.xml` has no legacy flag across its 400 entries, and `PlayerGoalInfo.goal_state` is a serialized `<Stats>` map, so the blob can't distinguish a legacy ambition. The generated header now says what the data is (`DIFFICULTY_*`, `LAW_*`, `TRAIT_*`). Re-baked: same 9 sources, same values.

**Two amendments to the review as written.**

Finding 12's "dedupes on (event, turn), which orders.ts doesn't" isn't a behavioural gap — `orders.ts` accumulates into `Map<turn, Set<string>>`, so it dedupes inherently. The duplication was the prefix strip and its comment, and that's what the shared helper removes.

Finding 10's "a player toggled off in Yields or Leaders reappears here" can't happen today. The only writers of `chartFilters` are the three init effects in `GameDetailView` (everything true), nothing dispatches `legendselectchanged`, and every other chart sets `legend.show: false` — so the cross-tab selection is inert plumbing at the moment. Wiring it in is still right for consistency, and worth noting it cuts the other way too: the Orders legend was the page's only clickable series toggle, though `Chart.svelte`'s `setOption(…, true)` is notMerge, so a toggle there never survived an option rebuild. Also folded in: a chart with no data now hides rather than drawing empty axes, which is what the shared builders do — that applies to the Leaders tab's legitimacy chart as well.

**One correctness item outside the fifteen**, found while verifying finding 12. `legitimacyEndBreakdown` netted each jump against that turn's ambition completions, but a finished goal's `completed_turn` is its **start** turn — an explicit placeholder in `parsePlayerGoals` (`parsers/player-data.ts`), and this is the first consumer to treat it as a real turn (`LeaderCard` only tests it for null). So the award was being deducted on the turn the ambition was taken up, where it could suppress a genuine event row, and never on the turn it actually paid out. The subtraction is removed: event rows read the raw jump, and one sharing a turn with a real completion over-claims by the award, which the signed remainder absorbs — the same shape as the cognomen overlap already documented. **This changes numbers on the tab**: an event row that previously had an ambition netted out of it now reads higher. The footnote picked up a clause.

becked:
Pushed `f80efeb` — the Orders tab now itemizes in one panel per nation, whatever the player count.

The duel face-off table is gone. It and the FFA cards were two presentations of the same two breakdowns, and the table's union walk existed only to reconcile them. Each nation panel stacks the two itemizations, orders over legitimacy, both sitting in the recessed subpanel the Economy tab's ledgers already use. `SECTIONS` is now the single source for heading, precision, footnote and empty state, so they can't drift apart.

Two consequences worth calling out:

- **Gained:** `row.detail` renders on screen in every panel. In the duel table it was only a `title=` tooltip, so an ambition's names and a jump's turn were invisible without hovering.
- **Lost:** the side-by-side comparison — both players' values for the same source label in facing columns, union-ordered, with `—` where only one side had it.

One open question on `orders.ts`. `daf64d7` made the ambitions row's label constant and moved the count into the detail line so the duel table would merge both sides at differing counts. That table no longer exists, so the constraint is gone — and the detail line runs long regardless, because `GOAL_NAMES` entries are full sentences:

`4 × 10: Control 100 Population · Kill 10 Military Units with Leader as General · Control 20 Urban Improvements · +100 Food, Iron, Stone, and Wood / Turn`

Leaning toward one row per completed ambition. The `{#each}` key collision that forced the merge on event rows doesn't apply here, since goal types are distinct. Thoughts?

becked (commented):
Stepping back for a full re-review after the fix commits. I re-derived the tab's arithmetic against the 11 real saves in `test-data/saves/` rather than reading only the diff — that's where the substantive findings are.

## Correctness

### 1. The legitimacy itemization double-counts, and "Other" is frequently the largest row — with a sign that flips depending on the save

`src/lib/game-detail/orders.ts:142-253`

`LegitimacyHistory` records `getLegitimacy()`, which is a **stock**: it's spent (`RecruitLegitimacy`, `ConvertLegitimacy`, and the Schemer archetype's `bLegitimacyOrders` — "Use Legitimacy to buy Orders") and re-divided at succession. Summing only positive jumps ≥ 2 and adding them to the cognomen and ambition rows counts each cognomen/ambition award twice — once as its own row, once inside the jump on the turn it landed — while ignoring every decrease.

`OrdersTab.svelte:81` acknowledges the overlap and relies on the signed remainder to net it out. Measured across the corpus, it doesn't net to small:

| save | player | legit total | cognomen+ambition | event jumps | **Other** |
|---|---|---|---|---|---|
| `Sion vs Moroten` | P1 | 180 | 135 | 65 | **−20** |
| `OW-Rome-Year123` | P0 | 129 | 100 | 50 | **−21** |
| `OW-Aksum-Year117` | P1 | 204 | 72 | 0 | **+132** (65% of total) |
| `match_426504745` | P1 | 220 | 100 | 0 | **+120** (55%) |
| `OW-Maurya-Year111` | P1 | 144 | 46 | 0 | **+98** (68%) |

### 2. `story_events` is capped at the newest 100 for the entire game

`src/lib/parser/derive/story-events.ts:11,48`

The saves here carry 214–688 story events; the blob ships 100, shared across all players. Event attribution can therefore only ever see the tail window. **24 of 38 players across the 11 saves get zero event rows**, and in 4 saves (`match_426504745`, `OW-Maurya-Year111`, `match_426504772`, `OW-Babylonia-Year123`) not one player gets a single row. Whether a jump gets named depends on the player count and game length, not on what happened in the game.

This limit is pre-existing — `expeditionEvents` has the same ceiling — but the feature is built on it and `LEGITIMACY_NOTE` doesn't mention it.

### 3. An eliminated player renders a nonsense Orders panel

`src/lib/game-detail/OrdersTab.svelte:115-141`

`finalOrdersRate` is `0`, not null, for a player knocked out mid-game, so the breakdown renders anyway. In `OW-Aksum-Year117` player 3 died on turn 106: Total **0.0**, with Legitimacy 15.2 + Difficulty 6.0 + laws = 24.2 itemized, and Other **−24.2**.

## Fit

### 4. Difficulty renders the raw enum instead of the game's name

`orders.ts:100` produces "Difficulty (Great)" — the game calls that tier **Fragile**. `DIFFICULTY_NAMES` (`src/lib/generated/difficulty-names.ts`) exists for exactly this, and both sibling surfaces use the `DIFFICULTY_NAMES[x] ?? formatEnum(x, "DIFFICULTY_")` idiom: `SettingsTab.svelte:99,107` and `RecentSaveCard.svelte:278`.

### 5. `dynastyLeaders` duplicates LeadersTab's inline build

`orders.ts:54-61` is character-for-character `LeadersTab.svelte:76-83` — same filter, same sort. The helper was extracted, but the existing call site wasn't switched to it.

### 6. `characterLabel` duplicates `LeaderCard.rulerName`

`orders.ts:67-73` vs `LeaderCard.svelte:45-51` — same regnal-numeral rule, comment copied verbatim. `MilitaryTab.svelte:245-246` is a third variant of the same name+cognomen render. One shared helper.

### 7. Two dead guards

- `orders.ts:106` — `if (l.law == null) continue`, but `PlayerLaw.law` is a non-nullable `string`.
- `orders.ts:115` — `if (t.removed_turn != null) continue` never fires: the parser hardcodes `removedTurn: null` (`src/lib/parser/parsers/character-data.ts:100`). The "active at end" the doc comment claims isn't actually enforced.

### 8. The PR description is stale

It lists ruler traits as "Master Strategist +2, **Trusted +1**". `TRAIT_TRUSTED`'s grant is `<aiYieldRateGenerals>` (orders per *general*) — correctly excluded by ccfc7fb; only `TRAIT_MASTER_STRATEGIST` is in the baked table. The difficulty examples ("The Able", "The Great") are enum tokens rather than the game's labels, per finding 4.

## Verified clean

- **Cognomen recency math is right.** `Sion vs Moroten` P0 drops exactly 34 legitimacy at the turn-66 succession — the precise 2→3 leader re-division under `trunc(worth / (n − i))`.
- **Units line up.** `iPerLegitimacy` 1 ÷ `YIELDS_MULTIPLIER` 10 = 0.1, and blob yield rates are already ÷10 (`derive/yield-history.ts`).
- **The bake's granter sweep is complete.** Of the 13 ORDERS-granting `effectPlayer` entries in the current XML, the 9 reachable from law/trait/difficulty are baked; the other 4 are `advantage.xml` (MP compensation), which `ORDERS_NOTE` already names.
- `npm run lint` and `svelte-check` (0 errors, 1312 files) pass; the branch is rebased on current `main`.
- Chart-filter wiring, panel classes, `formatEnum` prefixes for law/trait/cognomen, and `ownedByPlayer`/`findByPlayer` usage all match their siblings.

Findings 1–3 are the ones that change what a user sees; 4–7 are mechanical.

becked:
Findings 3–7 are in as three commits (`64993e3..de8fa32`), fixed on the branch rather than handed back. `npm run lint` and `npm run check` are clean (0 errors, 0 warnings, 1312 files), and `prettier --check` is clean on every file touched. Findings 1–2 are untouched — they're one decision about what the legitimacy itemization should be, and it isn't a patch.

I re-derived each fix against the 11 saves in `test-data/saves/` rather than reasoning from the diff.

**`de8fa32` — finding 3, the eliminated realm.** The realm is gone when its reigning ruler is dead: succession appends a new id on every handover, so the tail of `<Leaders>` is dead only for a realm that fell without anyone taking the throne. `TechsTab` already reads exactly this signal for the science breakdown, with the same reasoning in its comment. Across the corpus the gate fires for 3 players of 38, and all three hold **no cities and no units** — it finds eliminated realms and nothing else.

Two of the three turned out worse than the finding described. `OW-Babylonia` P4 Carthage (fell T87 of 123) and `OW-Egypt` P2 Babylonia (fell T76 of 127) don't report a rate of 0 like the Aksum case — they report a stale forward-fill, 29.6 and 15.9 orders/turn, which itemized *plausibly* for a realm that had been gone 36 and 51 turns. Those two never looked wrong. The panel now names the turn the realm fell instead.

The Ending Legitimacy panel still renders for those players. That's finding 1's territory, so I left it.

**`0759af7` — findings 4 and 7.** `DIFFICULTY_NAMES[x] ?? formatEnum(x, "DIFFICULTY_")`, matching `SettingsTab` and `RecentSaveCard`; all five difficulty enums in the corpus resolve through the table, so "Great" reads as **Fragile** everywhere it appears. Both dead guards are gone, and the trait doc now says what the data supports rather than what the guard implied: the save records acquisition but not removal, so a trait the ruler later lost is still priced and the signed remainder absorbs it.

**`64993e3` — findings 5 and 6.** `dynastyLeaders`, `rulerName` and `rulerCognomen` live in `helpers.ts` now, next to the other per-player selectors. Both findings had a call site the review didn't name: `MilitaryTab:384` was a *third* copy of the dynasty filter-and-sort, feeding the leader rail.

That rail is also where this shows up as a fix rather than a wash — its tooltip had its own name render that dropped the regnal numeral. 12 of the 149 rulers across the corpus carry `suffix > 1`, so `OW-Maurya`'s Orrisa II and Orrisa III have been rendering identically. One other behaviour change worth flagging: `rulerName` returns null for an unnamed ruler instead of `formatEnum`'s "Unknown", which makes the Orders tab's `#${xml_id}` fallback reachable — `formatEnum("")` returns "Unknown", so it was dead code. No ruler in the corpus is unnamed, so nothing visible changes today.

---

**Finding 8 — the description.** Below is a replacement body. Beyond the two errors named in the review it corrects two more: the **Layout** section still describes the duel face-off table that `f80efeb` removed, and the event-attribution bullet claims jumps are netted against that turn's ambitions, which `daf64d7` deliberately made *not* the case (a finished goal's `completed_turn` is its start turn). I also dropped "on the demo duel it's small on the legitimacy side (+2 for one player)" rather than restate it — measured across the corpus, Other runs 55–68% of the legitimacy total in several saves, which is finding 1.

The screenshots need your hand: the two duel face-off images show a UI that no longer exists, so they're dropped below, and the remaining two predate `f80efeb` and want re-shooting against the per-nation panels.

```markdown
Adds an **Orders** tab to the game detail view — the action economy, charted and itemized. Sits after Techs; works for duels and FFAs; **no parser change** (every input already ships in today's blobs, so the whole existing corpus renders it).

## Why orders and legitimacy share a tab

They're mechanically one system: `Player.calculateNonCityYield` includes `getLegitimacy() × yield.xml ORDERS <iPerLegitimacy>` for **every** player — a tenth of an order per legitimacy point, per turn. In most games it's the single biggest orders line. So the tab pairs two charts (orders/turn and legitimacy, all players, nation colours, succession sawtooths visible) with end-of-game itemizations of both.

## The itemization, anchored on the game's calculation

**Orders/turn** (exact rows, from `bake-orders-sources.ts` walking source → `EffectPlayer` → `aiYieldRate` over `YIELDS_MULTIPLIER`):
- **Legitimacy** — final legitimacy × 0.1;
- **Difficulty** — the per-player MP handicap from `game_details.players`, named as the game names it (Affluent +10 … Fragile +5);
- **Laws** (Serfdom +5) active at end, and the reigning **ruler's traits** (Master Strategist +2). Only granters paying a flat player-wide orders rate are baked: `Trusted` pays `<aiYieldRateGenerals>` — orders per general, which the blob can't count — so it isn't priced.

**Legitimacy** (`Player.getLegitimacy` / `Character.getLegitimacy`):
- **Dynasty cognomens** — each ruler's cognomen worth divided by reign recency (integer division; current ruler full, predecessor half): "Amata the Fountainhead +60", "Ifra the Noble 70÷3";
- **Ambitions** — flat `FINISHED_AMBITION_BONUS` (+10) per completion, named from the goal data;
- **Story events, by timing** — a jump on a turn a story event fired for that player is credited to the event by name: "Mysterious Sword + Character Warmongers +30", "Calamities What Of The Gods +12". Succession turns are skipped (the dynasty term reshuffles), sub-2 leftovers stay unattributed, and event names are normalised across their `P.N.`/family/religion prefixed copies.

Everything the blob can't reconstruct — council/court ratings (a good chancellor is real orders), cathedrals/shrines, agents, trade — lands in a **signed Other** row against the save's true totals, with a footnote naming the suspects.

One correctness note from the game source, documented in the footnote: the save's yield-rate history records `calculateYield(eYield, false)` — **gross** production. Orders consumed by working/fortifying units feed only the in-game display, never the recorded series, so they are structurally absent from both the chart and the totals here.

**Eliminated realms don't get an orders itemization.** A dead reigning ruler means the realm fell without a successor; its recorded rate is 0 or a stale forward-fill, while the save still carries the legitimacy, difficulty and laws it held. Itemizing those prices a realm that no longer plays, so the panel says when it fell instead.

## Layout

One panel per nation, whatever the player count, each stacking the two itemizations — orders over legitimacy — in the recessed subpanel the Economy tab's ledgers use.

## Screenshots

(to re-shoot against the per-nation panels)

## Checks

`npm run lint`, `svelte-check`, prettier clean. New bake step `bake:orders-sources` registered in `bake:all`; the generated module is frontend-only (no worker mirror needed). Old blobs degrade gracefully (a 2.6.1 FFA lacks cognomen data — those rows simply stay in Other).
```

becked:
Status update — pushed `43b2e75`, rebased on current `main`.

## Finding 2 (`story_events` capped at 100) — addressed

The cap is gone; the blob ships the game's whole event history. Tracing where it came from: it was a verbatim port of the DuckDB query in `src-tauri/src/db/queries/match_data.rs` (`ORDER BY occurred_turn DESC, event_id DESC LIMIT 100`), added 2025-11-15 in `fe59b6c` to fill a "100 most recent events" table in the desktop Events tab. That table is gone; every consumer since reads these rows as per-(player, turn) evidence, so the cap stopped being a view limit and became a truncation of what gets persisted. It was the only `slice`/`LIMIT` in the derive layer.

Measured on the 12 saves in `test-data/saves/`, the window covered only the last 12–17 turns of games running 43–127 turns. Cost of lifting it, before → after:

| | uncompressed | gzipped |
|---|---|---|
| `OW-Aksum-Year117` (largest blob) | 8662 → 8733 KB | 404 → 410 KB |
| `match_426504745` (most events, 688) | 2700 → 2792 KB | 139 → 146 KB |

Worst case +92 KB uncompressed / +7 KB gzipped, against `MAX_BLOB_COMPRESSED` 10 MB and `MAX_BLOB_DECOMPRESSED` 50 MB. `PARSER_VERSION` 2.13.1 (value-changing per the §10 rules — old blobs render but under-report).

Duplicates are deliberately left alone. Roughly half the rows are audience copies of the same event, and it's tempting to collapse them in the parser — but the prefix (`P.1.`, `FAMILY_*`, `RELIGION_*`) is the only place the counterparty is recorded, it disagrees with the row's owning player 76 times out of 194 in `OW-Aksum-Year117`, and freezing a read-side dedupe key into the write layer is the same move that produced the `LIMIT 100` in the first place.

## A second defect this surfaced — also fixed

Validating a reparse showed every consumer joining story rows on `player_name`, and **single-player saves leave `player_name` empty for every player**. `OW-Aksum-Year117` has six players all named `""` and 582 events, so all six received all 582 — and lifting the cap widened that from 100 rows over 15 turns to the full history.

Rows now carry `player_xml_id`, and the three consumers (`expeditionEvents`, `scienceSpikes`, `legitimacyEndBreakdown`) join on it via a new `storyEventsFor` helper — the story-row counterpart to `ownedByPlayer`, falling back to the name for pre-2.14.0 blobs since a `StoryEvent` carries no nation. Verified: `OW-Aksum-Year117` now splits `{0:564, 1:3, 2:3, 3:3, 4:7, 5:2}`, and on MP saves the id join reproduces the name join exactly (`Sion vs Moroten`: 170/189 either way). `PARSER_VERSION` 2.14.0 — additive.

`scienceSpikes` also goes through `storyEventType` now rather than hand-rolling the same `EVENTSTORY_` slice, so there's one normalization for all three consumers.

**Deploy ordering matters here:** Worker first. A frontend emitting 2.14.0 against a Worker that doesn't know the version fails every upload with `INVALID_BLOB: Unknown parser_version`.

## Still open

- **Finding 1 — the legitimacy itemization double-counts and "Other" is often the largest row.** Still resting on the signed remainder (`orders.ts:77`, `:193`). Worth re-measuring before judging: three of the five rows in that table showed `event jumps = 0`, which was itself the cap dropping their events — with full history those jumps can now be named, so the remainder should shrink. I have not re-run those numbers.
- **Finding 8 — the PR description is stale.** Still lists `Trusted +1` (excluded by `ccfc7fb` — its grant is `<aiYieldRateGenerals>`) and names difficulty tiers by enum token ("The Able", "The Great") rather than the in-game names.

`npm run lint` clean, `svelte-check` 0 errors across 1312 files, `cd cloud && npm test` at 782 passed with the 4 pre-existing `canonical-map-options` failures unchanged (#115/#141, unrelated — this branch touches no map-options file).

becked:
Correction and detail on how `43b2e75` interacts with finding 1 — my previous comment said the unnamed remainder "should shrink," which is too strong and wrong for two of the five rows in that table.

Finding 1's `event jumps` column is what `legitimacyEndBreakdown` manages to name: a legitimacy jump ≥ 2 that has a same-turn story event. Three of those five rows show `0`, and the cap is why — those players' jumps happened at turns the newest-100 window never reached. So the cap was a direct cause of the column that let "Other" swallow 55–68% of the total.

Measuring how many story rows each of those players can actually see, before the change and after:

| finding-1 row | | visible before | visible after |
|---|---|---|---|
| `Sion vs Moroten` P1 | MP | 45 (T51–66, 15 turns) | 189 (T6–66, 44 turns) |
| `OW-Rome-Year123` P0 | SP, save owner | 100 (T103–123, 20 turns) | 486 (T1–123, 104 turns) |
| `match_426504745` P1 | MP | 56 (T106–122, 14 turns) | 329 (T6–122, 86 turns) |
| `OW-Aksum-Year117` P1 | SP, AI player | 100 (T106–117, 12 turns) | **3** (3 turns) |
| `OW-Maurya-Year111` P1 | SP, AI player | 100 (T98–111, 14 turns) | **6** (3 turns) |

The top three gain 4–8× the coverage, so their jumps can now be named and their remainder should genuinely fall.

The bottom two go the other way, and it isn't a regression — it's a false abundance being corrected. Their "before" 100 rows were *every* player's events, handed to them by the empty-`player_name` join that `player_xml_id` fixes. Even with all 100 they still scored 0 named jumps, because those rows sat in the last 12–14 turns. What they see now is what the save actually records on their own nodes: 3 and 6 events across a whole game. Their "Other" will not shrink, because there is nothing to attribute with.

That also suggests finding 1 is two failures wearing one number, and the sign separates them:

- **Negative "Other"** (`Sion vs Moroten` P1 −20, `OW-Rome-Year123` P0 −21) is the double-count signature — the itemization over-claims, because a named jump on a turn that also carried a cognomen or ambition award counts that award twice.
- **Large positive "Other"** (Aksum +132, `match_426504745` +120, Maurya +98) is the opposite failure: real legitimacy sources with nothing naming them.

This change adds named jumps. For the under-attributed rows that helps. For the over-attributed rows it gives the double-count more chances to fire, so those negatives plausibly get *more* negative. That last part is reasoning from the mechanism, not measurement — re-running `legitimacyEndBreakdown` over the corpus needs the `$lib` aliases my headless harness doesn't resolve, so I have not reproduced the table itself.

Net: `43b2e75` moves finding 1 rather than reducing it, and makes the two underlying failures separable. Whatever fixes the double-counting should be judged against the post-change numbers, not the ones in the original review.

becked:
## Legitimacy itemization: lifting the `story_events` cap makes the event rows over-claim

Lifting the newest-100 cap surfaces a pre-existing weakness in `legitimacyEndBreakdown` (`src/lib/game-detail/orders.ts`). Recording the options here rather than deciding them in the diff.

### What's happening

The panel explains a final legitimacy with three row kinds. Two are **prices** — recomputations of the game's own formula from constants baked out of the XML by `scripts/bake-orders-sources.ts`: cognomens as `COGNOMEN_LEGITIMACY[c] / (numLeaders − index)`, ambitions as `AMBITION_LEGITIMACY × count`.

The third kind is **attribution**, and it is not a price. The function walks the legitimacy series; when legitimacy rises by ≥2 on turn T and any story event fired for that player on turn T, the event is credited with *the entire jump*. The event's real legitimacy worth is never read — nothing in `story_events` carries one. `other` is then `final − Σrows`.

Under a newest-100 window only the last 12–22 turns had events, so the rule fired rarely. With the full history it fires on nearly every jump, through two mechanisms:

- **Double-counting.** Cognomens are awarded *by* events. `EVENTSTORY_LEGACY_OF_THE_RULER` shows up as its own `+16` row while the cognomen row already prices the same award.
- **False positives.** Most turns now carry some event, so jumps whose real cause the panel doesn't itemize (laws, wonders, religion spread, effect-granted legitimacy) get credited to whatever happened to fire — e.g. `+2 Marriage Offer Foreign (event on turn 9)`.

Measured by running the real function over `test-data/saves/`, old rows vs new:

| save | player | `other` before | `other` after | event rows |
|---|---|---|---|---|
| Sion vs Moroten | p0 Persia | +2 | **−105** | 6 → 27 |
| match_426504745 | p0 Carthage | −16 | **−189** | 2 → 19 |
| match_426504745 | p1 Assyria | +19 | **−162** | 2 → 20 |
| OW-Aksum-Year117 | p0 (uploader) | +29 | **−245** | 4 → 30 |
| OW-Maurya-Year111 | p0 (uploader) | +63 | **−190** | 5 → 31 |

Of the 12 multiplayer players whose breakdown attributes any event, `other` is negative for all 12 afterwards (6 of 12 before), and `|other|` goes from ≤24 for 11 of 12 to a 38–189 range — comparable to the final legitimacy itself (85–181). `OrdersTab.svelte` renders a negative `Other` in `text-red-400`, so it lands as a visible error state rather than a quiet inaccuracy.

The module header already states the intended design: *"Events, bonuses, and the rare effect-granted legitimacy land in the remainder."* The attribution block contradicts the contract the file opens with.

### Options, most accurate first

**B — price events from the XML.** The data supports this. `eventStory.xml` carries `aeBonuses`, documented as *"processed prior to Player being given the Options, and occur regardless of Option chosen"*, plus `aeOptions`; both story bonuses and option bonuses resolve into `bonus-event.xml`, whose entries carry `<iLegitimacy>N</iLegitimacy>`. So a story's **unconditional** legitimacy is exactly derivable — a true price row in the same class as cognomens. Option-dependent legitimacy is not, because the save records which story fired, not which option was chosen; but the option set is known, giving a min/max and — more valuably — a definitive allow-list of stories that can move legitimacy at all. That allow-list alone kills the false positives. Mechanically it extends `bake-orders-sources.ts`, which already reads this XML tree for the other two constants: a baker, not a new heuristic.

**C — net out what's already priced.** Skip turns whose jump the panel has itemized elsewhere. There is precedent inside the same function: `successionTurns` is skipped for exactly this reason. Extending the skip-set to cognomen-award turns is the same move. Ambitions can't be netted — `completed_turn` is a documented START-turn placeholder, and the existing comment explains why subtracting there would suppress genuine rows. Fixes double-counting only. Cheap, partial, folds into B or E.

**E — stop giving event rows a value.** Render them as annotations beneath the remainder (`T60 · Player Declare War Human`) with no number. `other` becomes the honest large remainder the module header says it should be, and the panel stops asserting a figure the save doesn't contain. Needs a look at the duel face-off table, which merges rows by label and sums values.

**D — raise the bar.** A larger jump threshold, or crediting only jumps no other row explains. Arbitrary — quieter, not more correct.

**A / F — clamp `other` at zero, or re-window events to recent turns.** Both hide the contradiction. F is the cap-as-feature: it makes the panel look right by starving it of data, which is what lifting the cap set out to stop.

### Recommendation

**E now, B when the tab needs to be exact.** E is a small change that stops the panel claiming a number the save doesn't contain; B is the only path to real accuracy and it reuses machinery that already exists. C is worth folding into whichever is picked.

becked:
## Review follow-ups pushed (`6a820fa`, `f556638`, `2e94dbf`)

Three commits on top of `43b2e75`. `npm run lint` clean, `npm run check` 1312 files / 0 errors, prettier clean.

### `fix(game-detail): a nameless player matches no story rows` — `6a820fa`

`storyEventsFor` falls back to `player_name` on blobs below `PARSER_VERSION` 2.14.0, and single-player saves leave that name empty for every player — so `"" === ""` handed each of them every realm's events. That is the exact bug `player_xml_id` was added to fix, still live on any blob not yet reparsed. An empty name now matches nothing, which is the choice the Techs tab already makes for memory rows carrying no usable key: *"Memory rows carry only player_xml_id (no nation), so there is no nation fallback — legacy blobs simply yield no espionage markers."*

Verified against `test-data/saves/` with the id stripped to simulate a legacy blob:

- 7/7 multiplayer saves still reproduce the id join exactly (Sion vs Moroten 170/189, match_426504745 329/359, …) — nothing lost where the name join already worked.
- 5/5 single-player saves go from handing all 4–6 players all 426–606 events to handing them none.

### `docs(parser): the story_events header states what the corpus measures` — `f556638`

Two figures in the derivation header were off. The newest-100 window spanned **12–22** turns, not ~12–17 — OW-Babylonia covered T102–123 and OW-Rome T103–123, both past the stated ceiling. And **0.8–3.4%** measured the cap lift alone, while the same change also puts `player_xml_id` on every row. Measured parent to HEAD across all twelve saves it is **0.9–3.9%**, worst case **+104 KiB** uncompressed and **+8 KiB** gzipped. The header names those against the Worker's 10 MB / 50 MB limits now, so the headroom is legible without re-deriving it.

### `fix(parser): story_events read oldest turn first` — `2e94dbf`

The sort was `occurred_turn DESC, event_id DESC` because it existed to pick the newest 100. Nothing selects a prefix any more, and descending was the odd one out — `stealTurns` sorts ascending, key-tech markers sort ascending, a legitimacy row lists its turns ascending.

The turn direction itself turns out to be invisible: all three consumers bucket by turn, and flipping it alone changed nothing for any of the 40 players in the corpus. The **tiebreaker** is what carries, in the two places that truncate — a spike tooltip keeps the first `SPIKE_SOURCES_MAX` same-turn events, a legitimacy row names the first two — where flipping it moved 14 of those 40. Ascending `event_id` is parse order, so a player's own events now outrank their characters' and cities' for those slots.

No `PARSER_VERSION` bump: 2.14.0 is unreleased, so the order change ships inside it.

---

## Two things left open, neither addressed here

**The per-audience duplicate rows.** The justification for keeping them holds — I checked, and 47–85 of the `P.N.`-prefixed rows per single-player save are owned by a *different* player than `N`, so the prefix genuinely records a counterparty that deduping would destroy. The counter-argument is that nothing reads it: all three consumers strip the prefix through `storyEventType` and then immediately dedupe, so a write-time dedupe on (player, turn, normalized type, character, city) would produce byte-identical consumer output while cutting the array ~47% — 688 → 365 on match_426504745, 359 → 189 on Sion vs Moroten. That would roughly halve the blob cost the header now quantifies. Live trade either way, worth deciding with the numbers attached rather than by default.

**Nothing pins `storyEventsFor`.** It has three branches now — id match, name match, and the empty-name skip added in `6a820fa` — and the frontend has no test harness to hold them (`vitest.config.mts` exists only in `cloud/`). `legitimacyEndBreakdown` is in the same position and is the more consequential of the two. Not #166, which is the separate `cloud/tsconfig.json` typecheck gap; there's no open issue covering this one. Standing up a frontend vitest project is its own change, not this PR's.

becked:
## `2.13.1` collapsed into `2.14.0` (`8630779`)

`KNOWN_PARSER_VERSIONS` is the record of blob shapes the Worker can receive, and every other patch entry in it earned its place — `git log -S` over `src/lib/parser/types.ts` finds 2.3.1, 2.4.1, 2.5.1, 2.6.1 and 2.9.1 each holding in `PARSER_VERSION` for a stretch of history (2 commits apiece: the one that set it, the one that moved past it).

`2.13.1` appears in **zero**. The constant went 2.13.0 straight to 2.14.0 in `43b2e75`, so no build ever emitted it and no blob can carry it. The changelog paragraph above the set was describing a release that didn't happen, and the three code comments reading *"blobs below parser 2.13.1"* pointed at a boundary nothing sits on — "below 2.13.1" and "below 2.14.0" partition the real world identically.

Changed:

- `cloud/src/schemas/game.ts` — both `story_events` changes (whole history + `player_xml_id`) now sit under one `2.14.0` entry, and `"2.13.1"` is out of the set.
- `science-techs.ts:29`, `science-techs.ts:882`, `TechsTab.svelte:326` — now name 2.14.0.

The merged paragraph also picks up the window figure corrected in `f556638`: **12–22 turns, not ~12–17** (OW-Babylonia covered T102–123, OW-Rome T103–123, both past the old ceiling). That number lived in two places and the earlier commit only fixed the parser copy — this is the sibling.

Nothing else referenced `2.13.1`: `grep -rn "2\.13\.1" src cloud docs` is empty, and no test or fixture pinned it. Verified `npm run lint`, `npm run check` (1312 files / 0 errors), `cd cloud && npm run typecheck` clean, and `cd cloud && npm test` at 782 passed / 4 failed — the same four pre-existing `canonical-map-options.test.ts` DOTA-drift failures as before this branch, unchanged.

One consequence worth naming: a blob hand-stamped `2.13.1` (the local test artifact from developing this branch) will now be rejected on upload with `INVALID_BLOB: Unknown parser_version`. Reparsing it produces 2.14.0 with identical content.

---

Separately, the `event_logs` sibling defect raised in review is now filed as #213 — same name-join-with-no-id shape, live on `main`, not caused by or in scope for this PR.

PR #195, opened 2026-08-05
fix(tournament): match card leads with the next session; played sessions collapse

alcaras:
Fixes the TO bug report: *"if a match has more than 6 parts they don't show on the card, just when clicking schedule — you can see this on the live site with the fiddlers/nobody match, they have a scheduled part 7 but it doesn't show on the card."*

## The actual bug — the card's status chip

`matchDisplayStatus` checked `hasStartedPart` before `isMatchScheduled`, so the moment any session's time passed, the match read **"In progress"** forever. The bracket card renders only that chip — so a split match's scheduled next session was invisible on the card, discoverable only inside the schedule editor. Exactly the report.

Classification now runs **live → scheduled → started → unscheduled**: a session started within `LIVE_WINDOW_MS` still reads "In progress" (a match being streamed outranks its next booking), an upcoming session reads "Scheduled", and a match with everything aged out and nothing booked reads "In progress" as before. One shared classifier, so the bracket cards, championship tree, and the matches table's filter buckets all correct together — and the table's sort comment already expected this case ("a split match mid-schedule sorts by its next sitting").

The reported match, before and after:

before card
after card

## Also fixed along the way: the card scroll fold

No data cap anywhere: all seven parts arrive and all seven render. But the match card scrolls (`max-h-[85vh]`), the schedule panel listed sittings in play order, and six played timestamps plus casting rows push the **only sitting that matters — the next one — below the fold**, where macOS's auto-hiding scrollbars make it read as missing. Measured on the live match: Part 7's row sat at y≈1092 in a 900px viewport.

## The fix — lead with what matters

The schedule panel now renders **upcoming sessions first**; sittings whose time has passed collapse behind a **"N played sessions"** toggle when there are two or more. With nothing still ahead, the most recent played session stays inline so the panel never opens empty. Single-session matches and 1-played/1-ahead splits render exactly as before — only the wall-of-stale-dates case changes. The partition reads the shared clock, so a session slides into the played group as its time passes.

Before (live site, the reported match — Part 7 is far below this fold):

before

After (same seven parts on a local fixture — the next sitting is the first row):

after

Expanded:

expanded

Deliberately untouched: the **Casting** panel below still lists every cast session — that content is archival (who cast what, stream links) rather than a schedule, and collapsing it is a separate design call.

`lint` / `svelte-check` / prettier clean.

### DIFF
--- a/src/lib/tournament/MatchPopover.svelte
+++ b/src/lib/tournament/MatchPopover.svelte
@@ -44,6 +44,7 @@
 		matchSlotUserId,
 	} from "$lib/tournament/match-occupant";
 	import { upcomingScheduledParts } from "$lib/tournament/parts";
+	import { nowMs } from "$lib/stores/now.svelte";
 	import { seshMatchLine, seshVersus } from "$lib/tournament/sesh";
 	import { mapScriptLabel } from "$lib/tournament/map-scripts";
 	import {
@@ -379,6 +380,41 @@
 	// castingParts keeps each part's original 1-based number so a split match
 	// labels "Part N" consistently in both panels, and drops parts with no
 	// broadcast info so the casting panel only lists sittings that have some.
+	// Schedule rows, upcoming-first. A long split match accumulates played
+	// sittings that bury the one a viewer actually needs — the NEXT one —
+	// below the popover's fold (the report that prompted this was a match on
+	// sitting seven, whose only future time sat off-screen). Sittings that
+	// have started render behind a collapsed "N played sittings" toggle when
+	// there are two or more of them; with nothing still ahead, the most
+	// recent played sitting stays inline so the panel never opens empty.
+	// Reads the shared clock, so a sitting moves into the played group as
+	// its time passes.
+	const numberedSittings = $derived(
+		match.parts.map((part, i) => ({ part, partNumber: i + 1 })),
+	);
+	const startedSittings = $derived(
+		numberedSittings.filter(({ part }) => {
+			if (part.scheduled_at == null) return false;
+			const t = Date.parse(part.scheduled_at);
+			return !Number.isNaN(t) && t <= nowMs();
+		}),
+	);
+	const aheadSittings = $derived(
+		numberedSittings.filter((np) => !startedSittings.includes(np)),
+	);
+	const collapsedSittings = $derived(
+		startedSittings.length >= 2
+			? aheadSittings.length > 0
+				? startedSittings
+				: startedSittings.slice(0, -1)
+			: [],
+	);
+	const scheduleRows = $derived([
+		...startedSittings.filter((np) => !collapsedSittings.includes(np)),
+		...aheadSittings,
+	]);
+	let showPlayed = $state(false);
+
 	const castingParts = $derived(
 		match.parts
 			.map((part, i) => ({ part, partNumber: i + 1 }))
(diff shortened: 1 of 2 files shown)


alcaras:
## Self-review

One real behavioral finding, fixed in f96d590, plus a placement slip; the rest verified against the surrounding schedule machinery.

**Fixed**

- **"Played" was drawn at the wrong boundary — a live session collapsed.** The first cut classified a session as played the moment its start time passed, so a session that began half an hour ago — the one being *streamed right now* — vanished into the collapsed group. The app already draws this exact line: `liveAndUpcoming` keeps a sitting "plausibly live" for `LIVE_WINDOW_MS` (4h) after start, and #193's `owesNextSitting` uses the same cutoff from the other side. The collapse now uses it too: a session inside the live window stays inline, in part order, ahead of the future ones — so all three surfaces (Live list, sesh export, this card) flip a session's state at the same instant. `startedSessions` renamed `playedSessions` to match what it now means.
- **Comment interleave** — the new block had been inserted between `castingParts`' doc comment and its declaration (the same slip #193's review caught in `parts.ts`; apparently I need a lint rule for myself). Reunited.

**Verified clean**

- **No behavior change outside the wall-of-stale-dates case**: single-session matches (no toggle, no labels), 1-played/1-ahead splits (both inline, as before), all-unscheduled matches, and completed matches (nothing ahead → most recent played session inline, earlier ones collapsed — the panel never opens empty).
- **Reactivity**: partitions read the shared clock, so a session slides from ahead → live-inline → collapsed as real time passes, without a refetch; `showPlayed` resets naturally when the popover keys to another match.
- **Keyed rendering**: rows key on `part.id` (server-minted, unique — verified on the reported match's seven parts), so expanding/collapsing reorders without duplication; identity-based `includes` partitioning is safe within one derived graph evaluation and trivially small (parts cap is 30).
- **The report's match, reproduced**: the real seven parts pulled from prod onto a local fixture — before: Part 7 at y≈1092 in a 900px viewport (confirmed on the live site itself); after: "▸ 6 played sessions" with Part 7 as the first visible row, six hours out.
- Fit: `LIVE_WINDOW_MS` imported from `parts.ts` rather than a second constant; vocabulary is "sessions" (migration 0029's own word) after the first cut said "sittings"; `aria-expanded` on the toggle.

**Deliberately unchanged**: the Casting panel still lists every cast session — archival broadcast info rather than a schedule; collapsing it is a separate design call.

alcaras:
Addendum to the self-review: alcaras caught that the scroll fix only treated the symptom — the deeper half of the original report was the **Casting panel**, which filtered to sessions with broadcast info and therefore never listed a split match's *next* session (uncast by definition until someone signs up). Fixed in cc88b80: a pending match's not-yet-played sessions now always appear there with an italic "Needs a caster" placeholder — the panel advertises the open seat instead of hiding it. Played sessions still appear only when they have something to archive. (castingParts also now derives from the shared numberedSessions instances so the played-set membership check is identity-safe.)

alcaras:
Second addendum — alcaras rejected both earlier diagnoses, correctly. The actual bug is in `matchDisplayStatus`: `hasStartedPart` outranked `isMatchScheduled`, so once any session passed, the bracket card's chip — the only thing it renders — said **In progress** forever and the scheduled next session never surfaced. Fixed in c63b6af with a live-window-aware ordering (live → scheduled → started → unscheduled) in the one shared classifier, so cards, championship tree, and the matches-table buckets correct together. The two earlier fixes (fold collapse, casting-panel listing) remain — real adjacent problems, just not *the* problem. PR body rewritten to lead with the real one.

alcaras:
Closing — this PR chased the wrong surface(s) for the reported bug. The branch stays (the three changes here may come back individually if wanted), and a fresh PR will follow once the actual issue is pinned down with a screenshot.

PR #197, opened 2026-08-06
fix(tournament): collapse played sessions so the next one stays above the popover fold

alcaras:
Fixes the "a match with more than 6 sessions doesn't show its next time on the card" report (previously mis-chased in #195 — this is the minimal redo).

## The bug

The match popover's schedule panel lists every session in order, and the shared popover frame clamps at `max-h-[85vh]` with `overflow-y-auto`. On a short laptop screen that fold lands right around the sixth row — and since macOS overlay scrollbars are invisible until you scroll, there is no hint that anything sits below it. A split match on session seven looked like it had no upcoming time at all; the only future session was exactly the row below the fold. The bug is viewport-dependent, which is why it reproduced on one machine and not another.

Before (1280×700 viewport, 7-session match — Part 7 clipped at the very bottom edge; a slightly shorter window hides it entirely):

before

## The fix

Played sessions collapse behind a "▸ N played sessions" toggle, so the panel stays short at any viewport height and the schedule leads with what a viewer actually needs — the next (or live) session:

after

- A session counts as played once it has aged out of `LIVE_WINDOW_MS` — the same boundary `liveAndUpcoming` draws — so a session being streamed right now stays inline, in order, with the upcoming ones.
- The toggle appears only at two or more played sessions; when there is nothing else to show, the most recent played session stays inline so the panel never opens empty.
- Rows keep their original "Part N" numbers (shared `numberedSessions` derived, also reused by the casting panel so both panels number from the same instances).
- Reads the shared `nowMs` clock, so a session slides into the played group as it ages out, live.

One file, UI-only, no data or API changes. `npm run lint`, `svelte-check` (0 errors), prettier all clean.

### DIFF
--- a/src/lib/tournament/MatchPopover.svelte
+++ b/src/lib/tournament/MatchPopover.svelte
@@ -382,6 +382,9 @@
 			match.status !== "bye" &&
 			(isAdmin || (isParticipant && match.status === "pending")),
 	);
+	// Whether the collapsed played sessions are currently disclosed. Reset per
+	// match by the parent's {#key match.match_id} remount.
+	let showPlayed = $state(false);
 	// Schedule rows, played-last-and-collapsed. A long split match
 	// accumulates played sessions that push the one a viewer actually needs
 	// — the NEXT one — below the popover frame's max-h-[85vh] fold, where on
@@ -400,24 +403,23 @@
 	const numberedSessions = $derived(
 		match.parts.map((part, i) => ({ part, partNumber: i + 1 })),
 	);
-	const playedSessions = $derived(
-		numberedSessions.filter(({ part }) => partPlayed(part)),
-	);
-	const aheadSessions = $derived(
-		numberedSessions.filter((np) => !playedSessions.includes(np)),
-	);
-	const collapsedSessions = $derived(
-		playedSessions.length >= 2
-			? aheadSessions.length > 0
-				? playedSessions
-				: playedSessions.slice(0, -1)
-			: [],
-	);
-	const scheduleRows = $derived([
-		...playedSessions.filter((np) => !collapsedSessions.includes(np)),
-		...aheadSessions,
-	]);
-	let showPlayed = $state(false);
+	const schedule = $derived.by(() => {
+		const played = numberedSessions.filter(({ part }) => partPlayed(part));
+		const ahead = numberedSessions.filter(({ part }) => !partPlayed(part));
+		// A single count drives the whole partition: the hidden rows are the
+		// first `hidden` played ones, the rest render inline. Slicing by that
+		// count keeps played order intact without a membership test.
+		const hidden =
+			played.length < 2
+				? 0
+				: ahead.length > 0
+					? played.length
+					: played.length - 1;
+		return {
+			collapsedCount: hidden,
+			rows: [...(showPlayed ? played : played.slice(hidden)), ...ahead],
+		};
+	});
 
 	// Read view splits the old combined parts block into two stacked panels: the
 	// schedule (per-part times) and casting (per-part casters + stream links).
@@ -1113,7 +1115,7 @@
 			class="flex flex-col gap-2 rounded-lg p-3"
 			style="background-color: rgb(var(--color-surface-raised));"
 		>
-			{#if collapsedSessions.length > 0}
+			{#if schedule.collapsedCount > 0}
 				<button
 					type="button"
 					class="flex items-center gap-1 self-start text-[10px] font-bold uppercase tracking-wider text-muted transition-colors hover:text-tan"
@@ -1121,13 +1123,13 @@
 					aria-expanded={showPlayed}
 				>
 					<span aria-hidden="true">{showPlayed ? "▾" : "▸"}</span>
-					{collapsedSessions.length} played
-					{collapsedSessions.length === 1 ? "session" : "sessions"}
+					{schedule.collapsedCount} played
+					{schedule.collapsedCount === 1 ? "session" : "sessions"}
 				</button>
 			{/if}
-			{#each showPlayed ? [...collapsedSessions, ...scheduleRows] : scheduleRows as { part, partNumber }, i (part.id)}
+			{#each schedule.rows as { part, partNumber }, i (part.id)}
 				<div
-					class="flex items-center gap-2 {i > 0 || collapsedSessions.length > 0
+					class="flex items-center gap-2 {i > 0 || schedule.collapsedCount > 0
 						? 'border-t border-border-subtle pt-2'
 						: ''}"
 				>
--- a/src/lib/tournament/MatchPopover.svelte
+++ b/src/lib/tournament/MatchPopover.svelte
@@ -382,30 +382,29 @@
 			match.status !== "bye" &&
 			(isAdmin || (isParticipant && match.status === "pending")),
 	);
-	// Whether the collapsed played sessions are currently disclosed. Reset per
+	// Whether the collapsed played parts are currently disclosed. Reset per
 	// match by the parent's {#key match.match_id} remount.
 	let showPlayed = $state(false);
-	// Schedule rows, played-last-and-collapsed. A long split match
-	// accumulates played sessions that push the one a viewer actually needs
-	// — the NEXT one — below the popover frame's max-h-[85vh] fold, where on
-	// a short laptop screen it clips around the sixth row and the overlay
-	// scrollbar gives no hint there's more (the report that prompted this was
-	// a match on session seven whose only future time sat off-screen).
-	// Collapsing keeps the panel short at any viewport height. A session
-	// counts as played once partPlayed says so — aged out of LIVE_WINDOW_MS,
-	// the same boundary the live/upcoming surfaces draw — so a session being
-	// streamed right now stays inline, in order, with the upcoming ones.
-	// Played sessions render behind a collapsed "N played sessions" toggle
-	// when there are two or more; with nothing else to show, the most recent
-	// played session stays inline so the panel never opens empty. partPlayed
-	// reads the shared clock, so a session slides into the played group as it
-	// ages out.
-	const numberedSessions = $derived(
+	// Schedule rows, played-last-and-collapsed. A long split match accumulates
+	// played parts that push the one a viewer actually needs — the NEXT one —
+	// below the popover frame's max-h-[85vh] fold, where on a short laptop
+	// screen it clips around the sixth row and the overlay scrollbar gives no
+	// hint there's more (the report that prompted this was a match on part
+	// seven whose only future time sat off-screen). Collapsing keeps the panel
+	// short at any viewport height. A part counts as played once partPlayed
+	// says so — aged out of LIVE_WINDOW_MS, the same boundary the live/upcoming
+	// surfaces draw — so a part being streamed right now stays inline, in
+	// order, with the upcoming ones. Played parts render behind a collapsed
+	// "N played parts" toggle when there are two or more; with nothing else to
+	// show, the most recent played part stays inline so the panel never opens
+	// empty. partPlayed reads the shared clock, so a part slides into the
+	// played group as it ages out.
+	const numberedParts = $derived(
 		match.parts.map((part, i) => ({ part, partNumber: i + 1 })),
 	);
 	const schedule = $derived.by(() => {
-		const played = numberedSessions.filter(({ part }) => partPlayed(part));
-		const ahead = numberedSessions.filter(({ part }) => !partPlayed(part));
+		const played = numberedParts.filter(({ part }) => partPlayed(part));
+		const ahead = numberedParts.filter(({ part }) => !partPlayed(part));
 		// A single count drives the whole partition: the hidden rows are the
 		// first `hidden` played ones, the rest render inline. Slicing by that
 		// count keeps played order intact without a membership test.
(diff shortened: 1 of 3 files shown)


alcaras:
Self-review (correctness + fit per the pr-review checklist):

**Applied before review request:**
- **Fit — time classifiers live in `parts.ts`.** The first cut inlined the aged-out predicate (`Date.parse` + `nowMs() - LIVE_WINDOW_MS`) in the component; every other part-time classifier (`hasStartedPart`, `isMatchScheduled`, `upcomingScheduledParts`) lives in `parts.ts` beside `LIVE_WINDOW_MS` and reads the shared clock there. Extracted as `partPlayed(part)` with one consumer here — one definition of "played" for any future surface that groups past sessions (the sesh export in #193 wants the same boundary).

**Checked, no finding:**
- Both `MatchPopover` call sites (tournament page, matches page) get the change — it's internal to the component; `SchedulePopover` (the editor) deliberately keeps the full uncollapsed list since editing needs every row.
- Keyed each stays duplicate-free: expanded view is `[...collapsedSessions, ...scheduleRows]` and `scheduleRows` excludes collapsed by construction.
- Edge cases: 1 played session → no toggle, renders inline as before; all sessions played, none ahead (finished-but-unreported or completed match) → most recent played stays inline so the panel never opens empty; toggle pluralizes ("1 played session" is reachable at exactly 2 played, 0 ahead).
- `showPlayed` resets on match navigation via the parent's existing `{#key match.match_id}` remount.
- Reactivity: `partPlayed` reads `nowMs()` inside the `$derived`, so a live session slides into the collapsed group as it ages out, without a refetch.
- `castingParts` refactors onto the shared `numberedSessions` instances with identical filter behavior — no casting-panel change.

Screenshots in the body are the real 7-session fixture at 1280×700; re-verified after the extraction (identical render).

becked:
Read through this against `main`. Notes below, split into what I verified and what I think needs a decision.

## Verified

- **The fold is real.** `src/lib/ui/Popover.svelte:75` is `max-h-[85vh] overflow-y-auto`, and there's no cap on parts anywhere in the data path — all sessions arrive and all render. The diagnosis holds.
- **The boundary claim is accurate.** `liveAndUpcoming` classifies live as `t > now - LIVE_WINDOW_MS` (`schedule.ts:51`); `partPlayed` is `t <= now - LIVE_WINDOW_MS`, the exact complement. A session being streamed right now does stay inline.
- **`showPlayed` can't leak between matches.** All three `MatchPopover` call sites (`TournamentUpNextPanel`, `/tournaments/[slug]`, `/tournaments/[slug]/matches`) wrap it in `{#key match.match_id}`, so the toggle resets per match.
- **No tests isn't a deviation.** There are zero `*.test.ts` files under `src/` — untested frontend helpers are the status quo.
- **`castingParts` reusing `numberedSessions`** instead of re-mapping `match.parts` is a genuine dedup, and it keeps both panels numbering from the same instances as claimed.

Minor: the body says "one file" — it's two (`MatchPopover.svelte` + `parts.ts`).

## Needs a decision

**1. Which bug is this fixing?** #195 argued the real defect was `matchDisplayStatus` checking `hasStartedPart` before `isMatchScheduled`, so a split match reads "In progress" forever once any session's time passes. That is still true on `main`, and the bracket cards (`SwissFlowBracket`, `ChampionshipBracketTree`) render *only* that chip — no time at all. This PR calls #195 mis-chased and leaves the classifier alone. If the report's "card" meant the bracket card, this doesn't address it; if it meant the popover, it does. Quite possibly both are real and want separate PRs — but we should say which one this is.

**2. It changes every completed split match, not just the reported case.** Once a match is done every session is aged out, so `aheadSessions` is empty, `collapsedSessions` is `played.slice(0, -1)`, and only the final session renders inline. A finished five-session match's popover now opens showing one row. That's a change to an archival view the report didn't ask about — intended, or a side effect worth gating on `match.status === "pending"`?

**3. The aged-out boundary is now encoded twice.** `schedule.ts:51` has it inline in `liveAndUpcoming`; `parts.ts:166` now has it as a predicate. Same threshold, two definitions — exactly the drift the "extract, don't copy-paste" rule is about. `liveAndUpcoming` should call `partPlayed`. The obstacle is signature drift: `liveAndUpcoming` takes `now` as a parameter while `partPlayed` reads `nowMs()` internally. An optional `now` argument on `partPlayed` (defaulting to `nowMs()`) would let both share one definition.

**4. The derived chain is heavier than the logic.** Five deriveds, and `aheadSessions` is defined as "not present in `playedSessions`" via `.includes()` — O(n²) and indirect where `!partPlayed(part)` would say it directly. The expanded list is also rebuilt inline in the `{#each}` expression. One `$derived.by` returning `{ collapsed, inline }` would carry the whole partition and read as the single decision it is. Also, `let showPlayed = $state(false)` currently sits in the middle of the derived block.

**5. The toggle doesn't reuse the local disclosure pattern.** `TournamentMapsPanel.svelte:133` already does expand/collapse in this directory: SVG chevron with `class:rotate-90={isExpanded}` plus a descriptive `aria-label` ("Expand …"/"Collapse …"). This one uses a text `▸`/`▾` glyph and has `aria-expanded` but no `aria-label` or `aria-controls`. Worth matching the existing one.

**6. One edge case reorders rows.** Parts are an authored, ordered JSON list — `SchedulePopover` appends new ones and nothing sorts by `scheduled_at`. Partitioning into played-then-ahead means a part whose time is edited out of sequence changes position, where the panel previously always rendered authored order. Small, but it's a behavior the old `{#each match.parts}` guaranteed for free.

**7. Rebased.** The branch was 30 commits behind `main`, so I've rebased it onto current `main` and force-pushed. It applied cleanly — no conflict, even though `main` rewrote `canSchedule`, the derived immediately above where this PR inserts its block. `npm run lint`, `svelte-check` (0 errors, 0 warnings) and prettier are all still clean on the rebased tree.

One thing that rebase surfaces, which sharpens point 2: on current `main`, `canSchedule` now drops the Schedule button for participants on decided matches. Combined with this PR, a participant opening a finished split match gets a one-row schedule panel and no Schedule button — the expand toggle becomes their only route to the other sessions.

becked:
> **Update:** I've taken care of #4 myself and pushed it to this branch as `4f7f552` — details at the bottom. **#1, #2 and #3 are still open for you.**

Second pass, focused on cleanliness/fit. Four things to fix — none are correctness, all are "this diverges from what the repo already does."

## 1. "Session" isn't this codebase's word for a part

`docs/tournament-rules.md:236` is explicit: scheduling metadata is "a per-match list of **parts** (sittings)". The code follows it everywhere — `matchParts`, `NumberedPart`, `partNumber`, `upcomingScheduledParts`, `rowIsPendingSitting`, and ~96 occurrences of "sitting" across the tournament code and docs. "Session" appears nowhere as a noun for this concept; the only hits are the adjective in "single-session match".

This PR introduces it in three places:

- Identifiers: `numberedSessions`, `playedSessions`, `aheadSessions`, `collapsedSessions` (`MatchPopover.svelte:400-419`).
- A **user-visible** string, `"{n} played session(s)"` (`:1124-1125`), rendered directly above rows labelled "Part 1" / "Part 2" — so the panel now uses two different nouns for the same thing, a foot apart.
- An existing comment rewritten from "sittings" to "sessions" (`:426`), and new ones in `parts.ts:157-162`.

Please rename to parts throughout — `numberedParts` / `playedParts` / `aheadParts` / `collapsedParts`, toggle text "{n} played part(s)", and revert the comment at `:426`. This is the CLAUDE.md rule-2 case ("a term that appears nowhere else in the repo is a red flag").

## 2. The aged-out boundary is now defined twice

`schedule.ts:51` carries it inline (`else if (t > now - LIVE_WINDOW_MS)`), and `parts.ts:166` now carries its exact complement as a predicate. Same threshold, two definitions — the "extract, don't copy-paste" case.

Give `partPlayed` an optional clock so both can share one definition:

```ts
export function partPlayed(part: TournamentMatchPart, now = nowMs()): boolean
```

then `liveAndUpcoming`'s live branch becomes `else if (!partPlayed(np.part, now))` and the threshold exists once.

While you're in `parts.ts`: `partPlayed` is the fourth copy of the `scheduled_at == null` → `Date.parse` → `Number.isNaN` guard, alongside `hasStartedPart:58`, `nextScheduledAt:76` and `upcomingScheduledParts:186`. A `partInstant(part): number | null` would collapse all four. Pre-existing, so treat as optional — but this PR is what makes it four.

## 3. The disclosure toggle doesn't reuse the one in the same directory

`TournamentMapsPanel.svelte:128-164` is the established expand/collapse pattern here. The new toggle diverges from it three ways:

- **Chevron.** That one uses an inline SVG with `class="… transition-transform" class:rotate-90={isExpanded}`; this uses text glyphs `▸` / `▾` (`:1123`).
- **Label.** That one has `aria-label={`${isExpanded ? "Collapse" : "Expand"} …`}`; this has `aria-expanded` but no accessible name beyond the visible count.
- **Pluralisation.** That one is `{optsCount} option{optsCount === 1 ? "" : "s"}`; this re-implements it as `{n === 1 ? "session" : "sessions"}` (`:1125`).

Please match the existing component on all three.

## 4. Fold the derived chain into one partition — done in `4f7f552`

*(Original note, for context: `MatchPopover.svelte:400-420` was five deriveds expressing a single decision — which rows are hidden and which render — plus `let showPlayed = $state(false)` at `:420` wedged into the middle of the derived block, and the expanded list rebuilt inline inside the `{#each}` at `:1128`.)*

I went ahead and did this one rather than leave it on your plate, since it touched the same lines as a smaller identity bug I wanted out of the way. Pushed as `4f7f552`:

- **One `$derived.by`** (`schedule`) replaces the five, returning `{ collapsedCount, rows }` — the count for the toggle, the finished list for the `{#each}`.
- **No membership tests left.** `aheadSessions` was `numberedSessions.filter((np) => !playedSessions.includes(np))`, and `scheduleRows` filtered the same way against `collapsedSessions`. Those compare *object identity* against arrays that are rebuilt on every recompute — correct today only because `numberedSessions` happens to be memoised, and silently degrading to "every row is still ahead" the moment anything maps or clones the wrappers in between. A single `hidden` count now drives both slices, so the problem is gone by construction rather than papered over with a `Set`.
- **`showPlayed` moved out** of the derived block, with a note on why it survives a data refresh but not match navigation.
- **Template carries no logic**: `{#each schedule.rows …}` instead of the ternary-with-spread, and `schedule.collapsedCount` in the gate, the label and the border condition.

Rendered output is unchanged — I traced all three branches (under two played; played with something ahead; played with nothing ahead) against the old template expression. `npm run lint`, `svelte-check` (0 errors, 0 warnings) and prettier are clean.

Two things I deliberately did **not** do, so they stay yours:

- **The `session` → `parts` rename (#1).** I left every `session` identifier and the "played session(s)" label alone so your rename stays one mechanical sweep. `schedule` and `collapsedCount` are already neutral.
- **The ordering change.** My original note suggested pairing this with hiding-by-part-id so authored "Part N" order survives. That changes rendered order, which is still an open question on this PR, so `4f7f552` preserves played-then-ahead exactly. Worth settling separately.

becked:
All four review items are addressed in 0326a14.

**1. "Session" → "part."** `numberedSessions` → `numberedParts`, the comment block on the schedule partition, the "sittings" comment the PR had reworded, and `partPlayed`'s doc comment. The toggle label now reads "N played part(s)" so it matches the "Part 1" / "Part 2" rows directly beneath it. The pre-existing "single-session match" adjective is left alone.

**2. One aged-out boundary.** `partPlayed(part, now = nowMs())` takes an optional clock, so `liveAndUpcoming` calls it with the `now` it already threads: `else if (!partPlayed(np.part, now))`. `schedule.ts` no longer imports `LIVE_WINDOW_MS`. The three branches classify identically — that `else if` is reachable only when the instant parsed and `t <= now`, and `scheduledParts` has already dropped null `scheduled_at`, so `!partPlayed(part, now)` reduces exactly to the old `t > now - LIVE_WINDOW_MS`.

I took the optional `partInstant` collapse as well. It replaced five copies of the `scheduled_at`-null → `Date.parse` → `Number.isNaN` guard — the four you listed plus `liveAndUpcoming`'s own variant — and with them the two ad-hoc workarounds for the helper's absence: the `as string` cast in `upcomingScheduledParts`, sound only because `scheduledParts` had pre-filtered nulls, and the `?? ""` that laundered null into `NaN`. Unscheduled and unparseable already behaved identically at every site, so each conversion is behaviour-preserving. `scheduledParts` keeps its raw null check — it deliberately retains a part with an unparseable time, so routing it through `partInstant` would change what the schedule lists.

**3. Toggle matches `TournamentMapsPanel`.** SVG chevron with `transition-transform` + `class:rotate-90={showPlayed}` in place of the ▸/▾ glyphs, `aria-label={`${showPlayed ? "Collapse" : "Expand"} played parts`}` alongside `aria-expanded`, and the established `part{n === 1 ? "" : "s"}` pluralisation. Two deliberate differences from that panel's classes, both about local fit rather than the pattern: the chevron is `h-3 w-3` (that row is `text-xs`, this label is `text-[10px]`) and it inherits `currentColor` instead of `text-tan opacity-70`, so it follows the button's own `text-muted` → `hover:text-tan` transition rather than sitting at a fixed tan against a muted label. Happy to make them byte-identical if you'd rather.

Rendered row order and which rows are hidden are unchanged — the question of whether the played/ahead partition should preserve authored "Part N" order is still open and not settled here.

`npm run lint`, `npx svelte-check --threshold warning` (0 errors, 0 warnings), and prettier on the three touched files are all clean.

Issue #85, opened 2026-07-01
Sprites baked with inconsistent framing/size across categories (law vs trait, etc.)


becked:
Baked sprites are framed inconsistently across (and within) categories, so the same render box produces different visual glyph sizes. Surfaced by the Military-tab event rail in #79, which is the first UI to place law and trait/archetype icons side-by-side at a uniform size.

**Evidence** — native dimensions of a few baked PNGs (`static/sprites/`):

- laws: `128×128`, `40×39`, `128×128`
- traits: `64×64`, `32×31`, `28×28`

Some sprites are trimmed to the glyph (the odd sizes like `40×39`, `28×28`); others are a fixed padded canvas (the power-of-two `128×128`, `64×64`). When `SpriteIcon` scales each into, say, an 18px box, a padded 128² sprite shows a noticeably smaller glyph than a tightly-trimmed one — and the variance exists *within* a category, not just across, so a per-category size tweak can't fully fix it.

**Why it matters:** it's why the tooltip code already hand-tunes per-category pixel sizes (`law 14`, `unit 15`, `rating 13` in `MilitaryTab.svelte`), and why the event rail's law vs archetype icons look mismatched. It's latent everywhere sprites are rendered; it only became obvious when two categories were shown together at one size.

**Fix (asset pipeline, not a UI change):** bake sprites with consistent framing — trim every sprite to its content bounds (or pad them all identically) in `npm run bake:sprites` — so all categories render at a consistent visual size everywhere: the rail, tooltips, and any future consumer. A nice side effect: the per-category pixel-size hacks in the tooltips can go away.

**Not introduced by #79.** The sprite outputs are gitignored and baked locally; #79 doesn't touch the sprite baker, sprite outputs, or the manifest. It only exposed the pre-existing inconsistency.

becked:
## Plan (refined): normalize **tech** framing on the Military rail with a uniform **10% inset crop**; keep the plates

Decided approach after investigating the assets this session. Self-contained so it can be picked up fresh.

### Scope

[...]

PR #200, opened 2026-08-06
feat(tournament): Matches tab on the stats page — sortable match list with per-game links + archetype glyphs

alcaras:
Implements Siontific's request from #community-tools: *"stats page should have a match list with easy clickable access for each game in list form."*

## What it does

The tournament stats page gains a **Matches** tab (now the landing tab — happy to flip the default back to Players if preferred): every match as a list, default-sorted by the global match number, with sortable headers and **status facet chips** — completed-only by default, since the list's job is finding played games.

matches tab

Each row carries:

- **#** — the global match handle, its own sortable column (the Match cell drops its inline copy when the # column is displayed, so the handle isn't shown twice).
- **Match** — the shared player cells: nation crest, **starting-ruler archetype glyph** (new), avatar + profile-linked name, winner bolded orange / loser dimmed (the table's established result treatment — no perspective-dependent W/L column), bracket + map subtext.
- **Scheduled start** — the existing time cell (Completed badge for decided matches).
- **Game** — a per-row **View game** link to the uploaded save; "—" when none is linked.

Above the table, one multi-toggle chip per status bucket (Completed / In progress / Scheduled / Unscheduled) with live counts, faceting through the previously-dormant `MatchTableState.filters` (its first consumer — new `filterMatchRows` helper beside the shared sort/search helpers). An empty selection means no filter, the same convention the game-detail tables give the field, so toggling the last chip off widens to everything rather than emptying the table. All-matches view (chips off):

all matches

## How

- **No new fetches**: the `[slug]` layout already loads standings, bracket, and matches for every tournament subpage; the tab reuses the shared `MatchTable` (the one match table) with two new registry columns (`number`, `game`) rather than a second table component.
- **Worker**: `slot_a/b_archetype` join `slot_a/b_nation` on serialized matches, resolved from the same `player_summaries` row via the slot↔player_index bridge and null in the same cases (no save, bye, forfeit, legacy — the glyph simply doesn't render). The `loadNationsForMatches` projection collapsed into the shared `loadPlayerSummaryFieldsForMatches` loader, which now carries `(nation, archetype, is_winner)` for both match serialization and the player-picks stats — one loader, no second query shape. No schema change (`starting_ruler_archetype` has existed since migration 0002).
- Because the archetype glyph lives in the shared player cell, the matches page and cast surfaces gain it too — every surface that shows the nation crest shows the leader roll beside it.

## Tests / checks

- New integration test (colocated with the `linkGame` harness in `stats.test.ts`): the matches payload resolves both sides' nation + archetype, and an unlinked match yields nulls, not missing keys.
- `cloud npm test` — 716 passed; the only failures are the 4 pre-existing `canonical-map-options` mirror failures already on `main`.
- `npm run lint`, `svelte-check` (0 errors), prettier clean.

Screenshot is local dev against the imported OWCT data — some local roster rows predate the archetype derivation, so glyphs appear only where the data has them (e.g. match 012); prod data is fuller.


alcaras:
Self-review (correctness + fit):

**Checked, no finding:**
- **All serializeMatch call sites** (matches list, bracket rounds, single match, per-user tournaments) receive the widened summary map — enforced by the param's type change (`Map<string, string>` → `Map<string, PickSummary>`), so a missed caller would fail `tsc`, and cloud `tsc` is clean. Bracket-nested matches keep their declared `TournamentMatch[]` shape honest.
- **Parallel surfaces**: the archetype glyph lives in the shared player cell, so the matches page, Up Next panel, Cast view, and the profile's Tournaments tab all gain it together. The Swiss/championship *bracket cards* are a different, deliberately compact component (crest + name only) and were left alone. The Match cell's inline number is suppressed only when a surface displays the `#` column — existing surfaces render unchanged.
- **Reuse**: no second table — the stats list is the shared `MatchTable` plus two registry column defs; `matchSlotArchetype` mirrors `matchSlotNation` (snapshot-only, no live fallback, so substitutions never change it); the game link uses the `resolve("/games/[id]")` idiom.
- **Nulls degrade**: no linked save / bye / forfeit / pre-derivation roster rows → null archetype → no glyph, same contract as the crest.
- The placeholder synthesizer (`bracket-placeholders.ts`) emits the two new keys as null, keeping the client-side literal assignable.

**Known trade-offs, flagged in the body:** Matches is now the stats page's default tab (one-line revert if you'd rather land on Players); the new matches-payload test is colocated in `stats.test.ts` because that's where the `linkGame` roster harness lives, at the cost of the filename being slightly broader than the endpoint it tests.

alcaras:
Review pass (full diff, correctness + fit per the checklist) — one finding, fixed in 86b0bd1:

- **Domain vocabulary (test fixture)**: the linkGame harness seeded `starting_ruler_archetype` as an invented `ARCHETYPE_HERO`; the parser actually stores `TRAIT_<name>_ARCHETYPE` (the shape `formatArchetype`/`archetypeSpriteKey` are built to strip — confirmed against real player_summaries rows). Production code round-trips either, so this was test-only, but the fixture now speaks the game's vocabulary.

Verified, no change needed:
- Every `serializeMatch` caller (matches list, bracket rounds, match detail, per-user tournaments) passes the widened summary map — the param's type change makes a missed caller a compile error, and cloud `tsc` is clean; bracket-nested matches keep `TournamentMatch[]` honest.
- Every existing MatchTable surface picks columns without `number`/`game` (matches page, Cast view, Up Next, profile tab: `time/matchup/broadcast[/actions]`), so their renders are unchanged except the intended archetype glyph.
- Sprite path: `traits-trimmed` category with fallback to `traits/` per the #85 rail treatment; glyph absent whenever the roster row predates the derivation — same graceful-absence contract as the crest.
- Nulls: `??`/`!= null` in the data layer throughout; the `game` column's sort pins game-less rows last via the shared nulls-last comparator.

One known quirk, deliberate: on a fully-completed tournament the **Scheduled start** sort is inert (completed rows have no sort instant — the matches page's "completed pin last" semantics, shared). The **Game** column sorts by `reported_at`, which is the meaningful chronological order for decided matches, so the capability exists where the data does.

alcaras:
Added status facet chips per alcaras: the list opens completed-only (per-bucket counts on the chips — Completed / In progress / Scheduled / Unscheduled), and deselecting chips widens the view; empty selection = everything. Implementation gives `MatchTableState.filters` its first real consumer via a shared `filterMatchRows` helper (empty-means-no-filter, the game-detail tables' convention), so the matches page can pick the same faceting up later. Lint/svelte-check/prettier clean; commit 54b79e4.

becked:
Rebased this onto current `main` and pushed four review fixups straight to the branch. The rebase rewrote history, so you'll want `git fetch && git reset --hard origin/feat/stats-match-list` (or whatever you call your fork's remote) before you touch it again.

The rebase hit one conflict, in `matches-table.ts`: `main` had rewritten the `actions` column comment — it now covers the schedule editor as well as the cast buttons — while this branch inserted the `game` column immediately above it. Kept both.

## Fixes

**Archetype glyph on every surface that shows a crest** (`44065a7`) — the glyph only reached the shared match-table player cell. Three other surfaces resolve `matchSlotNation` and were still rendering the nation alone: `MatchPopover`, `SwissFlowBracket`, and `ChampionshipBracketTree`. `svelte-check` can't catch this, which is why it passed clean — those surfaces read the slot fields through structural subset types, so a missing field is a narrower type rather than an error. The championship tree re-shapes matches into its own positioned model, so `slot_a/b_archetype` needed threading through both local types and all three object literals before the card could render it.

Same commit corrects the documented wire shape on `TournamentMatch`. The parser stores `TRAIT_<name>_ARCHETYPE` (see `derive-player-summary.ts`), not `ARCHETYPE_HERO` — and since `formatArchetype`/`archetypeSpriteKey` both strip a trailing `_ARCHETYPE`, the documented form would have missed the label and the sprite alike. Your own test fixture had it right.

**Scheduled start now follows the tournament clock** (`b9d1f6c`) — the table was pinned to `zone="local"`, but the `[slug]` layout owns a single `ZoneClock` (seeded `?zone=` > cookie > local) and renders the UTC/Local toggle on Stats as well as Matches. Before this tab the toggle was inert here, so nothing disagreed; now that the page carries times, flipping it left them in local. It reads the clock from context, the way the matches page does. The profile tab's `ZONE = "local"` isn't precedent for pinning — that surface lives outside the layout that provides the clock.

**Status chips are now the segmented switch** (`ff216c5`) — swapped the hand-rolled multi-toggle chip row for the segmented control the matches page's view switch and the game-detail tabs already use. Status is one bucket at a time now, so `All` leads the row: single-select has no equivalent of "deselect the last chip" to get back to everything. Each segment keeps its count.

The selection still lives in `MatchTableState.filters` as a 0-or-1 array rather than a second state field, so the lit segment derives from what the table actually filters on and the two can't drift. `All` is the empty selection `filterMatchRows` already reads as "no filter", so the segment needs no special case.

**Hoisted the `#` column check** (`bcd94cf`) — `columns.some()` was running inside the rows × columns render loop to answer a question about the table rather than the row.

## Left as-is

Matches stays the landing tab. It does mean bare `/stats` links land somewhere new than before, but it's the tab that answers the original request, and keeping the first tab and the default tab the same is worth more here than preserving those links.

## Checks

`npm run lint`, `svelte-check` (1309 files, 0 errors), `prettier --check src/`, and `cloud npm test` — 782 passing, with the 4 `canonical-map-options` failures that are already on `main`.

Two calls in here that were the right ones and worth naming: folding `loadNationsForMatches` into `loadPlayerSummaryFieldsForMatches` rather than adding a second query shape, and pointing the shared `MatchTable` at two new registry columns instead of standing up a second table component.

becked:
Swapped the game link's chevron for Old World's own replay glyph (`318f204`) — the icon the game uses for reviewing a played game, which is where the link goes.

Baked from `events_images/UI_HUD_Icon_Replay.png` into `icons/REPLAY`, renamed on the way in like the neighbouring `ICON_MAPPINGS` entries (`Cycle_Normal_EndTurn` → `TURN`). Its glyph already fills the source canvas, so unlike `LAWS` it needs no `trim`. `bake:sprites` + `bake:finalize` produced a one-line manifest addition with no churn in the other 1271 entries.

Rendered at 18px via the shared `SpriteIcon`, already imported in `MatchTable` for the crest and archetype glyphs, replacing the hand-rolled inline SVG.

Left alone deliberately: `MatchPopover`'s "View full game" is a bordered button with no icon — a different treatment rather than the same arrow, so it seemed like a separate call rather than a parallel surface to keep in sync.

Lint, `svelte-check` (0 errors / 0 warnings, 1309 files), and prettier all clean.

PR #206, opened 2026-08-09
fix(tournament): the casting panel lists a pending match's upcoming sittings

alcaras:
Follow-up to #197, from the same match (OWCT 038, now on part 8): the schedule block correctly leads with the upcoming Part 8, but the **casting panel below lists only parts 1–7** — it drops parts with no broadcast info, so a scheduled next sitting with nobody signed up is invisible there. Reading the casting list top-to-bottom, part 8 doesn't exist; its only trace is the schedule row above, which is easy to read past.

## The fix

A part still ahead (or live, within `LIVE_WINDOW_MS`) of a **pending** match now appears in the casting panel regardless of broadcast info, carrying a small italic *Needs a caster* note — an upcoming sitting with nobody signed up is exactly the seat this panel should be advertising, and the match tables already flag the same state ("needs a caster" + the I'll cast button). Played parts keep the archive rule: listed only when they have a caster or stream to show, so a played part that archived only a stream link renders just the link, as before.

Local repro of the live state (7 played parts + an uncast Part 8 in 24 minutes) — the casting panel now ends with Part 8 / *Needs a caster*, consistent with the Live & Upcoming flag up top; played-but-uncast Part 7 correctly stays out:

casting panel

One file, read-view only, no data changes. `partPlayed` reads the shared clock, so an uncast part drops out of the panel as it ages into the played group. `npm run lint`, `svelte-check` (0 errors), prettier clean.


alcaras:
Self-review (correctness + fit) — two findings, both fixed in the follow-up commit:

1. **Unscheduled parts leaked into the seat ad (correctness).** `partPlayed` is false for a part with no `scheduled_at`, so the widening also listed *unscheduled placeholder* parts with a "Needs a caster" note — but there's nothing to turn up for, the very line the match tables draw (`matches-table.ts`: "Casting an unscheduled match is meaningless"), and blank parts do occur in practice (the add-a-part-so-it-shows-up admin flow). The predicate now requires `scheduled_at != null`.
2. **Duplicated predicate (DRY).** The advertise condition appeared verbatim in the `castingParts` filter and the template's placeholder gate — the classic drift pair. Extracted as one `partAdvertisesSeat(part)` shared by both.

Verified, no finding: a *played* part that archived only a stream link renders just the link (no false "Needs a caster"); bye/complete/forfeit matches are excluded by the `pending` check; reactivity holds (the predicate reads the shared clock via `partPlayed`, so an uncast sitting drops out of the panel as it ages into the played group, and the panel's `length > 0` gate follows); a single-part pending match now advertises its seat too, consistent with the Live & Upcoming flag for the same state. Re-verified the render on the 8-part fixture after the fix — identical. Lint/svelte-check/prettier clean.

becked:
Rebased on current `main` (clean — nothing had moved near `MatchPopover.svelte`) and pushed one cleanup commit on top, so the head is now `456a84b`. Heads up that the rebase rewrote your two commits' SHAs, so reset rather than pull if you have this checked out locally.

The bug is real and the diagnosis was right — reading the casting list top-to-bottom, part 8 didn't exist. What I changed is how "this sitting is still advertising a seat" gets expressed.

**`partAdvertisesSeat` → the shared helper.** This file already computes "this match's still-upcoming scheduled sittings" eleven lines above your predicate: `casterPostText` calls `upcomingScheduledParts([match])`. `parts.ts` bills that helper as "the one definition of 'upcoming' shared by every surface that lists it … so they can't drift." The panel now derives `advertisedPartIds` from `upcomingScheduledParts([match], CAST_GRACE_MS)`, and both the filter and the flag test membership in it — so the anti-drift property your second commit was after is preserved.

**`CAST_GRACE_MS` rather than `partPlayed`.** `!partPlayed` draws the line at `LIVE_WINDOW_MS` (4h), the viewer-facing "is this plausibly still being broadcast" window. Every caster-facing surface uses `CAST_GRACE_MS` (2h) — `CastView`, and the matches page's `needs-casters` scope — and that constant's own comment covers this case exactly: a sitting stays upcoming "and so still claimable / **still listed as needing a caster**" for 2h past its start. On the 4h window the popover advertised seats the Cast tab had already retired.

Your implicit argument for `partPlayed` — keeping the two stacked panels on one boundary — was the thing I weighed hardest here. What decided it: the panels are already deliberately not row-for-row, because a played part with no broadcast info is absent from casting by design (the archive rule your PR keeps). So the alignment that choice protects doesn't actually exist.

**Guards come along for free.** `upcomingScheduledParts` owns the `pending` / `slot_b_id != null` / has-a-parseable-time checks. The `slot_b_id` one was missing, and every sibling predicate carries it (`rowIsPendingSitting`, and `upcomingScheduledParts` itself). I couldn't construct a state where it actually bit — the panel is gated on `!isPlaceholder` and the worker refuses to schedule a bye — so consistency rather than a live bug.

**The flag is now the tables' chip.** `MatchTable` renders this same state as an orange uppercase badge, and the muted italic was the quietest element in the panel whose whole job is recruiting. Same markup now, plus `shrink-0` so the badge survives beside a stream link pinned right.

Two behavior deltas, both intended:

- a casterless sitting 2–4h past its start is no longer advertised — it stays in the schedule panel above, which still groups on `partPlayed`
- a part whose `scheduled_at` doesn't parse is excluded; the old check was `!= null` alone

Also updated the block comment above the panel — it still described the caster-or-stream rule this PR replaces.

`npm run lint`, `svelte-check` (0 errors), and prettier are clean. The rendered behavior is being checked by hand against a local fixture with seven parts straddling both boundaries; I'll follow up here if that turns anything up.

One thing I deliberately left alone: `MatchTable`'s flag has no window at all (`rowIsPendingSitting` → `mostRecentScheduledPart`), so it still lights up on pending matches whose sittings are weeks past. Unifying that would give the concept a single boundary everywhere, but it changes existing table behavior and belongs in its own PR.

PR #207, opened 2026-08-10
feat(tournament): per-division progress lanes + projected match totals in the overview hero

alcaras:
The overview's progress hero showed only the current round ("Round 2 of 5 · 6 of 8 matches reported"), with no overall sense of stage or total. Prompted by alcaras + a look at what happens when the divisions' rounds split — which the current header mis-renders twice over, since Swiss rounds generate **per division** and one division routinely starts round N+1 while the other still plays N.

## The hero, redesigned as a stage map

clean

- **One Swiss lane per division** — a cell per round; the division's *open* round renders as **per-match pills** (filled as reports land), played rounds solid, future rounds empty. Each lane is labeled beneath with its division and round, with its open-round count at the right.
- **The lanes merge into a single full-width Championship bar** — the divisions play Swiss apart and reunite in one bracket. Until the bracket exists it advertises its projected size ("31 matches · awaiting Swiss"); once live it switches to pills like the lanes.
- **Shared round labels** light for whichever rounds are open *anywhere*. During a division split both labels glow, each lane sits on its own round, and the header text collapses to "Rounds 1–2 of 5" (simulated split below):

split

- **Overall tally**: "24 of ~157 matches" — played so far against the **projected eventual total**.

## The projected total (`projected-totals.ts`)

The double-threshold Swiss makes the eventual match count *nearly* deterministic: within a record bucket, every game sends one player up and one down regardless of who wins, so bucket sizes — and each round's match count — evolve identically under any results. The only exception is a cross-record game (an odd bucket's floater playing down): its outcome changes how many players leave the field, swinging later rounds by a match. The new census walk forks exactly there and returns a min/max envelope — the header shows an exact number when the envelope is closed and "~N" when a pivot is still open (for a real field the spread is 0 or 1). Qualifier counts ride along (championship = qualifiers − 1, "no cap" per docs/tournament-rules.md) and in practice come out deterministic even when the match count doesn't — one more early advancer means one fewer game in the tail bucket.

Validated against 2,000 randomized futures simulated through the real pairing engine (`pairSwissRound`): identical envelopes — on live OWCT data, Division B's remaining count is exactly 34, Division A is 31–32 with one open pivot, and both divisions' qualifier counts are already deterministic (15 + 16 → a 30-match bracket).

## Notes

- Byes are excluded from every count (they auto-resolve and were never played).
- The old `overall` fraction bar and `effectiveRoundsDone` are replaced wholesale; `HeaderHero`'s in-progress shape changes accordingly (TournamentHeader is its only consumer).
- No API or Worker changes — everything computes from the loads the page already has.
- `npm run lint`, `svelte-check` (0 errors), prettier clean. Split-state screenshot from a locally simulated division split (fake round-2 rows), since the live data is still in lockstep.


alcaras:
Deep review pass (correctness of the census walk + fit) — two findings, both fixed in the follow-up commit; everything else verified clean.

**Fixed:**
1. **Negative census on an unplaceable pending pair (correctness).** The walk decremented both participants' buckets unconditionally; a pending match whose player isn't in the active census (a withdrawn player's not-yet-forfeited game — withdrawal normally auto-forfeits, but the module shouldn't bet on every caller upholding that) would drive a count negative and corrupt every subsequent round. Such pairs are now skipped per-branch, degrading to "ignore that game" instead.
2. **Undocumented approximation (docs).** The bye recipient is modeled as the worst-record bucket, but the engine skips players who already had a bye — invisible at census granularity. Now documented at the site; even fields (the common case, including OWCT) never reach the bye path at all.

**Verified, no change needed:**
- The round fold iterates start-of-round buckets shared across branches — sound, because forks only ever write the *next* census; the start census is immutable within a round.
- Floater mechanics mirror the engine: one cross-record game per odd bucket (the only fork point), leftover floats on; a floater passing a bye-emptied tail bucket falls through; an orphaned floater is parity-impossible on an even field (defensive re-add only).
- Same-record games never fork — one up, one down under any result — which is the whole reason the projection is near-deterministic.
- Envelope semantics: dedupe on (census, matches, qualifiers) keeps branch count tiny; min/max collapse to an exact display when closed, "~N" when open.
- **Re-validated after the fix** against 2,000 randomized futures through the real `pairSwissRound`: envelopes identical (A 31–32 / 15 qualifiers, B exactly 34 / 16).
- Header side: lanes and labels index the same `rounds` arrays (equal length by construction); byes excluded from every count; the `{#key}`-free re-render is pure derived state; `Progress` reused for solid cells, pill rows carry `role=progressbar` with value semantics.
- Removed code (`effectiveRoundsDone`, the 0–0.5/0.5–1 `overall` fraction) had no other consumers — `TournamentHeader` is `HeaderHero`'s only renderer.

**Disclosed:** `projected-totals.ts` has no CI test — the frontend has no unit-test runner (the validation harness lives outside the repo). If you'd like, the module is pure enough to move behind a small vitest setup or mirror into `cloud/src` for its test infra, but I didn't want to invent a new test lane unilaterally.

becked:
Rebased the branch onto current `main` — clean, no conflicts. `svelte-check` is 0 errors.

Two things I went ahead and fixed locally:

- **`npm run lint` was red** — 3 × `'_' is defined but never used` in `TournamentHeader.svelte`. Worth noting this isn't the usual stale-branch case: `eslint.config.js` is unchanged between this branch's base (`06a4fb8`) and `main`, so the branch never linted clean. There's no `varsIgnorePattern` in the config and `_` as an `{#each}` item name appears nowhere else in `src/`, so there was no existing idiom to copy. Fixed by making the loops iterate something they use — index arrays for the round labels, and a boolean-per-pill array (`(_, p) => p < r.done`) for the two pill rows, which also drops the index comparison out of the class expression.
- **The `pendingPairs` loop forked exponentially with no dedup.** The per-round loop collapses branches by census signature; the pending-pairs loop didn't, so each cross-record pending pair doubled the branch set. Measured on the real module: 12 pairs 23ms → 16 pairs 325ms → 20 pairs **5.2s**, clean 2^k. Fixed by extracting the signature helper and applying the same collapse as it lands. Same input now runs in 1ms with an identical envelope, and 400 randomized A/B cases against the pre-fix version came back with zero mismatches.

The rest I haven't touched:

1. **`projectedExact` doesn't cover the bye approximation.** The envelope forks only on cross-record floater games. The odd-division bye is a documented approximation — `pickByeRecipient` skips players who already had one, which a record census can't see — but it doesn't widen the envelope, so an odd division can report a confidently *exact* number that's wrong. `docs/tournament-rules.md` treats 29/27 (both odd) as a configuration admins actually hit. Either that case should force the `~`, or the limitation should be stated where the number is produced.
2. **`~N` is the ceiling, not a midpoint.** `projectedTotal: projectedMax`. At tournament start a 28-player division projects 57–59 matches, so the opening two-division figure runs ~2–4 high and progress reads behind. A range, or the midpoint, would match how a reader takes "~".
3. **No tests, and nowhere to put them.** `projected-totals.ts` is 254 lines of non-trivial combinatorics with zero coverage. All vitest infra lives in `cloud/`; there's no test runner configured for `src/`. The central correctness claim in the description — 2,000 randomized futures through the real `pairSwissRound` — is a script that isn't in the diff, so it can't be re-run or regression-guarded. This is the biggest gap; the options are standing up vitest at the repo root or relocating the pure projection where `cloud/`'s runner reaches it.
4. **`Math.min(...openRounds)` on an empty array → `Infinity`**, rendering "Rounds Infinity–-Infinity". Only reachable with status `swiss` and zero Swiss matches, and `handleStartTournament` writes the status flip and round-1 generation in the same D1 batch, so I don't believe it's live-reachable — robustness nit only.
5. **Three stacked round indicators**: "Rounds 1–2 of 5" in the header, "Swiss 1…5" above the lanes, and "Division A · Round 2" beneath each lane.
6. **`roundLabel` is a pre-formatted display string built in `+page.svelte`**, though `header-status.ts` is the module that shapes display state — the component previously composed "Round {n} of {total}" itself.
7. **Pill density**: a 28-player division's round 1 is 14 pills inside one of five lane cells. Worth eyeballing in the running app before this merges.

Things I checked that hold up: `["A", "B"] as const` matches the established pattern (7 other sites, 4 in this same file), and two divisions is always right since start requires ≥1 slot in each; `HeaderHero` has exactly one consumer, so the shape change is fully wired; byes are excluded consistently and the matches endpoint has no limit, so the tally denominator is the complete list; and the floater arithmetic matches `pairSwissRound` — it unshifts the floater onto the next bucket and pops a new one when the combined size is odd, which produces the same per-round counts and leftover-floater records the walk assumes. `bracket.ts` inserts phantom byes as `status='bye'`, so playable championship matches really are qualifiers − 1.

becked:
Worked findings 1–3 from my last comment and pushed three commits. Findings 4, 5 and 7 are untouched; **6 I'm withdrawing** — see the bottom.

Before the specifics: to check any of this I built the validation harness the description mentions (randomized futures through the real `pairSwissRound`), and it moved my read on finding 1 in both directions. Details below, since one half of it is me correcting myself.

| | |
|---|---|
| `304bc8f` | `fix(tournament): the census walk plays out a lone survivor's byes` |
| `8c76126` | `test(tournament): cross-check the match projection against the pairing engine` |
| `ad40be8` | `fix(tournament): the projected match total reads from the envelope's midpoint` |

## 1 — it was two defects, and I had the reach of each backwards

**Your bye note was right about the mechanism; I was wrong to think it was the only thing there.** Driving 3,000 seeded futures per field size through the real engine and comparing against `projectSwissDivision` turned up violations that split cleanly in two:

**(a) A lone-survivor bug, unrelated to the bye approximation.** `playRound` stopped at `if (total < 2) return [start]`, but the engine doesn't stop — `pairSwissRound` hands the lone active slot a bye every remaining round, and those free wins can carry it past the win threshold. The walk reported a confidently *exact* count that was too low: a 4-player division projects `2–2` qualifiers where 3 can qualify. Fixed in `304bc8f` — only an empty census ends the walk now. The odd-field branch below already handled a one-player field correctly on its own, so the early return was the whole bug.

**(b) The bye-recipient approximation — real, but it needs a *repeat* bye, which is much narrower than "odd division".** I isolated it by patching a copy of `pairing.ts` to drop the had-a-bye skip: with (a) fixed and the skip removed, every violation across 6 configs and n = 2…34 disappears. So it is precisely and only that. But the condition it needs is the same player byed twice:

```
n= 5:  >=1 bye 3000/3000   >=2 byes 3000/3000   same player byed twice  371/3000
n=26:  >=1 bye  758/3000   >=2 byes  372/3000   same player byed twice    0/3000
n=27:  >=1 bye 3000/3000   >=2 byes 3000/3000   same player byed twice    0/3000
n=28:  >=1 bye 1485/3000   >=2 byes  745/3000   same player byed twice    0/3000
n=29:  >=1 bye 3000/3000   >=2 byes 3000/3000   same player byed twice    0/3000
```

At 27 and 29 — the odd split I pointed at as the risk — a bye lands in *every round of every future* and the recipient is a different player every time, so the census guess is right every time. The repeat only appears once the field is small enough to run out of never-byed players: n=5 under 3W/3L, up to n=13 under 4W/4L.

So I've **not** forced the `~` for odd divisions. Doing so would fire 100% of the time at 27/29 and ~50% at 28, degrading a display that is provably exact there, to guard a case that only exists below n≈6. Instead the comment now states the actual boundary, and the test pins it from both sides.

Two claims in the existing comments were also false and are corrected: `projected-totals.ts` said *"Even fields (the common case) never reach this"*, and `f771c9b`'s message says *"even fields never reach it"*. They do — advancers and eliminated players don't leave in pairs, so parity flips mid-Swiss and a 28-player division takes a bye in about half its futures. Byes are routine at every size; it's the repeat that isn't.

## 2 — `~N` now reads from the midpoint

```
28/28: envelope 141-147  ->  "0 of ~144 matches"  (was ~147)
30/26: envelope 140-146  ->  "0 of ~143 matches"  (was ~146)
16/16: envelope  81- 81  ->  "0 of 81 matches"    (unchanged - closed envelope, no ~)
```

One thing I extended past the finding as written, flagging it rather than burying it: the championship bar's `{exact ? "" : "~"}{total} matches · awaiting Swiss` draws on the *same* envelope and was also taking the max. Left alone it would have contradicted the total a line above it, so it takes the midpoint too (`~28` instead of `~29` on a 28/28 field). `projectedExact` is untouched, so a closed envelope still prints a bare number.

## 3 — the test, and where it ended up

`cloud/src/tournament/projected-totals.test.ts` — 32 tests, ~0.6s, in cloud's existing **`unit`** project (plain Node pool, no miniflare). The module stays where its consumer is; the test reaches across the root exactly the way `canonical-map-options.test.ts:15` already does, and nothing about the Worker bundle changes.

That home is the point rather than a compromise: `pairSwissRound` and `computeRecord` are siblings there, so the description's "2,000 randomized futures" claim becomes a thing that actually runs in the repo. Standing up vitest at the root was the alternative, but it would still have had to reach into `cloud/` to keep the cross-check — same coupling, more setup, and it wouldn't run in CI either way.

Coverage: envelope containment **and exact tightness** over 300 seeded futures per size at n = 6…32 plus a 4W/4L 7-round config (tightness is the claim the display rests on — a loose envelope prints `~` for a total that was actually pinned); the lone-survivor bug from (a), verified to fail on the pre-fix code; the bye boundary from both sides; and the two earlier fixes on this branch that shipped untested — the unplaceable pending pair (`f771c9b`) and the branch collapse (`72cf697`, whose 2^k regression is caught by a 1s timeout).

## 6 — withdrawn

`roundLabel` being a pre-formatted display string built in `+page.svelte` isn't a departure. `phaseLabel: string` was already exactly that on `main`'s `HeaderHero`, and `header-status.ts` only ever exported `headerStatusMeta` — it never built the hero. My mistake; there's no convention here to break.

## Still open

- **4** — `Math.min(...openRounds)` → `Infinity`. Re-verified as not live-reachable (status flip and both divisions' round 1 land in one `env.SHARE_DB.batch`; withdrawal is rejected during setup; the layout loader throws rather than defaulting `matches` to `[]`). Robustness nit, one line, your call.
- **5** — three stacked round indicators. Design call, left for you.
- **7** — pill density. Still wants an eyeball in the running app.

## Verification

`npm run lint` clean · `svelte-check` 1315 files, **0 errors / 0 warnings** · `cloud` `tsc --noEmit` clean · `cloud` `npm test` **78/79 files, 817/821 tests** — the 4 failures are `canonical-map-options.test.ts`, the known DOTA schema drift that is red on `main` too · prettier clean.

